PR #36, opened 2023-10-24
Font loading and more v4 documentation refinements

MattCheely:
I also switched the concurrent runner to run-pty like in our nav app so things can be restarted individually. I kinda just went with the lazy setup though, so it doesn't have nice error icons and all that. Somebody should berate me into doing that and/or refuse to approve the PR until it's done, probably. 😅


gavin-everett-genesys:
Looks like `4.6` was only released 2 days ago https://github.com/ionic-team/stencil/pull/4967

gavin-everett-genesys (approved):

MattCheely:
🤔 I need to add some tests to this.

PR #45, opened 2023-10-25
fix(list): fixed list container style

daragh-king-genesys:
fixed list container style

COMUI-2398

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-context-menu/gux-context-menu.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-context-menu/gux-context-menu.scss
@@ -1,3 +1,5 @@
+@use '../../../style/mixins';
+
 :host {
   display: block;
   max-width: fit-content;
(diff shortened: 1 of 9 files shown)


packages/genesys-spark-components/src/components/stable/gux-context-menu/gux-context-menu.scss
```diff
@@ -50,16 +52,6 @@ button[disabled] {
...
-  list-style: none;
-  background-color: var(--gse-ui-menu-backgroundColor);
-  border: var(--gse-ui-menu-border-width) var(--gse-ui-menu-border-style)
-    var(--gse-ui-menu-border-color);
-  border-radius: var(--gse-ui-menu-borderRadius);
-  box-shadow: var(--gse-ui-menu-boxShadow-x) var(--gse-ui-menu-boxShadow-y)
-    var(--gse-ui-menu-boxShadow-blur) var(--gse-ui-menu-boxShadow-spread)
-    var(--gse-ui-menu-boxShadow-color);
+
+  @include mixins.gux-menu-container;
```
jason-evans-genesys, line 56: Thoughts on us defining an element within the `gux-list` component (one that is an outer container element in that component) that has these styles defined so we don't need to set these same styles on multiple components that use `gux-list`?

jordanstith15: Isn't that practically gux-listbox then? I think the point of gux-list was to allow the user to make their own container, but maybe there are slight differences between the two that had us use gux-list as an inner component instead of gux-listbox.
That is something to look into.

jordanstith15 (approved):

PR #52, opened 2023-10-26
chore(gux-form-field-text-like): add example of programmatically sett…

katie-bobbe-genesys:
…ing input value

COMUI-2166

Added an example to the html file of how to force update the component after setting the input value 

@daragh-king-genesys I want to confirm that using `guxForceUpdate` method is the way we want users to update the form field components after setting the input value. If so, I will update [the ticket](https://inindca.atlassian.net/browse/COMUI-2166) to inform the reporter of this method


daragh-king-genesys:
@katie-bobbe-genesys ya that is all we have at the moment. Maybe we should have another look at it again after v4 is released. A `guxSetValue` method would be more user friendly or we could try again to find something native to hook into to notify us about the value change.

daragh-king-genesys (approved):

PR #57, opened 2023-10-31
fix(asset-loading): fixed default asset loading

daragh-king-genesys:
fixed default asset loading

COMUI-2421


packages/genesys-spark-components/src/utils/date/time-zone.spec.ts
gavin-everett-genesys, line 1: I think this file can be removed aswell with the phone-input readme.

daragh-king-genesys: Done

gavin-everett-genesys (approved):

MattCheely (approved):
This looks good, but also has me thinking I might revisit it in v4 to see if we need it in two places (components and genesys-spark)

PR #63, opened 2023-10-31
fix(tooltip): hiding tooltip when target is scrolled out of view

jordanstith15:
Closes [COMUI-2114](https://inindca.atlassian.net/browse/COMUI-2114)

Ticket mentions it is in Safari and FF, but I found it in Chrome too.

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-tooltip/gux-tooltip.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-tooltip/gux-tooltip.tsx
@@ -13,6 +13,7 @@ import {
   autoUpdate,
   computePosition,
   flip,
+  hide,
   offset,
   shift,
   Placement,
@@ -111,7 +112,7 @@ export class GuxTooltip {
   }
 
   private updatePosition(): void {
-    const middleware = [offset(16), flip(), shift()];
+    const middleware = [offset(16), flip(), shift(), hide()];
 
     if (this.anchor) {
       middleware.push(arrow({ element: this.arrowElement }));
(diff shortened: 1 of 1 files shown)


jordanstith15:
It still overflows over outside divs until the full target scrolls out of view. Looking into why it doesn't flip properly

packages/genesys-spark-components/src/components/stable/gux-tooltip/gux-tooltip.tsx
```diff
@@ -124,7 +125,8 @@ export class GuxTooltip {
     }).then(({ x, y, middlewareData, placement }) => {
       Object.assign(this.root.style, {
         left: `${x}px`,
-        top: `${y}px`
+        top: `${y}px`,
+        visibility: middlewareData.hide?.referenceHidden ? 'hidden' : 'visible'
```
daragh-king-genesys, line 129: @gavin-everett-genesys in popup you used 
```
        referenceHidden
          ? this.popupElementContainer.classList.add('gux-sr-only-clip')
          : this.popupElementContainer.classList.remove('gux-sr-only-clip');
```
Do you remember why you picked that over `visibility` (if there was a reason)?

gavin-everett-genesys: Using the `visibility:hidden` will hide the popup visually from the screen reader. So if the popup goes out of view the focus would be lost from the popup. If we use the `gux-sr-only-clip` the popup will also be hidden but be still accessible to screen-reader.

jordanstith15: I don't think the tooltip in this case is focusable, so would the popup reasoning apply here?

daragh-king-genesys: No I think this PR is fine as is if that was the reasoning for popup.

daragh-king-genesys (approved):

PR #69, opened 2023-11-01
Linting improvements/fixes

MattCheely:
Mostly focused on getting linting working in genesys-spark

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,4 @@
 node_modules/
-packages/
 playgrounds/
 shared-configs/
 web-apps/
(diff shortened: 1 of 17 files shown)


.prettierignore
```diff
@@ -1,5 +1,4 @@
 node_modules/
-packages/
```
MattCheely, line 2: This was preventing prettier from running in vscode for me.

packages/genesys-spark/package.json
```diff
@@ -9,6 +9,9 @@
   "scripts": {
     "build": "rollup -c",
     "dev": "rollup -c --watch",
+    "eslint": "eslint --fix './**/*.ts'",
```
MattCheely, line 12: I tried using `eslint --fix .` here as in other packages, but it wasn't picking up all of the files. It makes me a bit concerned that we might be missing some checks in other places, but I haven't fully looked into it yet.

daragh-king-genesys (approved):

katie-bobbe-genesys (approved):

PR #72, opened 2023-11-01
refactor(tokens): update tokens

gavin-everett-genesys:
update tokens

COMUI-2426


packages/genesys-spark-tokens/snapshot/css/gse-ui.css
```diff
@@ -283,8 +283,12 @@
   --gse-ui-segmentedControl-button-active-foregroundColor: #ffffff;
   --gse-ui-segmentedControl-borderRadius: 4px;
   --gse-ui-segmentedControl-height: 32px;
-  --gse-ui-segmentedControl-iconOnly-width: 32px;
```
daragh-king-genesys, line 286: How have you been dealing with stuff like this? Just merge this PR and fix in in another?

gavin-everett-genesys: Ye, I noticed that usually I check if there is any changes outside of the components which they have added tokens for. There should only be changes to tabs in this change. For this since it is small I will fix this in my `tabs` missing tokens PR. This is the first PR in which there has been a change to another component outside of the ones which require tokens.

daragh-king-genesys (approved):

PR #76, opened 2023-11-01
fix(gux-form-field-color): added border styles

jason-evans-genesys:
https://inindca.atlassian.net/browse/COMUI-2406

Added border styling.

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.light.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.light.scss
@@ -1,0 +1,22 @@
+gux-form-field-color {
+  input {
+    &:focus,
+    &:focus-visible {
+      border: 0;
+      outline: none;
+      box-shadow: none;
+    }
+
+    &::-webkit-color-swatch-wrapper {
+      padding: 0;
+    }
+
+    &::-webkit-color-swatch,
+    &::-moz-color-swatch {
+      width: var(--gse-ui-colorPicker-input-swatchSizing);
+      height: var(--gse-ui-colorPicker-input-swatchSizing);
+      border: none;
+      border-radius: var(--gse-ui-colorPicker-input-swatchBorderRadius);
+    }
+  }
+}
--- a/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.scss
@@ -18,11 +18,12 @@
 }
 
 ::slotted(input) {
-  flex: 1 1 auto;
+  flex: 0 1 auto;
   align-self: auto;
   order: 0;
-  width: var(--gse-ui-formControl-input-colorPicker-size);
-  height: var(--gse-ui-formControl-input-colorPicker-size);
+  min-width: var(--gse-ui-colorPicker-input-swatchSizing);
+  width: var(--gse-ui-colorPicker-input-swatchSizing);
+  height: var(--gse-ui-colorPicker-input-swatchSizing);
   overflow: hidden;
   font-size: var(--gse-ui-formControl-input-contentText-fontSize);
   color: var(--gse-ui-formControl-input-populatedColor);
@@ -30,8 +31,8 @@
   text-overflow: ellipsis;
   background-color: var(--gse-ui-formControl-input-backgroundColor);
   border: none;
-  border-radius: var(--gse-ui-formControl-input-borderRadius);
   outline: none;
+  padding: 0;
 
   &::placeholder {
     color: var(--gse-ui-formControl-input-placeholderColor);
@@ -59,7 +60,8 @@
       align-items: center;
       align-self: auto;
       order: 0;
-      height: var(--gse-ui-formControl-input-textfield-height);
+      width: var(--gse-ui-formControl-input-colorPicker-size);
+      height: var(--gse-ui-formControl-input-colorPicker-size);
       font-family: var(--gse-ui-formControl-input-contentText-fontFamily);
       font-size: var(--gse-ui-formControl-input-contentText-fontSize);
       line-height: var(--gse-ui-formControl-input-contentText-lineHeight);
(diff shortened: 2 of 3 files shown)


packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.scss
```diff
@@ -60,14 +60,20 @@
       align-self: auto;
       justify-content: center;
       order: 0;
-      height: var(--gse-ui-formControl-input-textfield-height);
+      width: var(--gse-ui-formControl-input-colorPicker-size);
+      height: calc(
+        var(--gse-ui-formControl-input-colorPicker-size) + 2px
+      ); // Adding 2px here so that the width and height are both 34px to make room for the border to render, otherwise it's cut off
```
MattCheely, line 66: This isn't a suitable fix, because it breaks the aspect ratio of the component - it's no longer a square. You can see this in the layout tab of the CSS inspect, which is always a good place to look when you run into this kind of issue:

Screen Shot 2023-11-03 at 9 04 28 AM

The design also calls for a 32x32 square, rather than 34x34. When you're having trouble with an element's size, it's often a good idea to check it's children. In this case, the slotted input is also out of square, which is likely part of the problem:

Screen Shot 2023-11-03 at 9 08 48 AM

The input is bigger than I expected as well at 32x32px instead of the 24x24 swatch from the designs. After removing the extra 2px on this line, I was able to get the correct sizing with these styles on the input:

```css
::slotted(input) {
  ...
  flex: 0 1 auto; //don't grow to the container, we want a fixed width
  width: 24px; // swatch size from designs
  height: 24px;
  padding: 0; // get rid of the browser padding
  ...
}
```

MattCheely (changes requested):

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.scss
MattCheely: Edit: Looks like those styles work in Firefox, but maybe not in chrome. It may take some more digging to find a solution for both, or Chrome may need some additional overrides for browser CSS.

MattCheely: It might be worth looking at [-moz-color-swatch](https://developer.mozilla.org/en-US/docs/Web/CSS/::-moz-color-swatch) and -webkit-color-swatch as options to get the styles we want.

jason-evans-genesys: Thanks, Matt for this really helpful feedback and great catches. The swatch and its container looks to render better in both chrome and firefox, but I'm still working on a few more issues with the component, such as the focus ring styling, and getting the aspect ratio to be 24x24 for the swatch, instead of 22x24 right now. I'll let you know when this is ready for review again.

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.scss
```diff
@@ -18,20 +18,21 @@
 }
 
 ::slotted(input) {
-  flex: 1 1 auto;
+  flex: 0 1 auto;
   align-self: auto;
   order: 0;
-  width: var(--gse-ui-formControl-input-colorPicker-size);
-  height: var(--gse-ui-formControl-input-colorPicker-size);
+  min-width: var(--gse-ui-colorPicker-input-swatchSizing);
```
jason-evans-genesys, line 24: I added this min-width so that the aspect ratio can be 24x24 for the swatch. Without the min-width the swatch aspect ratio is 22.2x24 (I believe this is because the browser automatically sets a default width to the input, so this min-width style is a way to override that)

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.light.scss
```diff
@@ -0,0 +1,20 @@
+gux-form-field-color {
+  input {
+    &:focus,
+    &:focus-visible {
```
jason-evans-genesys, line 4: This focus styling here will prevent 2 focus rings from rendering - we only want one focus ring on the outer container to render.

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.scss
jason-evans-genesys: Thanks for the very helpful feedback @MattCheely 

I pushed up changes for this.

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.light.scss
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    &::-webkit-color-swatch-wrapper {
+      padding: 0;
+    }
+
+    &::-webkit-color-swatch,
+    &::-moz-color-swatch {
+      width: var(--gse-ui-colorPicker-input-swatchSizing);
+      height: var(--gse-ui-colorPicker-input-swatchSizing);
```
MattCheely, line 17: This is good, but now that we can style the swatch, you can get a little closer to the designs by also adding:

```
  border: none;
  border-radius: var(--gse-ui-colorPicker-input-swatchBorderRadius);
```

MattCheely (changes requested):

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.light.scss
MattCheely: Once this change is in, let's ask the UX team to review it before we merge.

jason-evans-genesys: @MattCheely I got the approval from the UX team.

jason-evans-genesys: There's currently a conflict on style.scss, but I was thinking I could resolve this conflict when I do a git rebase when I merge this PR after it's approved.

daragh-king-genesys:
you will have to rebase this on main. It should be something like this.
```
git checkout main
git fetch --prune
git pull
git checkout feature/COMUI-2406
git rebase main
git push --force
```

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.light.scss
jason-evans-genesys: @MattCheely Just wanted to check in and see if the changes I pushed for this PR look ok and this PR is ready for approval. Thanks!

MattCheely (approved):
👍 Looks good!

PR #79, opened 2023-11-02
refactor(form-field-checkbox): ux review comments

gavin-everett-genesys:
ux review comments

[COMUI-2404](https://inindca.atlassian.net/browse/COMUI-2404)

[COMUI-2404]: https://inindca.atlassian.net/browse/COMUI-2404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/gux-form-field-checkbox.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/gux-form-field-checkbox.scss
@@ -84,6 +84,7 @@
   cursor: pointer;
   border: 0;
   outline: none;
+  margin: 0;
 }
 
 ::slotted(input[type='checkbox'])::before {
--- a/packages/genesys-spark-components/src/components/stable/gux-form-field/functional-components/gux-form-field-help/gux-form-field-help.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-form-field/functional-components/gux-form-field-help/gux-form-field-help.scss
@@ -6,7 +6,8 @@
     flex-wrap: nowrap;
     place-content: stretch flex-start;
     align-items: flex-start;
-    padding: var(--gse-ui-formControl-helper-padding);
+    justify-content: flex-start;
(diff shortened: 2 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/gux-form-field-checkbox.scss
```diff
@@ -68,6 +68,10 @@
   flex-direction: row;
   gap: var(--gse-ui-checkbox-gap);
 
+  .gux-label-beside {
+    align-self: center;
```
thomas-c-dillon, line 72: Inspecting the elements, it looks like the checkbox and not the label is what needs to be centered. Here is a quick picture without this change:

image

Aligning the text to center of the not-centered checkbox is actually making the component almost a pixel larger (of whitespace).

thomas-c-dillon: There is still some really funky stuff going on here. The heights of the items in the flexbox are not correct. In theory, if you can get them to be the correct hight then using align-self or align-items should work. I play around for about a half hour in the browser and got it really close. I think there is potentially some underlying height issues with the input.

The closest I got was by specifically setting the height on `.gux-input` to be equal to the input height and then using align-self to center. Still looked a bit off but got the overall component height to 18px. Without that height setting, it kept putting in a couple of extra pixels which tells me there is something weird going on... somewhere.

daragh-king-genesys (approved):

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/gux-form-field-checkbox.scss
gavin-everett-genesys: @thomas-c-dillon  will create issue to re-review this.

PR #83, opened 2023-11-02
fix(genesys-spark): ensure asset loading is done on the same domain for new regions

MattCheely:
I had hoped to find a way to share this code with what's in `registerCustomElements` in the components package, but I couldn't find a good way to do it.

### DIFF
--- a/packages/genesys-spark/src/hosts.ts
+++ b/packages/genesys-spark/src/hosts.ts
@@ -7,23 +7,16 @@ declare global {
 // Default domain to load assets from
 const DEFAULT_DOMAIN = 'mypurecloud.com';
 
-// List of Genesys UI domains
-const DOMAIN_LIST = [
-  'apne2.pure.cloud',
-  'aps1.pure.cloud',
-  'cac1.pure.cloud',
-  'euw2.pure.cloud',
+// List of Genesys UI domains that do not follow the ${region}.pure.cloud format
+const NON_STANDARD_DOMAINS = [
   'inindca.com',
   'inintca.com',
   'mypurecloud.com.au',
   'mypurecloud.com',
   'mypurecloud.de',
   'mypurecloud.ie',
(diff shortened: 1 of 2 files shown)


packages/genesys-spark/src/hosts.ts
```diff
@@ -51,5 +44,14 @@ export function getFontOrigin(): string {
 
 function getRegionDomain() {
   const pageHost = window.location.hostname;
-  return DOMAIN_LIST.find(regionDomain => pageHost.endsWith(regionDomain));
+
+  // We can automatically handle the standard domain format: ${region}.pure.cloud
+  if (pageHost.endsWith('.pure.cloud')) {
+    return pageHost.split('.').slice(-3).join('.');
```
jordanstith15, line 50: I'm a bit confused by this. From my understanding, if the pageHost is something like 'aps1.pure.cloud', wouldn't this just also return 'aps1.pure.cloud' in a very roundabout way? Is there a case where there could be more than 2 periods? I don't see a current existing domain from the ones you deleted that do.

Also the `pageHost.endsWith` will not match on of the domains deleted above: `use2.maximus-pure.cloud`

jordanstith15 (changes requested):

packages/genesys-spark/src/hosts.ts
MattCheely: use2.maximus-pure.cloud is no longer in use, so it doesn't need a replacement. I should have called that out in the PR.

`pageHost` will normally be 4 segments long. Our apps are hosted on a subdomain, eg. `apps.aps1.pure.cloud` or `app.aps.pure.cloud`. Some projects could create new sub-domains as well. We want all of these and as many future variations as possible to resolve back to just the region portion - the last 3 segments.

jordanstith15 (approved):

PR #87, opened 2023-11-06
feat(genesys-spark): add date/time formatting utilities

MattCheely:
This adds date/time formatting functions that are proxies to the native Intl constructors, but can handle locale lookup on their own from our standard locale locations

COMUI-2431

### DIFF
--- a/packages/genesys-spark/src/intl.ts
+++ b/packages/genesys-spark/src/intl.ts
@@ -1,0 +1,88 @@
+/**
+ * Provides an [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat)
+ * object for formatting dates and times. Unlike the native version, `locale` is
+ * an optional argument. If not provided, the function will try to determine the
+ * locale from the DOM, where it should be set for a11y reasons.
+ * @param locale optional locale to use for formatting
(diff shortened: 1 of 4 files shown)


MattCheely:
I was planning work for the upcoming PURE epic and figured out a way to get correct region behavior 99% of the time, started to write it up and then realized it would take me about as long to code it as to explain it, so now there's a PR.

packages/genesys-spark/src/intl.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+
+/**
+ * Returns true if the provided locale only has a language, but the user's
+ * browser settings have the same language with a locale.
+ * @param localeString The locale to check
+ * @returns true if the region can be guessed from the browser settings.
+ */
+function browserHasRegionData(localeString: string): boolean {
```
MattCheely, line 83: This is the key idea that lets us get correct region formatting mostly for free. If we look for the Genesys locale and it has no region, but the browser has the same language setting with a region, then we can use the browser language setting. This will only go wrong when the user's browser is set to their preferred language, but with a region they don't want. So an en-US user who has selected `en` will only not see their preferred date format if their computer is set to to some locale other than `en-US` (browsers default 'en' to the more common international format of dd/mm/yyyy)

katie-bobbe-genesys (approved):

PR #102, opened 2023-11-08
fix(gux-listbox): fix keyboard accessibility issue for list boxes wit…

caitlanspronk:
Fixes an accessibility issue for listboxes that contain dividers. When identifying the next or previous option in the list to select on arrow down or up, gux-listbox was selecting the next element sibling. If the list contained a divider or heading, it selected that element, preventing further arrow navigation to the rest of the list items. 

See gux-time-zone-picker for a working example.

This fix is somewhat similar to the solution in place for gux-list, in that it identifies a set of valid selectable tag names. 
[ENGAGEUI-7811]

[ENGAGEUI-7811]: https://inindca.atlassian.net/browse/ENGAGEUI-7811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-listbox/gux-listbox.service.ts
+++ b/packages/genesys-spark-components/src/components/stable/gux-listbox/gux-listbox.service.ts
@@ -1,5 +1,7 @@
 import { ListboxOptionElement } from './options/option-types';
 
+const validOptionTagNames = ['GUX-OPTION', 'GUX-OPTION-ICON'];
+
 export function getListOptions(
   list: HTMLGuxListboxElement
 ): ListboxOptionElement[] {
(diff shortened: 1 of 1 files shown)


packages/genesys-spark-components/src/components/stable/gux-listbox/gux-listbox.service.ts
```diff
@@ -26,14 +28,21 @@ function getActiveOption(list: HTMLGuxListboxElement): ListboxOptionElement {
   return getListOptions(list).find(option => option.active);
 }
 
+function isNonSelectableOption(option: HTMLElement): boolean {
+  return !validOptionTagNames.includes(option.tagName);
+}
+
```
daragh-king-genesys, line 34: I think this would be simpler if you use an `isSelectableOption` function and include the disabled and filtered check in it.

```javascript
function isSelectableOption(option: HTMLElement): boolean {
  return validOptionTagNames.includes(option.tagName) && (!option.disabled || !option.filtered);
}
```

Then you would have calls like 

```javascript
const active = (option === element) && isSelectableOption(option) ;
```

and 

```javascript
while (firstOption && !isSelectableOption(firstOption)) {
  firstOption = firstOption.nextElementSibling as ListboxOptionElement;
}
```

caitlanspronk: Good idea--I've updated to this with the modification that the option should be both enabled and non-filtered to be selectable, which I think was just an error in your quick example.

daragh-king-genesys: 🙈

daragh-king-genesys (approved):

caitlanspronk:
@daragh-king-genesys Thanks for approving and merging!  Should I also fix this in v4? This will be an issue there too.

PR #111, opened 2023-11-09
docs(custom-properties): removed custome properties docs from components

daragh-king-genesys:
removed custome properties docs from components


daragh-king-genesys:
I think our manually maintained docs approach used in some components is a bad idea. Once v4 is released we can look into auto generating it during our build step.

katie-bobbe-genesys (approved):

jason-evans-genesys (approved):

PR #113, opened 2023-11-09
feat(gux-popup): add offset property to gux-popup

thomas-c-dillon:
COMUI-2479

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-popup/gux-popup.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-popup/gux-popup.tsx
@@ -40,6 +40,12 @@ export class GuxPopup {
   disabled: boolean = false;
 
   /**
+   * Number of pixels the popup is offset from the target.
+   */
+  @Prop()
+  offset: number = 2;
+
+  /**
    * set if parent component design allows for popup exceeding target width
    */
   @Prop()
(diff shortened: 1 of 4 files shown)


packages/genesys-spark-components/src/components/stable/gux-popup/gux-popup.tsx
```diff
@@ -39,6 +39,12 @@ export class GuxPopup {
   @Prop()
   disabled: boolean = false;
 
+  /**
+   * Number of pixels the popup is offset from the target.
+   */
+  @Prop()
+  offset: number = 2;
```
daragh-king-genesys, line 46: This seems like we are adding a prop to allow for design inconsistencies. Did design confirm they wanted the context button to be 4px and everything that is similar should stay at 2px.

thomas-c-dillon: I agree 100%. I tried to ask about this and the designer I was working with responded with, "If we put 2px it won’t look as good as. Could we keep the 4px as it’s it is in the design?"

That doesn't seem like a great reason to me but I acknowledge I don't have any eye for visual design. I can only tell the difference when seeing them side by side.

daragh-king-genesys (approved):

daragh-king-genesys:
you will have to rebase this on main. It should be something like this.
```
git checkout main
git fetch --prune
git pull
git checkout feature/COMUI-2479
git rebase main
git push --force
```

PR #120, opened 2023-11-10
refactor(dropdown,dropdown-multi,listbox): fixing focus in Firefox

jordanstith15:
Closes [COMUI-2465](https://inindca.atlassian.net/browse/COMUI-2465)

Firefox apparently doesn't support :host-context.
Using :has anywhere breaks the whole selector, including ones separated by a comma.

[COMUI-2465]: https://inindca.atlassian.net/browse/COMUI-2465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi.scss
@@ -142,8 +142,7 @@
     var(--gse-ui-formControl-input-active-border-color);
   border-radius: var(--gse-ui-formControl-input-borderRadius);
 
-  &:focus-visible,
-  &:focus-within:has(:focus-visible) {
+  &:focus-visible {
     @include focus.gux-focus-ring;
   }
 
@@ -178,8 +177,7 @@
     var(--gse-ui-formControl-input-default-border-color);
   border-radius: var(--gse-ui-formControl-input-borderRadius);
 
-  &:focus-visible,
-  &:focus-within:has(:focus-visible) {
+  &:focus-visible {
     @include focus.gux-focus-ring;
   }
 }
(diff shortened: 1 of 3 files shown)


packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi.scss
```diff
@@ -142,8 +142,7 @@
     var(--gse-ui-formControl-input-active-border-color);
   border-radius: var(--gse-ui-formControl-input-borderRadius);
 
-  &:focus-visible,
-  &:focus-within:has(:focus-visible) {
+  &:focus-visible {
```
daragh-king-genesys, line 145: Would something like 
```
  &:focus-visible {
    @include focus.gux-focus-ring;
  }

  &:focus-within:has(:focus-visible) {
    @include focus.gux-focus-ring;
  }
```
be a smaller change?

I think we would not be changing anything in chrome at all and the failing firefox section would not affect the ` &:focus-visible` code anymore.

jordanstith15: Yeah I think that would work. Though I tested all 3 browsers after I removed it entirely and didn't see any difference in Safari and Chrome

daragh-king-genesys: 🤷 maybe it is just redundant code so. I just just looking at the PR in isolation and trying to make the minimum change possible. Whichever approach you want to go with I am fine with if it does not change any behaviours in non firefox browsers.

jordanstith15: I did it your way just in case I missed something

daragh-king-genesys (approved):

PR #123, opened 2023-11-13
fix(tabs): prevent focus of disabled tabs

gavin-everett-genesys:
prevent focus of disabled tabs

[COMUI-2485](https://inindca.atlassian.net/browse/COMUI-2485)

[COMUI-2485]: https://inindca.atlassian.net/browse/COMUI-2485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-tabs/gux-tab/gux-tab.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-tabs/gux-tab/gux-tab.tsx
@@ -86,6 +86,7 @@ export class GuxTab {
           'gux-tab': true,
           'gux-active': this.active
         }}
+        disabled={this.guxDisabled}
         type="button"
         aria-disabled={this.guxDisabled.toString()}
         id={`gux-${this.tabId}-tab`}


packages/genesys-spark-components/src/components/stable/gux-tabs/gux-tab/gux-tab.tsx
```diff
@@ -86,6 +86,7 @@ export class GuxTab {
           'gux-tab': true,
           'gux-active': this.active
         }}
+        disabled={this.guxDisabled}
         type="button"
         aria-disabled={this.guxDisabled.toString()}
```
daragh-king-genesys, line 91: Because of this attribute I think not disabling the button was a choice we made before. Maybe this is not a bug.

katie-bobbe-genesys: I think it is intentional to be able to use the arrow keys to focus disabled tabs. For example, truncated text in tabs is shown on focus with the keyboard. Without being able to focus disabled tabs, truncated text can not be viewed by keyboard only users

katie-bobbe-genesys (changes requested):
not a defect

packages/genesys-spark-components/src/components/stable/gux-tabs/gux-tab/gux-tab.tsx
gavin-everett-genesys: Makes sense 👍  The only other thing I am wondering is should the the `gux-tooltip` not have disabled stylings ? In `v3` it does not have disabled stylings. I can create an issue for this if so.

PR #124, opened 2023-11-13
fix(time-picker): fixed focus styles

daragh-king-genesys:
fixed focus styles

COMUI-2488


thomas-c-dillon (approved):
Looks like focus rings to me!

PR #132, opened 2023-11-14
GDS-2213 action button chevron position

conor-darcy:
Centering the icon vertically.


jordanstith15:
We are merging v4 stuff into main now. beta/v4 is just used for deploying a timemachine of it for now and is updated manually by Daragh

PR #150, opened 2023-11-17
chore(z-index): default values

katie-bobbe-genesys:


packages/genesys-spark-tokens/data/tokens.json
```diff
@@ -3508,6 +3508,40 @@
...
+      },
+      "500" : {
+        "value": "500",
+        "type": "zindex"
+      },
+      "600" : {
+        "value": "600",
+        "type": "zindex"
+      }
+    },
```
daragh-king-genesys, line 3544: @conor-darcy does the figma tokens plugin sync both ways or how can we add tokens from dev to UX?

daragh-king-genesys (approved):
You will have to duplicate your `packages/genesys-spark-tokens/data/tokens.json`changes into the design-token branch so UX picks them up. I had a chat with Conor about getting them to work off `main` but we have a few issues to solve first.

PR #151, opened 2023-11-20
feat(time-zone-picker) - ux review comments

gavin-everett-genesys:
Updates made in the PR : 

- Removed brackets from `subtexts` eg previously was `(Local)` now becomes `Local`.
- Removed underscores from city names in the dropdown.
- Push all timezone-options with are just `Etc/GMT` to the bottom of the list.
- You can now search the list on any word from a timezone options eg `United States` will return all united states options.
- The format for the timezone options has also changed to `City, Country(Offset)`.
- Within the `Default Zones` dropdowns I have set the `height` to `auto` as it was requested to have the subtext underneath the value which would have been restrictive as the `height` is set to `32px` on the `gux-option`


packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
```diff
@@ -112,11 +130,25 @@ export class GuxTimeZonePickerBeta {
...
+      } else if (
+        !a.displayTextNameFormatted.startsWith('Etc/GMT') &&
+        b.displayTextNameFormatted.startsWith('Etc/GMT')
+      ) {
+        return -1;
+      } else {
+        return a.displayTextNameFormatted?.localeCompare(
+          b.displayTextNameFormatted
+        );
+      }
```
gavin-everett-genesys, line 151: This is sorting the list pushing all `time-zone-options` which start with `Etc/GMT` to the bottom of the list. If someone has a cleaner way let me know

daragh-king-genesys (approved):

PR #162, opened 2023-11-21
fix(gux-calendar-beta): handling disabled date state

jason-evans-genesys:
https://inindca.atlassian.net/browse/COMUI-2459


packages/genesys-spark-components/src/components/beta/gux-calendar-beta/components/single/gux-calendar.tsx
```diff
@@ -86,6 +86,17 @@ export class GuxCalendar {
...
+      this.root.shadowRoot.querySelector(
+        `.gux-content-date[data-date="${date.getTime()}"]`
+      );
+    const isCurrentDateDisabled = Array.from(
+      currentDateElement.classList
+    ).includes('gux-disabled');
+    if (isCurrentDateDisabled) {
+      return;
+    }
+
```
daragh-king-genesys, line 99: This seems like a very round about way to check this. We have a `date: Date` parameter and `minValue: Date` and `maxValue: Date` properties so we should be able to check it if is a valid date or not directly.

In the `getMonthDays` function we check this to get the disabled property already.
```
 const disabled =
        (this.minValue && currentDate.getTime() <= this.minValue.getTime()) ||
        (this.maxValue && currentDate.getTime() > this.maxValue.getTime());
```
We can move this into a function for reuse. I generally prefer "positive" functions like `isValidDate` not `isInvalidDate` but either will obviously work.

jason-evans-genesys: Ah good call. Thanks for pointing this out.

packages/genesys-spark-components/src/components/beta/gux-calendar-beta/components/single/gux-calendar.tsx
```diff
@@ -188,10 +199,8 @@ export class GuxCalendar {
         };
       }
 
-      // Disable a date that is outside the defined date range boundaries
-      const disabled =
-        (this.minValue && currentDate.getTime() <= this.minValue.getTime()) ||
-        (this.maxValue && currentDate.getTime() > this.maxValue.getTime());
+      // Check if a date is outside the defined date range boundaries
+      const valid = this.isValidDate(currentDate);
```
daragh-king-genesys, line 203: Do not make this change from `disabled` to `valid` it makes the code harder to follow below for no benefit
Use something like
```
const disabled = !this.isValidDate(currentDate);
```
or
```
const disabled = this.isInvalidDate(currentDate);
```

packages/genesys-spark-components/src/components/beta/gux-calendar-beta/components/single/gux-calendar.tsx
```diff
@@ -161,6 +165,13 @@ export class GuxCalendar {
     }
   }
 
+  private isValidDate(date: Date): boolean {
+    return (
+      (this.minValue && date.getTime() <= this.minValue.getTime()) ||
+      (this.maxValue && date.getTime() > this.maxValue.getTime())
+    );
+  }
+
```
daragh-king-genesys, line 174: I think this is wrong. This will return `true` for an invalid date not a valid one.

jason-evans-genesys: good catch. let me rework the code.

daragh-king-genesys (approved):

PR #165, opened 2023-11-21
fix(tag): fixing gux-tooltip-title

jordanstith15:
Closes [COMUI-2109](https://inindca.atlassian.net/browse/COMUI-2109)

[COMUI-2109]: https://inindca.atlassian.net/browse/COMUI-2109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-tooltip-title/gux-tooltip-title.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-tooltip-title/gux-tooltip-title.tsx
@@ -120,6 +120,7 @@ export class GuxTooltipTitle {
         .querySelector('slot')
         .assignedNodes()
         .map(node => node.textContent)
+        .map(node => node.replace('\n', ''))
         .join('')
         .trim();
     } else if (this.root.querySelector('.gux-title-container')) {
--- a/packages/genesys-spark-components/src/components/stable/gux-tooltip-title/gux-tooltip-title.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-tooltip-title/gux-tooltip-title.tsx
@@ -9,7 +9,7 @@ import {
 } from '@stencil/core';
 import { logWarn } from '@utils/error/log-error';
 import { OnMutation } from '@utils/decorator/on-mutation';
-import { afterNextRenderTimeout } from '@utils/dom/after-next-render';
+import { OnResize } from '@utils/decorator/on-resize';
 
 @Component({
   styleUrl: 'gux-tooltip-title.scss',
@@ -73,11 +73,12 @@ export class GuxTooltipTitle {
   onMutation(): void {
     this.titleName = this.setTooltipTitleText();
     this.checkForTooltipHideOrShow();
+  }
 
-    // Handle checking if the tooltip should hide/show for dynamic content
-    afterNextRenderTimeout(() => {
-      this.checkForTooltipHideOrShow();
-    }, 500);
+  @OnResize()
+  onResize(): void {
+    this.titleName = this.setTooltipTitleText();
+    this.checkForTooltipHideOrShow();
   }
 
   componentWillLoad() {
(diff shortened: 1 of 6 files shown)


packages/genesys-spark-components/src/components/stable/gux-tooltip-title/gux-tooltip-title.tsx
```diff
@@ -73,11 +73,12 @@ export class GuxTooltipTitle {
...
 
-    // Handle checking if the tooltip should hide/show for dynamic content
-    afterNextRenderTimeout(() => {
-      this.checkForTooltipHideOrShow();
-    }, 500);
+  @OnResize()
+  onResize(): void {
+    this.titleName = this.setTooltipTitleText();
+    this.checkForTooltipHideOrShow();
   }
```
jordanstith15, line 82: Apparently a slot change doesn't always hit the OnMutation, but this case does hit the OnResize

packages/genesys-spark-components/src/components/stable/gux-tooltip-title/gux-tooltip-title.tsx
```diff
@@ -119,6 +120,7 @@ export class GuxTooltipTitle {
         .querySelector('slot')
         .assignedNodes()
         .map(node => node.textContent)
+        .map(node => node.replace('\n', ''))
```
jordanstith15, line 123: added because when the text is slotted in a span and the span is on a middle line, the mapping of textContent returns something like `['\n       ','Text','\n     ']`, which then breaks the tooltip

daragh-king-genesys (approved):

packages/genesys-spark-components/src/components/stable/gux-tooltip-title/gux-tooltip-title.tsx
MattCheely: Would it be simpler to just use `node.textContent.trim()` to remove all preceding & trailing whitespace?

jordanstith15: Yeah looks like that solves it without the extra map too.

I found another issue where it's not updating the tooltip if there is no resize and a wrapping element is slotted instead of just the text, because there doesn't seem to be a way to check if child nodes of a slot have been changed..

MattCheely: > I found another issue where it's not updating the tooltip if there is no resize and a wrapping element is slotted instead of just the text, because there doesn't seem to be a way to check if child nodes of a slot have been changed..

I'm ok with this limitation. I don't think we need to support complex structure inside most of these contexts.

MattCheely: For the original issue, there was also an item around the screenreader text not updating - do you plan to address that in this PR or a separate one? 

I'd also wonder if it's possible to write an E2E test for this that would render a badge with content that fits, verify there's no truncation, update the conent, and verify it truncates.

daragh-king-genesys (approved):

MattCheely (approved):

packages/genesys-spark-components/src/components/stable/gux-tooltip-title/gux-tooltip-title.tsx
jordanstith15: So I checked and maybe because of my fix or maybe because of something else, but it only happens when it is text inside a span that is slotted because we have no way to trigger a rerender when a child of the slotted element changes. The slotted element itself has to be replaced in the dom to trigger an onslotchange

MattCheely (approved):

PR #173, opened 2023-11-23
feat(badge): centred icon in badge

321gillian:
Centred icon in badge

GSD-2243

Does this need to be backported to v3?

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-badge/gux-badge.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-badge/gux-badge.scss
@@ -24,9 +24,12 @@
   gux-tooltip-title {
     white-space: nowrap;
 
-    ::slotted(gux-icon) {
-      height: 16px;
-      font-size: 16px;
+    .gux-title-container {
+      ::slotted(gux-icon) {
+        top: 1px;
+        height: 16px;
+        font-size: 16px;
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-badge/gux-badge.scss
```diff
@@ -24,9 +24,12 @@
   gux-tooltip-title {
```
daragh-king-genesys, line 24: The issue probably effects all components that use `gux-tooltip-title` so this fix should be made in that component.

321gillian: Hi Daragh, 

Thanks for flagging this. I made some changes to make the tooltip-title more consistent. Tooltip-title is used in the following components: tabs, table-toolbar-menu, tag and badge. The icon aspect of the tooltip title seems to be currently used in the tabs and badge components. The others use either the text alone or the icon alone. 

The height (16px) which was set in the Badge file as an amendment to the implementation of the tooltip-title (currently no explicit height set) matches the current Figma design for Badge and Tabs. I updated the rule in Tooltip-title to match 16px for height. Then to accommodate the larger icon, I changed the top: 2px rule to 1px as well. 

Thanks,

Gillian

daragh-king-genesys:
Yes, backport to v3

PR #178, opened 2023-11-27
feat(phone-input): set bg color to token

321gillian:
Set bg color to token

GSD-2251


thomas-c-dillon (approved):

daragh-king-genesys:
No, don't backport to v3. There were no tokens in v3 so anything only token related won't need to be backported.

PR #179, opened 2023-11-27
feat(form-field-radio): align input and label

321gillian:
Align input and label

GSD-2249

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/gux-form-field-checkbox.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/gux-form-field-checkbox.scss
@@ -55,7 +55,7 @@
   display: inline-grid;
   width: var(--gse-ui-checkbox-icon-width);
   height: var(--gse-ui-checkbox-icon-height);
-  margin: 1px 0 0;
+  margin: 0;
   text-align: center;
   vertical-align: middle;
   -webkit-appearance: none;
@@ -120,10 +120,4 @@
   line-height: var(--gse-ui-checkbox-label-text-lineHeight);
   color: var(--gse-ui-checkbox-label-foregroundColor);
   vertical-align: middle;
-}
-
-.gux-input {
-  display: inline-grid;
-  grid-template-rows: var(--gse-ui-checkbox-label-text-lineHeight) auto;
-  align-items: center;
 }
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-radio/gux-form-field-radio.scss
```diff
@@ -126,3 +126,8 @@
   color: var(--gse-ui-radioButton-label-foregroundColor);
   vertical-align: middle;
 }
+
+.gux-input {
```
jason-evans-genesys, line 130: Looks like this css change is causing the error gux-input element to shift down too much. 
Screen Shot 2023-11-27 at 1 34 30 PM

daragh-king-genesys: This may be a more structural issue with how the components internal structure is composed.

321gillian: Hi Daragh, 

I just want to check. Do you mean by that I should look at restructuring the HTML rather than just changing the styling?

daragh-king-genesys: Maybe, if you can't get the design to work or if restructuring the HTML would allow you to implement the design in a simpler way. The components have evolved over time. They started out with just the radio button/checkbox and label. Then an error message was added and then a help message was added. Basically if you think the css is getting overly complicated (or just not working) because of structural issues make a structural change.

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/gux-form-field-checkbox.scss
```diff
@@ -55,7 +55,7 @@
   display: inline-grid;
   width: var(--gse-ui-checkbox-icon-width);
   height: var(--gse-ui-checkbox-icon-height);
-  margin: 0;
+  margin: 1px 0 0;
```
daragh-king-genesys, line 58: This change seems a bit hacky and does not make a big difference in my opinion. I would remove it.

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-radio/gux-form-field-radio.scss
321gillian: Ok, I went a different direction with the change. Kept it very light touch. No changes were required on the checkbox side

daragh-king-genesys (approved):

daragh-king-genesys:
No need to backport

PR #183, opened 2023-11-28
feat(dropdown-multi): make option svg 16 by 16

321gillian:
Make option svg 16 by 16

GDS-2239

Does this need to be backported?

Also, this is hacky so if anyone knows a better option here, happy to try it.


321gillian:
> @321gillian I think the following may work,
> 
> ```
> min-width: var(--gse-ui-menu-option-startIcon-width);
> max-width: var(--gse-ui-menu-option-startIcon-width);    
> height: var(--gse-ui-menu-option-startIcon-height);
> ```
> 
> Also this does not need to be backported :)

That did the trick, thanks Gavin!

daragh-king-genesys (approved):

gavin-everett-genesys (approved):

daragh-king-genesys:
Yes, backport to v3

PR #185, opened 2023-11-28
feat(rating): added new property to customize increment

daragh-king-genesys:
added new property to customize increment

COMUI-2419


docs/migration-guides/v4/readme.md
```diff
@@ -388,6 +389,19 @@ If possible, avoid the usage of legacy components that have a migration path and
 - These changes should also make it clearer to application teams if the icons they are using are official Spark icons or not.
 - The process for adding Official Spark icons has been streamlined and you should contact the UX Design System Team if you require an official replacement for a legacy icon used in your application
 
+#### gux-rating
+
+- **_New Prop_**: A new `increment` property has been added to `gux-rating`. This property can be set to 0.5 or 1 (default). In V3 0.5 was the default so if needed you must add `increment="0.5"` to your rating component for it to behave in an identical manor.
```
jason-evans-genesys, line 394: Was it intended to say `manner` instead of `manor`?

packages/genesys-spark-components/src/components/stable/gux-rating/example.html
```diff
@@ -8,11 +8,19 @@ <h2>Defaults</h2>
...
+  aria-label="feedback"
+  increment="1"
+  onchange="notify(event)"
+  oninput="notify(event)"
+>
+</gux-rating>
+
 <h2>Set initial value</h2>
-<label for="rating">Feedback</label><br />
+<label id="rating">Feedback</label><br />
```
jason-evans-genesys, line 21: Do we need an `id` here if it's not used for anything?

packages/genesys-spark-components/src/components/stable/gux-rating/tests/__snapshots__/gux-rating.spec.ts.snap
```diff
@@ -169,6 +169,90 @@ exports[`gux-rating #render should render component as expected (12) 1`] = `
...
+      <gux-icon decorative="" icon-name="rating"></gux-icon>
+      <gux-icon decorative="" icon-name="rating"></gux-icon>
+      <gux-icon decorative="" icon-name="rating"></gux-icon>
+    </div>
+  </mock:shadow-root>
+</gux-rating>
+`;
+
+exports[`gux-rating #render should render component as expected (15) 1`] = `
+<gux-rating 1"="" aria-label="Feedback" aria-readonly="false" aria-valuemax="5" aria-valuemin="0" aria-valuenow="2" role="spinbutton" tabindex="0" value="2 increment=">
```
jason-evans-genesys, line 200: `1"=""` looks off here

packages/genesys-spark-components/src/components/stable/gux-rating/example.html
```diff
@@ -8,11 +8,19 @@ <h2>Defaults</h2>
 >
 </gux-rating>
 
+<h2>Integers only</h2>
```
thomas-c-dillon, line 11: If the default is integers only, wouldn't it be better to give an example of half step increment instead?

packages/genesys-spark-components/src/components/stable/gux-rating/gux-rating.tsx
```diff
@@ -28,6 +28,9 @@ export class GuxRating {
   @Prop()
   readonly: boolean = false;
 
+  @Prop()
+  increment: 0.5 | 1 = 1;
```
thomas-c-dillon, line 32: This attribute/property can be set to any number. If you really want to support only these two options, it might be best to move to two specific string values instead (default and halfStep?).

packages/genesys-spark-components/src/components/stable/gux-rating/example.html
daragh-king-genesys: Used by the `aria-labelledby` below for accessibility.

packages/genesys-spark-components/src/components/stable/gux-rating/gux-rating.tsx
daragh-king-genesys: What do you mean "can be set to any number"? You will get a type error if you set it to something other than `0.5` or `1`.

thomas-c-dillon: The TS types don't really exist once this gets transpiled to JS. JS only has number, so that is what we get. I tested both the attribute and property in the demo and was able to set it to a variety of other numbers (.25. 1.5, 2, 3). I can even set it to `steve` without getting a type error (I think it is just getting coerced to `0` based on behavior).

packages/genesys-spark-components/src/components/stable/gux-rating/tests/__snapshots__/gux-rating.spec.ts.snap
daragh-king-genesys: fixed

packages/genesys-spark-components/src/components/stable/gux-rating/gux-rating.tsx
katie-bobbe-genesys: I agree with Thomas, I think it might would be better to set the increment as a string. `default` and `halfStep` sound good to me

daragh-king-genesys: Done

thomas-c-dillon (approved):

jason-evans-genesys (approved):

PR #194, opened 2023-11-30
feat(dropdown-multi): increase checkbox size to 16px

321gillian:
Increase checkbox size to 16px

GDS-2239


daragh-king-genesys:
I was wrong this should not be backported

PR #202, opened 2023-12-01
feat(datepicker): update datepicker focus style for calendar

321gillian:
Update datepicker focus style for calendar

GDS-2248

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-datepicker/gux-datepicker.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-datepicker/gux-datepicker.scss
@@ -180,7 +180,9 @@
             @include gse-ui-focus;
 
             border: none;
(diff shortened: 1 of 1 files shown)


packages/genesys-spark-components/src/components/stable/gux-datepicker/gux-datepicker.scss
```diff
@@ -173,7 +179,9 @@
           &:focus-within {
             @include gse-ui-focus;
 
-            padding: 2px; // TODO: COMUI-2299 - token needed?
+            border: none;
+            border-radius: 2px;
+            outline-offset: 3px;
```
daragh-king-genesys, line 184: We should not have hardcoded px values. Everything should be tokens.

MattCheely (approved):

PR #228, opened 2023-12-07
fix(popover-list): improved focus handling and added events

daragh-king-genesys:
improved focus handling and added events

COMUI-1576

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
@@ -5,7 +5,9 @@ import {
   EventEmitter,
   h,
   JSX,
-  Prop
+  Listen,
+  Prop,
+  Watch
 } from '@stencil/core';
 import {
   autoUpdate,
@@ -19,6 +21,7 @@ import {
 
 import { OnClickOutside } from '@utils/decorator/on-click-outside';
 import { trackComponent } from '@utils/tracking/usage';
+import { afterNextRender } from '@utils/dom/after-next-render';
 import { findElementById } from '@utils/dom/find-element-by-id';
 
 /**
@@ -28,7 +31,9 @@ import { findElementById } from '@utils/dom/find-element-by-id';
 @Component({
   styleUrl: 'gux-popover-list.scss',
   tag: 'gux-popover-list',
-  shadow: true
+  shadow: {
+    delegatesFocus: true
+  }
 })
 export class GuxPopoverList {
   private popupElement: HTMLDivElement;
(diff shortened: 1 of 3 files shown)


packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
```diff
@@ -74,6 +79,40 @@ export class GuxPopoverList {
...
+  @Event()
+  guxopen: EventEmitter<void>;
+
+  /**
+   * Fired when the popover is close
+   */
+  @Event()
+  guxclose: EventEmitter<void>;
+
+  @Watch('isOpen')
```
gavin-everett-genesys, line 94: Will we need to check this also when the component first loads if the `is-open` attribute is already set to `true`.

packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
```diff
@@ -74,6 +79,40 @@ export class GuxPopoverList {
...
+   * Fired when the popover is open
+   */
+  @Event()
+  guxopen: EventEmitter<void>;
+
+  /**
+   * Fired when the popover is close
+   */
+  @Event()
+  guxclose: EventEmitter<void>;
```
gavin-everett-genesys, line 92: I was just wondering is there any difference between this event and the `guxdismiss` event.

packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
```diff
@@ -74,6 +79,40 @@ export class GuxPopoverList {
...
+
+  @Watch('isOpen')
+  onIsOpenChange(isOpen: boolean): void {
+    if (isOpen) {
+      this.guxopen.emit();
+
+      console.log('Focus', this.root);
```
katie-bobbe-genesys, line 99: remove console.log

packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
```diff
@@ -74,6 +79,40 @@ export class GuxPopoverList {
...
+      afterNextRender(() => {
+        this.root.focus();
```
katie-bobbe-genesys, line 101: I think I would expect the first `gux-list-item` to get focus, not the `gux-list` itself

packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
```diff
@@ -74,6 +79,40 @@ export class GuxPopoverList {
...
+    } else {
+      this.guxclose.emit();
+    }
+  }
+
+  @Listen('keydown')
+  onKeyDown(event: KeyboardEvent): void {
+    switch (event.key) {
+      case 'Escape':
+        this.dismiss();
```
katie-bobbe-genesys, line 112: Focus is lost after pressing `esc`. Normally we would expect the focusable 'target' element to gain focus after the popover is closed. For an example, see the focus behavior in `gux-modal`

daragh-king-genesys: Might have to go to UX with this as we currently have no guidance for how a popover should be triggered. In the example I went with on `mouseover` and on `focus`. That means that if I focus the target element on dismiss it will just open it again.

packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
daragh-king-genesys: I don't think so
The event is for changes.

packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
daragh-king-genesys: There are other issues with that so I don't think this PR is a good place to address this.

packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
daragh-king-genesys: `guxdismiss` is cancelable

gavin-everett-genesys:
In firefox the focus does not move to the `popover-list` when tabbing to `Example Element`.

Also I was wondering should a `gux-list-item` have the same sort of selected stylings as `gux-option` as currently there is no indication if the `gux-list-item` has been selected except for the event emitting.

PR #229, opened 2023-12-08
fix(calendar): fixed selected date on previouse month css

daragh-king-genesys:
fixed selected date on previouse month css

COMUI-2538


daragh-king-genesys:
[This bug](https://inindca.atlassian.net/browse/COMUI-2538) was not present in v4 because of our design changes so I fixed it in v3 to match our current v4 approach.

gavin-everett-genesys (approved):

PR #230, opened 2023-12-10
chore: update @types/new-relic-browser dependency

MattCheely:
This was flagged to do in a ticket. Only needed some minor adjustments.

COMUI-1257

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3962,9 +3962,9 @@
       "dev": true
     },
     "node_modules/@types/new-relic-browser": {
-      "version": "0.1118.2",
-      "resolved": "https://registry.npmjs.org/@types/new-relic-browser/-/new-relic-browser-0.1118.2.tgz",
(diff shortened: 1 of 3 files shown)


MattCheely:
🤦 Of course everything builds fine, but fails with Jest.

jordanstith15:
> 🤦 Of course everything builds fine, but fails with Jest.

Is the stuff with the CI failure still being worked on?

MattCheely:
> > 🤦 Of course everything builds fine, but fails with Jest.
> 
> Is the stuff with the CI failure still being worked on?

This fell off my radar over the holiday. It was intended to be a quick fix. I'll circle back to it tomorrow morning.

jason-evans-genesys (approved):

jason-evans-genesys (changes requested):
Looks good, just seeing some failing tests, but other than that all good to go.

MattCheely:
I've hit a whole new set of issues trying to come back to fix the problem with stubbing newrelic in Jest, and it's just not worth the effort for a type upgrade that wasn't causing any problems in the first place.

PR #239, opened 2023-12-11
fix(popover): console errors if target element removed

gavin-everett-genesys:
console errors if target element removed first

to be backported

[COMUI-2416](https://inindca.atlassian.net/browse/COMUI-2416)

[COMUI-2416]: https://inindca.atlassian.net/browse/COMUI-2416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
@@ -118,7 +118,7 @@ export class GuxPopover {
     // This is 13 because this makes the arrow look aligned
     const arrowLen = 13;
 
-    if (this.popupElement) {
+    if (this.popupElement && forElement) {
       void computePosition(forElement, this.popupElement, {
         placement: this.position,
         middleware: [


packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
```diff
@@ -114,11 +115,14 @@ export class GuxPopover {
...
 
   private updatePosition(): void {
-    const forElement = findElementById(this.root, this.for);
     // This is 13 because this makes the arrow look aligned
     const arrowLen = 13;
 
-    if (this.popupElement) {
+    if (
+      this.popupElement &&
+      getContainingDocument(this.root).getElementById(this.for)
```
thomas-c-dillon, line 123: If I am understanding the purpose of the `findElementById` utility, won't this fail if the target element is in a shadow DOM? Is there a reason to not just check if `forElement` exists instead?

packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
```diff
@@ -118,10 +117,7 @@ export class GuxPopover {
     // This is 13 because this makes the arrow look aligned
     const arrowLen = 13;
 
-    if (
-      this.popupElement &&
-      getContainingDocument(this.root).getElementById(this.for)
-    ) {
+    if (this.popupElement && findElementById(this.root, this.for)) {
```
thomas-c-dillon, line 120: I recommend doing this operation once and checking if it was successful instead of repeating it in the for loop. Also, it looks like more or less the same code is in `gux-popover-list`.

gavin-everett-genesys: I dont think executing this just once would work as there would be cases where its dynamically removed?

thomas-c-dillon: What I mean is, put it in a variable immediately before the if, then check the variable. Only synchronous logic so there shouldn't be any problem with it being removed.

```
  const forElement = findElementById(this.root, this.for);
  // This is 13 because this makes the arrow look aligned
  const arrowLen = 13;

  if (this.popupElement && forElement) {
```

thomas-c-dillon (approved):

PR #253, opened 2023-12-18
feat(listbox): first item does not receive focus automatically on filter

321gillian:
First item does not receive focus automatically on filter

GDS-2271

Does this need to be backported to v3?


daragh-king-genesys (approved):
No need to backport

PR #254, opened 2023-12-19
[WIP] feat(option): add subtext to option and list option groups

321gillian:
As part of the change to menu component in GDS-2287, there's a request to display a subtext on a menu option. I've just put in a bare suggestion here of how I would implement this subtext. Feedback and suggestions welcome

Add subtext to option

Update: Now contains a rough example of option group as well. This approach would require maybe refactoring the listbox to accept gux-option-group as a child instead of just listboxoptionelements. Open to other ideas here

GDS-2287

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
@@ -7,12 +7,14 @@ import {
   Host,
   JSX,
   Prop,
+  State,
   Watch
 } from '@stencil/core';
 
 import { randomHTMLId } from '@utils/dom/random-html-id';
 import { buildI18nForComponent, GetI18nValue } from '../../../../i18n';
 import translationResources from './i18n/en.json';
+import { hasSlot } from '@utils/dom/has-slot';
 
 /**
  * @slot - text
@@ -57,6 +59,9 @@ export class GuxOptionMulti {
   @Event()
   internalselectcustomoption: EventEmitter<string>;
 
+  @State()
+  private hasSubtext: boolean = false;
+
   @Watch('selected')
   emitRemoveCustomOption() {
     if (!this.selected && this.custom) {
@@ -78,6 +83,7 @@ export class GuxOptionMulti {
     if (this.custom) {
       this.internalselectcustomoption.emit(this.value);
     }
+    this.hasSubtext = hasSlot(this.root, 'subtext');
   }
 
   // SVGs must be in DOM for tokenization to work
@@ -133,7 +139,9 @@ export class GuxOptionMulti {
           >
             <slot />
           </gux-truncate>
-          <p>{this.subtext}</p>
+          <span>
+            <slot name="subtext"></slot>
+          </span>
         </div>
       ) as JSX.Element;
     } else {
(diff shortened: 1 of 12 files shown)


packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
```diff
@@ -33,6 +33,9 @@ export class GuxOptionMulti {
   @Prop()
   value: string;
 
+  @Prop()
+  subtext: string;
```
daragh-king-genesys, line 37: We typically don't use props to get text into the dom of componentes because we have found that nearly every time we do we run into the case where devs want to include an icon or special markup etc... and this stops them doing it.
See can you make an optional `subtext` slot work.

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
```diff
@@ -118,6 +121,33 @@ export class GuxOptionMulti {
...
+    if (this.subtext) {
+      return (
+        <div class="gux-option-text">
+          <gux-truncate
+            gux-slot-container
+            ref={el => (this.truncateElement = el)}
+          >
+            <slot />
+          </gux-truncate>
+          <p>{this.subtext}</p>
```
daragh-king-genesys, line 136: Have you a reason for using `<p>` over `<div>` or `<span>` here? Using elements like `<p>` may leads to inconsistencies as they may have different default styles in browsers.

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
321gillian: Thanks, that's really handy to know. I've updated the branch to use a slot instead.

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
321gillian: No reason, it was just the first one that came to mind. I'll go with span instead.

packages/genesys-spark-components/src/components/stable/gux-listbox/options/option-defaults.scss
```diff
@@ -53,6 +53,16 @@
     outline-offset: -2px;
   }
 
+  :host(.gux-show-subtext) {
+    place-content: stretch flex-start;
+
+    .gux-option-text {
+      p {
```
jason-evans-genesys, line 60: Do we want to change this `p` tag selector to `span`?

packages/genesys-spark-components/src/components/stable/gux-listbox/gux-listbox.tsx
```diff
@@ -185,20 +185,24 @@ export class GuxListbox {
...
+      child => child.tagName === 'GUX-OPTION-GROUP'
+    );
+
+    if (!isGroupedList) {
+      this.listboxOptions = Array.from(
+        this.root.children
+      ) as ListboxOptionElement[];
+    } else {
+      const groups = Array.from(this.root.children);
+      groups.forEach(child => {
```
jason-evans-genesys, line 198: Is there any way to use `map` here instead of `forEach`? If not, no worries.

321gillian: Good point, thanks Jason!

PR #258, opened 2023-12-20
feat(form-field-text-like): add loading spinner

321gillian:
Add loading spinner

GDS-2094

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-text-like/example.html
+++ b/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-text-like/example.html
@@ -204,6 +204,14 @@
       <span slot="help">This is a help message</span>
     </gux-form-field-text-like>
   </fieldset>
+
+  <fieldset>
+    <legend>Loading</legend>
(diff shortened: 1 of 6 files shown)


thomas-c-dillon:
The only issue I see (and it is a pretty small one), at excessively small widths the loading indicator (and clear input button) kind of break. There should probably be a minimum width to display both of these.

gavin-everett-genesys (approved):

321gillian:
> The only issue I see (and it is a pretty small one), at excessively small widths the loading indicator (and clear input button) kind of break. There should probably be a minimum width to display both of these.

Hi Thomas, I'll follow this up with design and add a follow-on ticket for it. In the meantime, I'm going to move this along if that's ok?

thomas-c-dillon (approved):

thomas-c-dillon:
> > The only issue I see (and it is a pretty small one), at excessively small widths the loading indicator (and clear input button) kind of break. There should probably be a minimum width to display both of these.
> 
> Hi Thomas, I'll follow this up with design and add a follow-on ticket for it. In the meantime, I'm going to move this along if that's ok?

That seems reasonable to me.

PR #263, opened 2024-01-03
[COMUI-2290] feat(token): Add compact button padding

JackieYada:
Moving gse-ui-button-padding and gse-ui-button-paddingIconOnly to default
and creating compact version


thomas-c-dillon (approved):
The token json file looks good. I ran the script to update the other token files and renamed the tokens that were being used. I would appreciate a second set of eyes on the pr though!

PR #265, opened 2024-01-04
feat(dropdown listbox dropdown-multi): add subtext to options on dropdowns

321gillian:
Add subtext to options on dropdown options (option, option-icon and option-multi)

GDS-2307


packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
```diff
@@ -118,6 +124,35 @@ export class GuxOptionMulti {
...
+        <gux-truncate
+          gux-slot-container
+          ref={el => (this.truncateElement = el)}
+        >
+          <slot />
+        </gux-truncate>
+      ) as JSX.Element;
+    }
+  }
+
```
jordanstith15, line 155: Could this simplified? Perhaps to something like
```
 <div class="gux-option-text">
     <gux-truncate
            gux-slot-container
            ref={el => (this.truncateElement = el)}
     >
            <slot />
     </gux-truncate>
     <slot  class="gux-subtext" name="subtext"></slot>
 </div>
```

Also instead of the class on the slot, you can also use ::slotted in the css. Either way possibly removes the need for the if-statement for hasSubtext.
The div wrapper might be able to stay for both cases or possibly removed entirely and the styles be put on the host `gux-show-subtext` class depending on how it affects things with the renderCustomOptionInstructions (since its position is absolute, it _should_ be kicked out of the flexbox flow)

Example in the renderText:
```
[
     <gux-truncate
            gux-slot-container
            ref={el => (this.truncateElement = el)}
     >
            <slot />
     </gux-truncate>,
     <slot  class="gux-subtext" name="subtext"></slot>
] as JSX.Element;
```

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
```diff
@@ -118,6 +124,35 @@ export class GuxOptionMulti {
     }
   }
 
+  private renderText(): JSX.Element {
+    // The gux-slot-container attribute is used in gux-listbox-multi and gux-dropdown-multi as a selector to get the slotted gux-option-multi text.
+    // This attribute is required because we need to get the slotted text and exclude the screen reader text.
+    if (this.hasSubtext) {
+      return (
+        <div class="gux-option-text">
```
jordanstith15, line 132: is this supposed to be `gux-option-wrapper`? I don't see gux-option-text in the scss file

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option-icon/gux-option-icon.tsx
```diff
@@ -52,6 +54,9 @@ export class GuxOptionIcon {
   @Prop({ mutable: true })
   hovered: boolean = false;
 
+  @State()
+  private hasSubtext: boolean = false;
```
jordanstith15, line 58: same comment as above

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option-icon/gux-option-icon.tsx
```diff
@@ -82,6 +88,27 @@ export class GuxOptionIcon {
...
+      ) as JSX.Element;
+    } else {
+      return (
+        <gux-truncate ref={el => (this.truncateElement = el)}>
+          <slot />
+        </gux-truncate>
+      ) as JSX.Element;
+    }
+  }
+
```
jordanstith15, line 111: same as above comment

jordanstith15 (changes requested):

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
```diff
@@ -48,6 +50,9 @@ export class GuxOptionMulti {
   @Prop()
   custom: boolean = false;
 
+  @State()
+  private hasSubtext: boolean = false;
```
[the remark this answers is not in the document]
321gillian: Hi Jordan, 

I had a look at removing the state flag here. I ran into some issues with removing it. The class on subtext for the host doesn't seem to update without the @state flag included. Can you have a look and see if I've missed anything that would get around this?

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
321gillian: Done

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
321gillian: Good spot

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option-icon/gux-option-icon.tsx
321gillian: As above

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option-icon/gux-option-icon.tsx
```diff
@@ -74,6 +80,14 @@ export class GuxOptionIcon {
     this.root.id = this.root.id || randomHTMLId('gux-option-icon');
   }
 
+  componentWillRender(): void {
+    this.onSubtextChange();
```
jordanstith15, line 84: same as above.

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
```diff
@@ -77,6 +83,10 @@ export class GuxOptionMulti {
     }
   }
 
+  componentWillRender(): void {
+    this.hasSubtext = hasSlot(this.root, 'subtext');
+  }
```
jordanstith15, line 88: If hasSubtext is kept as a state variable, then this probably isn't a good idea. My suggestion to move it to componentWillRender was only if it stopped being a state variable. Otherwise it leads to a possible infinite render loop as changing a state variable calls the re-render process, which then calls this function possibly changing the state again.

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option/gux-option.tsx
```diff
@@ -44,6 +58,14 @@ export class GuxOption {
     this.root.id = this.root.id || randomHTMLId('gux-option');
   }
 
+  componentWillRender(): void {
+    this.onSubtextChange();
+  }
```
jordanstith15, line 63: same as above again

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option/gux-option.tsx
```diff
@@ -60,14 +82,18 @@ export class GuxOption {
...
         aria-disabled={this.disabled.toString()}
       >
-        <gux-truncate ref={el => (this.truncateElement = el)}>
-          <slot />
-        </gux-truncate>
+        <div class="gux-option-wrapper">
+          <gux-truncate ref={el => (this.truncateElement = el)}>
+            <slot />
+          </gux-truncate>
+          <slot onSlotchange={() => this.onSubtextChange()} name="subtext" />
```
jordanstith15, line 95: So I noticed the other gux-option types don't have this onSlotchange. Should they have this logic as well?

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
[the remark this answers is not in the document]
jordanstith15: No worries. I just wanted to double check I wasn't missing something. Looks good to me

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
321gillian: Thanks for flagging this. Updated now

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option/gux-option.tsx
321gillian: Updated

jordanstith15 (approved):

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
jordanstith15: You're on a Macbook, right? There is a built-in screenreader that, if it works there, you can be sure it will work elsewhere, but if it doesn't it could still work on other more common third-party screenreaders

daragh-king-genesys (approved):

PR #266, opened 2024-01-04
fix(gux-action-button): changed placement to bottom-end

jason-evans-genesys:
Added a `placement` property in `gux-popup` component and then passed `bottom-end` as the `placement` value to `gux-popup` from `gux-action-button` and `gux-button-multi`. This was done in order to right align the popup below these two components, instead of being left aligned before.

https://inindca.atlassian.net/browse/COMUI-2492

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-popup/gux-popup.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-popup/gux-popup.tsx
@@ -16,7 +16,8 @@ import {
   offset,
   size,
   shift,
-  hide
+  hide,
+  Placement
 } from '@floating-ui/dom';
 
 /**
@@ -32,6 +33,12 @@ export class GuxPopup {
   private targetElementContainer: HTMLElement;
   private popupElementContainer: HTMLElement;
   private cleanupUpdatePosition: ReturnType<typeof autoUpdate>;
+
+  /**
+   * Placement of the popup. Default is bottom-start
+   */
+  @Prop()
+  placement: Placement = 'bottom-start';
 
   @Prop()
   expanded: boolean = false;
@@ -85,7 +92,7 @@ export class GuxPopup {
         this.popupElementContainer,
         {
           strategy: 'fixed',
-          placement: 'bottom-start',
+          placement: this.placement,
           middleware: [
             offset(this.offset),
             flip(),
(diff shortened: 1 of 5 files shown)


packages/genesys-spark-components/src/components/stable/gux-popup/gux-popup.tsx
```diff
@@ -33,6 +34,12 @@ export class GuxPopup {
   private popupElementContainer: HTMLElement;
   private cleanupUpdatePosition: ReturnType<typeof autoUpdate>;
 
+  /**
+   * Placement of the tooltip. Default is bottom-start
```
gavin-everett-genesys, line 38: `tooltip` -> `popup`

packages/genesys-spark-components/src/components/stable/gux-popup/gux-popup.tsx
```diff
@@ -33,6 +34,12 @@ export class GuxPopup {
   private popupElementContainer: HTMLElement;
   private cleanupUpdatePosition: ReturnType<typeof autoUpdate>;
 
+  /**
+   * Placement of the popup. Default is bottom-start
+   */
+  @Prop({ mutable: true })
```
daragh-king-genesys, line 40: I don't think setting `mutable: true` here is correct as you are not changing the value of the property from inside the component. Look at the Stencil Docs https://stenciljs.com/docs/properties#prop-mutability-mutable

gavin-everett-genesys: I was thinking this also but then thought isnt the `placement` property dynamically updated within the component when the position changes. I am not sure does a dynamic update to a property within the component mean its mutable or not 🤔

jason-evans-genesys: @daragh-king-genesys good catch. I had copied the `placement` prop declaration from `gux-tooltip` and notice it's set as `mutable: true` in `gux-tooltip` as well. It doesn't look like either component is setting a new value of the property from inside the component like you mentioned. Mind if I also remove `mutable: true` for `gux-tooltip`?

jason-evans-genesys: @gavin-everett-genesys is this where you're referring to `placement` being dynamically changed inside `gux-popup`?

```
void computePosition(this.forElement, this.root, {
      placement: this.placement,
      strategy: 'fixed',
      middleware: middleware
    }).then(({ x, y, middlewareData }) => {
...
```

^^ I don't think `this.placement` is set a new property value, but correct me if I'm wrong here.

daragh-king-genesys (approved):

PR #267, opened 2024-01-05
feat(table) - add fixed columns to table

gavin-everett-genesys:
https://inindca.atlassian.net/browse/COMUI-1744

Confirmed with UX that the first column will only need to be fixed.


jordanstith15:
So the ticket mentions fixed columns and only the comments reference using a fixed first column. The final design handoff also seems to allude to other columns being fixed. Is the first column supposed to be the only one that will have the ability to be fixed?

jordanstith15 (approved):

packages/genesys-spark-components/src/components/stable/gux-table/gux-table.tsx
```diff
@@ -79,6 +79,12 @@ export class GuxTable {
   @Prop()
   emptyMessage: string;
 
+  /**
+   * Indicates if the tables first column is fixed and will stay in place even when scrolling horizontally.
+   */
+  @Prop()
+  fixedFirstColumn: boolean = false;
```
daragh-king-genesys, line 86: Not sure about this. What happens when the inevitable `fixedSecondColumn` is needed?
Can we discuss this at standup?

PR #268, opened 2024-01-05
[WIP] feat(option-group): Option Group Listbox and Dropdown Options

321gillian:
Feedback needed on API design of gux-option-group and how any input on tidying the implementation is also welcome.

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/example.html
+++ b/packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/example.html
@@ -457,7 +457,8 @@
     <gux-listbox-multi aria-label="Animals"> </gux-listbox-multi>
   </gux-dropdown-multi>
 
-  <h2>Grouped Options</h2>
+  <h2>Grouped</h2>
+  <h3>Standard Grouped Options</h3>
   <gux-dropdown-multi filter-type="starts-with" placeholder="Select an animal">
     <gux-listbox-multi aria-label="Animals">
       <gux-option-group label="Group 1">
@@ -473,6 +474,31 @@
       </gux-option-group>
       <gux-option-group label="Last Group">
         <gux-option-multi value="h">Horse</gux-option-multi>
+        <gux-option-multi value="i">Ibis</gux-option-multi>
+      </gux-option-group>
+    </gux-listbox-multi>
+  </gux-dropdown-multi>
+
+  <h3>Grouped & Filterable Options</h3>
+  <gux-dropdown-multi filter-type="starts-with" placeholder="Select an animal">
+    <gux-listbox-multi aria-label="Animals">
+      <gux-option-group label="First Group">
+        <gux-option-multi value="a" disabled>Ant</gux-option-multi>
+        <gux-option-multi value="b">Bat</gux-option-multi>
+      </gux-option-group>
+      <gux-option-group label="Second Group">
+        <gux-option-multi value="c">Cat</gux-option-multi>
+        <gux-option-multi value="d">Dog</gux-option-multi>
+      </gux-option-group>
+      <gux-option-group label="Three A Group">
+        <gux-option-multi value="e">Eel</gux-option-multi>
+        <gux-option-multi value="f">Frog</gux-option-multi>
+      </gux-option-group>
+      <gux-option-group label="Three B Group">
+        <gux-option-multi value="g">Goat</gux-option-multi>
+        <gux-option-multi value="Horse">Horse</gux-option-multi>
+      </gux-option-group>
+      <gux-option-group label="Fifth Group">
         <gux-option-multi value="i">Ibis</gux-option-multi>
       </gux-option-group>
     </gux-listbox-multi>
--- a/packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/example.html
+++ b/packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/example.html
@@ -456,6 +456,27 @@
   <gux-dropdown-multi id="loading-2" filter-type="starts-with" loading>
     <gux-listbox-multi aria-label="Animals"> </gux-listbox-multi>
   </gux-dropdown-multi>
+
+  <h2>Grouped Options</h2>
+  <gux-dropdown-multi>
+    <gux-listbox-multi aria-label="Animals">
+      <gux-option-group>
+        <span slot="title">Group 1</span>
+        <gux-option-multi value="a">Ant</gux-option-multi>
+        <gux-option-multi value="b">Bat</gux-option-multi>
+        <gux-option-multi value="c">Cat</gux-option-multi>
+        <gux-option-multi value="d">Dog</gux-option-multi>
+      </gux-option-group>
+      <gux-option-group>
+        <span slot="title">Group 2</span>
+        <gux-option-multi value="e">Eel</gux-option-multi>
+        <gux-option-multi value="f" disabled>Frog</gux-option-multi>
+        <gux-option-multi value="g">Goat</gux-option-multi>
+        <gux-option-multi value="h">Horse</gux-option-multi>
+        <gux-option-multi value="i">Ibis</gux-option-multi>
+      </gux-option-group>
+    </gux-listbox-multi>
+  </gux-dropdown-multi>
 </form>
 
 <style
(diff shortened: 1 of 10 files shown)


packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/example.html
```diff
@@ -456,6 +456,27 @@ <h3>Loading a filterable list</h3>
   <gux-dropdown-multi id="loading-2" filter-type="starts-with" loading>
     <gux-listbox-multi aria-label="Animals"> </gux-listbox-multi>
   </gux-dropdown-multi>
+
+  <h2>Grouped Options</h2>
+  <gux-dropdown-multi>
+    <gux-listbox-multi aria-label="Animals">
+      <gux-option-group>
```
daragh-king-genesys, line 463: The native HTML element (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup) this mimics uses a `label` property. Usually I prefer the slot approach but maybe in this case we should copy the native one. At the very least I think we should change the slot name to `label`.

packages/genesys-spark-components/src/components/stable/gux-listbox/option-group/gux-option-group.tsx
```diff
@@ -0,0 +1,54 @@
...
+  styleUrl: 'gux-option-group.scss',
+  tag: 'gux-option-group',
+  shadow: true
+})
+export class GuxOptionGroup {
+  @Element()
+  root: HTMLElement;
+
+  @Prop()
+  divider: boolean = true;
```
daragh-king-genesys, line 20: What are the use cases for excluding the divider? It seems odd to me that if it is part of the Design System it is optional.

packages/genesys-spark-components/src/components/stable/gux-listbox/option-group/gux-option-group.tsx
```diff
@@ -0,0 +1,54 @@
...
+ */
+@Component({
+  styleUrl: 'gux-option-group.scss',
+  tag: 'gux-option-group',
+  shadow: true
+})
+export class GuxOptionGroup {
+  @Element()
+  root: HTMLElement;
+
```
daragh-king-genesys, line 18: The native `optgroup` (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup) element has a disabled property we should implement.

packages/genesys-spark-components/src/components/stable/gux-listbox/option-group/gux-option-group.tsx
```diff
@@ -0,0 +1,54 @@
...
+  private renderDivider(): JSX.Element {
+    if (this.divider) {
+      return (<gux-list-divider></gux-list-divider>) as JSX.Element;
+    }
+  }
+
+  render(): JSX.Element {
+    return (
+      <Host role="group" class="gux-option-group">
```
daragh-king-genesys, line 47: The `gux-option-group` class is not used.

daragh-king-genesys, line 47: Is the `role` attribute on the right element? My first instinct is it should be on an element that just wraps the `<slot/>` element below.

packages/genesys-spark-components/src/components/stable/gux-listbox/option-group/gux-option-group.tsx
```diff
@@ -0,0 +1,54 @@
+import { Component, h, Host, JSX, Prop, State, Element } from '@stencil/core';
+import { hasSlot } from '@utils/dom/has-slot';
+
+/**
+ * The listbox component provides keyboard bindings and a11y patterns for selecting
+ * from a list of options.
+ *
+ * @slot - collection of elements conforming to the ListboxOptionElement interface
```
daragh-king-genesys, line 8: Docs for title slot missing.

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-listbox-multi.tsx
```diff
@@ -240,9 +241,27 @@ export class GuxListboxMulti {
...
+    });
+
+    const filteredOptions: HTMLGuxOptionMultiElement[] = options.filter(
+      element => element.tagName === 'GUX-OPTION-MULTI'
+    );
+    this.listboxOptions = [...filteredOptions];
+
+    const lastChild = listChildren[listChildren.length - 1];
+    if (lastChild && isOptionGroup(lastChild))
+      lastChild.setAttribute('divider', 'false');
```
daragh-king-genesys, line 263: We should see is it possible to do this with `:has-context` (https://developer.mozilla.org/en-US/docs/Web/CSS/:host-context) instead of in javascript.

Also I would use `lastChild.divider = false` instead of setting it via an attribute.

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-listbox-multi.tsx
```diff
@@ -240,9 +241,27 @@ export class GuxListboxMulti {
...
+      if (isOptionGroup(child)) {
+        const childOptions = Array.from(
+          child.children
+        ) as HTMLGuxOptionMultiElement[];
+        return options.push(...childOptions);
+      }
+      return options.push(child as HTMLGuxOptionMultiElement);
+    });
+
+    const filteredOptions: HTMLGuxOptionMultiElement[] = options.filter(
```
daragh-king-genesys, line 256: I move move this filter so that it is used to guard what is pushed into the options array instead of filtering the options array after adding everything.

Also is this the code that would be simplified if we used a `label` property instead of a named slot?

packages/genesys-spark-components/src/components/stable/gux-listbox/gux-listbox.service.ts
```diff
@@ -38,23 +54,53 @@ function setActiveOption(
...
+  list: HTMLGuxListboxElement,
+  location: 'next' | 'previous' | 'first' | 'last'
+): ListboxOptionElement {
+  const availableOptionsList = getListOptions(list).filter(
+    option => !option.disabled && !option.filtered
+  );
+  const activeOption = getActiveOption(list);
+  const activeOptionIndex = availableOptionsList.indexOf(activeOption);
+
+  switch (location.toUpperCase()) {
```
daragh-king-genesys, line 67: Why `toUpperCase`?

packages/genesys-spark-components/src/components/stable/gux-listbox/gux-listbox.service.ts
```diff
@@ -3,11 +3,27 @@ import { ListboxOptionElement } from './options/option-types';
 export function getListOptions(
   list: HTMLGuxListboxElement
 ): ListboxOptionElement[] {
-  return (Array.from(list.children) as ListboxOptionElement[]).filter(item => {
+  const listChildren = Array.from(list.children);
+
+  const allOptions = [];
+  listChildren.map(child => {
+    if (isOptionGroup(child)) return allOptions.push(...child.children);
+    return allOptions.push(child);
```
daragh-king-genesys, line 11: Why is this code different to the multi version?

packages/genesys-spark-components/src/components/stable/gux-listbox/gux-listbox.service.ts
```diff
@@ -38,23 +54,53 @@ function setActiveOption(
...
+    case 'LAST':
+      return availableOptionsList[availableOptionsList.length - 1];
+    case 'PREVIOUS':
+      return availableOptionsList[activeOptionIndex - 1];
+    default:
+      return availableOptionsList[activeOptionIndex + 1];
+  }
+}
+
 function getFirstOption(list: HTMLGuxListboxElement): ListboxOptionElement {
```
daragh-king-genesys, line 79: I think we may be able to simplify these function. I will give you a call on zoom later to discuss.

packages/genesys-spark-components/src/components/stable/gux-listbox/option-group/gux-option-group.tsx
321gillian: The request was that the divider not be shown if the group is the last item in a listbox.

packages/genesys-spark-components/src/components/stable/gux-listbox/option-group/gux-option-group.tsx
321gillian: Updated

packages/genesys-spark-components/src/components/stable/gux-listbox/option-group/gux-option-group.tsx
321gillian: I'll add in the prop for now. Will have to check with design about styling.

PR #272, opened 2024-01-10
feat(table) - Addition of fixed column table example

gavin-everett-genesys:
After investigating the different possible solutions I thinks its best to just have an example on the examples page instead of going down the route of attributes which started to get messy as I was attempting to do it as columns can be different widths etc. 

The example in this PR is just for a basic data table but I can see potentially developers trying to use this with `resizable-columns` which will add a bit more complexity and discussion to this task.

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-table/example.html
+++ b/packages/genesys-spark-components/src/components/stable/gux-table/example.html
@@ -389,55 +389,6 @@
 
 <h2>Data table with sortable columns</h2>
 
-<style
-  onload="(function () {
-    function ascending( a, b ) {
-      if ( a.textContent < b.textContent ){
-        return -1;
-      }
-      if ( a.textContent > b.textContent ){
-        return 1;
-      }
-      return 0;
-    }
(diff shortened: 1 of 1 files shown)


packages/genesys-spark-components/src/components/stable/gux-table/example.html
```diff
@@ -352,16 +352,20 @@ <h2>Data table with resizable columns</h2>
   <table slot="data">
     <thead>
       <tr>
-        <th data-column-name="first-name">First name</th>
-        <th data-column-name="last-name">Last name</th>
+        <th class="gux-sticky-cell" data-column-name="first-name">
```
thomas-c-dillon, line 355: Should these have the style classes as well or where they just for testing and should be removed?

gavin-everett-genesys: Ye, you are correct I have now removed them they were for testing.

thomas-c-dillon (approved):

PR #277, opened 2024-01-15
feat(form-field-dropdown) - remove extra disabled styling

gavin-everett-genesys:
Removing extra disabled styling within `form-field-dropdown` as it is already applied through the `gux-dropdown` component. Also remove the `disabled` states as I do not think they are needed anymore.

https://inindca.atlassian.net/browse/COMUI-1761


daragh-king-genesys (approved):

daragh-king-genesys:
Does this change need to happen in v4 too?

gavin-everett-genesys:
> Does this change need to happen in v4 too?

Yep I will make this change in v4 also

PR #289, opened 2024-01-22
fix(table): amend shadow to false for gux-all-row-select, gux-row-select

gavin-everett-genesys:
A bug was found in the `gux-all-row-select` and `gux-row-select` internal components in new versions of safari 16+. The root cause of this issue was from enabling `shadow` on these components during `v4` developments. I have now set `shadow` to false on these components for the time being.

https://inindca.atlassian.net/browse/COMUI-2582

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-table/gux-all-row-select/gux-all-row-select.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-table/gux-all-row-select/gux-all-row-select.tsx
@@ -16,8 +16,7 @@ import tableResources from '../i18n/en.json';
 
 @Component({
   styleUrl: 'gux-all-row-select.scss',
-  tag: 'gux-all-row-select',
-  shadow: true
+  tag: 'gux-all-row-select'
 })
 export class GuxAllRowSelect {
   private inputElement: HTMLInputElement;
--- a/packages/genesys-spark-components/src/components/stable/gux-table/gux-all-row-select/gux-all-row-select.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-table/gux-all-row-select/gux-all-row-select.tsx
@@ -16,7 +16,8 @@ import tableResources from '../i18n/en.json';
 
 @Component({
   styleUrl: 'gux-all-row-select.scss',
-  tag: 'gux-all-row-select'
+  tag: 'gux-all-row-select',
+  shadow: false
 })
 export class GuxAllRowSelect {
   private inputElement: HTMLInputElement;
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-table/gux-all-row-select/gux-all-row-select.tsx
```diff
@@ -17,7 +17,7 @@ import tableResources from '../i18n/en.json';
 @Component({
   styleUrl: 'gux-all-row-select.scss',
   tag: 'gux-all-row-select',
-  shadow: true
+  shadow: false
```
daragh-king-genesys, line 20: Moving components back into the light dom should be a last resort. Do you know what is actually causing the issue? We should add an explanation here as to why it is done so we don't do move it back into the shadow dom in v5.

gavin-everett-genesys: I have looked at this in depth and I think this is a safari bug. The reasons for this is that it functions correctly on every other browser. If I were to un-nuke the native styles by removing the `appearance:none` set in `gux-form-field-checkbox.scss` the native styles operate correctly which would indicate the events are being sent and received etc. Once appearance has been set to none for some reason if I click the `gux-all-row-select` it doesn't cause a render update to the `gux-row-select` components in the table but if you hover over them it will. Its a very weird bug maybe I could raise it on the safari dev form. Ive checked the `scss` files and nothing looks incorrect to me.

daragh-king-genesys: Have you tried using `forceUpdate` instead of the move to the shadow dom as a workaround? Maybe a Watch on `gux-row-select.selected` that calls `forceUpdate` or add a method to `gux-row-select` that can be called when the table component updates the `gux-row-select` elements.

gavin-everett-genesys: Ye I have tried multiple different solutions but to no avail. I have tried

- Watch decorator within `gux-row-select` that watches `selected` and once changed it would `forceUpdate` the root. Also tried updating the selected state within that or even the checked state of the input element even though they are correct.
- Tried using the Method decorator which I could call the method when the checkboxes are updated within `gux-table` but still no result.

gavin-everett-genesys: Going to re-look at this issue again.

katie-bobbe-genesys:
Where are we currently with this PR? Is it waiting for further review, or should we close this PR for now and revisit it with other future table work?

gavin-everett-genesys:
> Where are we currently with this PR? Is it waiting for further review, or should we close this PR for now and revisit it with other future table work?

Will close this off and add it back to the backlog, hopefully a newer version of safari may fix this on its own.

PR #291, opened 2024-01-22
fix(logical-css): button and button-slot now use logical css properties

daragh-king-genesys:
button and button-slot now use logical css properties to support rtl text direction

COMUI-2592

### DIFF
--- a/packages/genesys-spark-components/.stylelintrc.json
+++ b/packages/genesys-spark-components/.stylelintrc.json
@@ -1,3 +1,15 @@
 {
-  "extends": ["stylelint-config-genesys-spark-components"]
+  "extends": ["stylelint-config-genesys-spark-components"],
+  "overrides": [
+    {
+      "files": [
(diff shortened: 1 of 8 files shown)


packages/genesys-spark-components/.stylelintrc.json
```diff
@@ -1,3 +1,15 @@
 {
-  "extends": ["stylelint-config-genesys-spark-components"]
+  "extends": ["stylelint-config-genesys-spark-components"],
+  "overrides": [
+    {
+      "files": [
+        "src/components/stable/gux-button/**/*.{scss,html}",
+        "src/components/stable/gux-button-slot/**/*.{scss,html}"
```
daragh-king-genesys, line 7: I think it would be safe to make this change in all components at once but to be cautious I will add this change to a few components at a time and make sure there is no unexpected issues.

jordanstith15 (approved):

PR #292, opened 2024-01-23
fix(tab): removed fixed width from gux-tab

gavin-everett-genesys:
removed fixed width from gux-tab

[COMUI-2573](https://inindca.atlassian.net/browse/COMUI-2573)

[COMUI-2573]: https://inindca.atlassian.net/browse/COMUI-2573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-tabs/gux-tab/gux-tab.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-tabs/gux-tab/gux-tab.scss
@@ -69,13 +69,11 @@ gux-tabs[alignment='full-width'] > gux-tab-list {
 
 gux-tab {
   display: flex;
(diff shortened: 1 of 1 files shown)


jordanstith15 (approved):

daragh-king-genesys:
@gavin-everett-genesys can you follow up with UX if that token should be removed from the design tokens file too.

gavin-everett-genesys:
> @gavin-everett-genesys can you follow up with UX if that token should be removed from the design tokens file too.

Confirmed with UX token is not needed, ticket has been created on GDS board to remove from tokens file. 👍

PR #294, opened 2024-01-23
feat(phone-input): phone input restyle for v4

321gillian:


jordanstith15:
I noticed the alignment of the country dropdown button looks a bit weird with the hover/clicked color, mainly because of the lack of space on the right side of the caret. Is that how the design is supposed to be?
Otherwise the PR looks good.

321gillian:
> I noticed the alignment of the country dropdown button looks a bit weird with the hover/clicked color, mainly because of the lack of space on the right side of the caret. Is that how the design is supposed to be? Otherwise the PR looks good.

Hi Jordan, I took this back to design and they updated the spacing around the caret. I've applied that now. Can you review again if you have time?

jordanstith15:
> > I noticed the alignment of the country dropdown button looks a bit weird with the hover/clicked color, mainly because of the lack of space on the right side of the caret. Is that how the design is supposed to be? Otherwise the PR looks good.
> 
> Hi Jordan, I took this back to design and they updated the spacing around the caret. I've applied that now. Can you review again if you have time?

looks good!

jordanstith15 (approved):

PR #311, opened 2024-01-30
fix(time-zone-picker): return closest city to user entry

gavin-everett-genesys:
This will return the closest city based on what the user is searching for. For example if you type `Barcelona` it will return `Madrid` which is the same timezone. Or for example if you type `San Jose` it will return `Los Angeles`.

I think this may be useful to backport also 🤔 

https://inindca.atlassian.net/browse/COMUI-2536

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
+++ b/packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
@@ -76,20 +76,26 @@ export class GuxTimeZonePickerBeta {
   }
 
   private filterTimeZoneList(timeZoneList: GuxTimeZoneOption[]) {
-    const searchString = this.searchString;
+    const searchString = this.searchString.toLowerCase();
     return timeZoneList.filter(tzOption => {
-      return this.getFormattedTimeZoneOption(tzOption)
-        .toLowerCase()
-        .includes(searchString.toLowerCase());
+      return tzOption.mainCities.some(city => {
+        return tzOption.displayTextNameFormatted
+          .concat(tzOption.baseDisplayOffsetText)
+          .concat(city)
+          .toLowerCase()
+          .includes(searchString);
+      });
     });
   }
 
   private getTimeZoneOption(timeZone: GuxTimeZoneListing): GuxTimeZoneOption {
     const localizedGroupName = this.i18n(timeZone.name);
+
     const localizedCountryName = this.i18n(timeZone.countryName);
     if (!localizedGroupName) {
       return;
     }
+
     const formattedOffset = formatOffset(timeZone.currentTimeOffsetInMinutes);
     const localizedUTC = this.i18n('UTC');
     const displayTextName = `${localizedGroupName}`;
@@ -97,6 +103,7 @@ export class GuxTimeZonePickerBeta {
     const baseDisplayOffsetText = `${displayTextOffset}`;
     const displayTextNameFormatted = displayTextName.replace(/_/g, ' ');
     const countryName = `${localizedCountryName}`;
+    const mainCities = timeZone.mainCities;
     const defaultZone = '';
     const priority = displayTextName.startsWith('Etc/GMT') ? 2 : 1;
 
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
```diff
@@ -76,11 +76,15 @@ export class GuxTimeZonePickerBeta {
   }
 
   private filterTimeZoneList(timeZoneList: GuxTimeZoneOption[]) {
-    const searchString = this.searchString;
+    const searchString = this.searchString.toLowerCase();
```
jason-evans-genesys, line 79: Should we do a null check for `this.searchString` like so? `this.searchString?.toLowerCase();` Perhaps if `this.searchString` is ever null maybe we should check that a level higher than this. I see `this.searchString` gets set with this code: `this.searchString = event.detail;` on line 55. Perhaps we should do `this.searchString = event.detail || "";` instead?

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
```diff
@@ -76,11 +76,15 @@ export class GuxTimeZonePickerBeta {
...
     return timeZoneList.filter(tzOption => {
-      return this.getFormattedTimeZoneOption(tzOption)
-        .toLowerCase()
-        .includes(searchString.toLowerCase());
+      return tzOption.mainCities.some(city => {
+        return tzOption.displayTextNameFormatted
```
jason-evans-genesys, line 82: I feel like `displayTextNameFormatted` is a little confusing because it makes me wonder how exactly it's formatted and then I have to check elsewhere in this file to see. Perhaps we just format it when defining`displayTextName` initially and then just use that variable? Thoughts?

gavin-everett-genesys: I agree with this I dont think the names convey pretty well these were what they were called in the initial component implementation but I think they should change for sure 👍

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
gavin-everett-genesys: I dont think a `null` check would be required here as the `this.searchString` is set via the `gux-filter` which would mean a key has been pressed and the detail of that event would be assigned to `this.searchString` ?

jason-evans-genesys:
Should we add a sort on the timezone picker dropdown items? It seems like they're not sorted.

gavin-everett-genesys:
> Should we add a sort on the timezone picker dropdown items before you type in text in the search box? It seems like they're not sorted.

Good idea they are now sorting by `city`

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
```diff
@@ -76,11 +76,14 @@ export class GuxTimeZonePickerBeta {
...
     return timeZoneList.filter(tzOption => {
-      return this.getFormattedTimeZoneOption(tzOption)
-        .toLowerCase()
-        .includes(searchString.toLowerCase());
+      return tzOption.mainCities.some(city => {
+        return tzOption.displayTextName
+          .concat(tzOption.baseDisplayOffsetText)
+          .concat(city)
+          .toLowerCase()
+          .includes(this.searchString.toLowerCase());
```
daragh-king-genesys, line 85: Does this filter only work for english?

gavin-everett-genesys: Ye I tested it through Dutch and French and the filter still works as intended 👍

daragh-king-genesys: As discussed on zoom the city names need to be localized for this to work as expected and the may be a huge localisation effort .

daragh-king-genesys (changes requested):

gavin-everett-genesys:
Localisation of the main cities is currently being reviewed by UX.

PR #312, opened 2024-01-30
feat(icons): icon updates in v3

321gillian:
Icon updates in v3

GDS-2374


gavin-everett-genesys (approved):

packages/genesys-spark-components/src/components/stable/gux-icon/gux-icon.types.ts
```diff
@@ -182,9 +183,9 @@ export type GuxIconIconName =
   | 'fa/align-right-regular'
   | 'fa/arrow-down-left-and-arrow-up-right-to-center-regular'
   | 'fa/arrow-down-wide-short-regular'
-  | 'fa/arrow-left-to-line-solid'
+  | 'fa/arrow-left-to-line-regular'
   | 'fa/arrow-right-to-bracket-regular'
-  | 'fa/arrow-right-to-line-solid'
+  | 'fa/arrow-right-to-line-regular'
```
daragh-king-genesys, line 188: This is technically a breaking change so I don't think we should do it. Can we just let both versions of the icons exist in v3 and see what we can do in v4 to move people away from using the one we don't want them to.
ie. don't move `fa/arrow-left-to-line-solid` and `fa/arrow-right-to-line-solid` to legacy

321gillian: I reverted that last commit. Ok now?

daragh-king-genesys (approved):

PR #333, opened 2024-02-14
chore(gux-month-picker): region aware date formatting

katie-bobbe-genesys:
Date and time formatting should get now supplement region data from the browser.
This behavior is "feature toggled" using a cookie, called spark-enable-regional-dates. To test out the functionality, set document.cookie = 'spark-enable-regional-dates' in your console.

### DIFF
--- a/packages/genesys-spark-components/src/genesys-spark-utils/get-closest-element.ts
+++ b/packages/genesys-spark-components/src/genesys-spark-utils/get-closest-element.ts
@@ -1,0 +1,23 @@
+//https://inindca.atlassian.net/browse/COMUI-2673
+export function closestPassShadow(
+  node: Element | ParentNode | null,
+  selector: string
+): HTMLElement | null {
+  if (!node) {
+    return null;
+  }
+
+  if (node instanceof ShadowRoot) {
+    return closestPassShadow(node.host, selector);
+  }
+
+  if (node instanceof HTMLElement) {
+    if (node.matches(selector)) {
+      return node;
+    } else {
+      return closestPassShadow(node.parentNode, selector);
+    }
+  }
+
+  return closestPassShadow(node.parentNode, selector);
+}
--- a/packages/genesys-spark-components/src/genesys-spark-utils/get-closest-element.ts
+++ b/packages/genesys-spark-components/src/genesys-spark-utils/get-closest-element.ts
@@ -1,5 +1,5 @@
 //https://inindca.atlassian.net/browse/COMUI-2673
-export function closestPassShadow(
+export function getClosestElement(
   node: Element | ParentNode | null,
   selector: string
 ): HTMLElement | null {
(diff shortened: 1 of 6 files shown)


packages/genesys-spark/src/utils/intl.ts
```diff
@@ -53,6 +53,28 @@ export function relativeTimeFormat(
   }
 }
 
+function closestPassShadow(
```
jason-evans-genesys, line 56: Curious if this can be defined in one util instead of 2?

packages/genesys-spark-components/src/genesys-spark-utils/intl.ts
```diff
@@ -53,6 +53,28 @@ export function relativeTimeFormat(
   }
 }
 
+function closestPassShadow(
```
daragh-king-genesys, line 56: Have a look at genesys-spark/packages/genesys-spark-components/src/utils/dom/get-closest-element.ts and see if that can be used centralised etc...

packages/genesys-spark-components/src/genesys-spark-utils/get-closest-element.ts
```diff
@@ -0,0 +1,23 @@
+//https://inindca.atlassian.net/browse/COMUI-2673
```
jason-evans-genesys, line 1: Thoughts on having a more general element helper file? I feel like this current one will only ever have one method for getting the closest element.

packages/genesys-spark/src/utils/get-closest-element.ts
```diff
@@ -0,0 +1,23 @@
+//https://inindca.atlassian.net/browse/COMUI-2673
```
jason-evans-genesys, line 1: Is there a way to not duplicate this helper file in 2 places?

katie-bobbe-genesys: I agree with you, this is why I created the COMUI-2673 ticket to follow up on this issue. Do we this this is a blocker for this PR?

jason-evans-genesys: Ok awesome. I don't think it's a blocker for this PR.

packages/genesys-spark-components/src/genesys-spark-utils/get-closest-element.ts
```diff
@@ -0,0 +1,23 @@
...
+export function closestPassShadow(
```
321gillian, line 2: Could you explain a bit more about the idea behind this function name? I get what it does but I'm just wondering if there's a different way to describe it.

katie-bobbe-genesys: I updated the function name to match the name of the file, `getClosestElement`. I do have a follow up ticket, COMUI-2673 to centralize this function and remove duplication

daragh-king-genesys (approved):

PR #336, opened 2024-02-15
feat(time-zone-picker): allow custom defaults

gavin-everett-genesys:
Open to opinions on this change https://inindca.atlassian.net/browse/COMUI-2381

It was hard to come up with a solution that seemed good. My suggested approach works well but I am unsure. The `work-space-default` and `local-default` properties seem to be use case specific so I did not want to alter them. The only logical solution I could think of was to add a `custom-default` property which the end user would set along with a label.

I know this component is still in beta so maybe we could alter it a bit more.

Once we agree on an approach this feeback will be sent to the UX team to be reflected in the designs.

Any feedback is welcome :)


packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/example.html
```diff
@@ -8,16 +8,22 @@ <h3>Default Zones</h3>
   <gux-time-zone-picker-beta
     value="Etc/GMT+1"
     workspace-default="Etc/GMT"
-    ,
     local-default="America/Detroit"
```
321gillian, line 11: Do we have any validation on these strings to check if they are really timezones?

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
```diff
@@ -20,14 +20,20 @@ export class GuxTimeZonePickerBeta {
   @Element()
   private root: HTMLElement;
 
+  @Prop({ mutable: true })
+  value: string;
+
+  @Prop()
+  workspaceDefault: string;
```
321gillian, line 27: Could you add some annotation to indicate the difference between workspace default and custom default?

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
```diff
@@ -114,13 +120,13 @@ export class GuxTimeZonePickerBeta {
   }
 
   private getFormattedTimeZoneOption(option: GuxTimeZoneOption): string {
+    console.log(option);
```
321gillian, line 123: Remove

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
```diff
@@ -20,14 +20,20 @@ export class GuxTimeZonePickerBeta {
...
-  public value: string;
+  localDefault: string;
 
   @Prop()
-  public workspaceDefault: string;
+  customDefault: string;
 
   @Prop()
-  public localDefault: string;
+  customDefaultLabel: string;
```
jason-evans-genesys, line 36: What's the difference between the `customDefaultLabel` prop and `customDefault`?

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
```diff
@@ -167,18 +170,16 @@ export class GuxTimeZonePickerBeta {
...
 
     defaultZoneOptions.forEach(option => {
-      if (
-        defaultZoneOptions.length === 1 &&
-        this.workspaceDefault === this.localDefault
-      ) {
-        defaultZoneOptions[0].defaultZone = `${this.i18n(
-          'localAndWorkspaceDefault'
-        )}`;
+      if (this.workspaceDefault === this.localDefault) {
```
jason-evans-genesys, line 173: i wonder if we can have the word `zone` in the `workspaceDefault` and `localDefault` props to signify that they are related to zones?

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
gavin-everett-genesys: The main difference between the two is that for example if you were to choose `localDefault` or `workspaceDefault` only one property is needed as the the string is hardcoded to either `Local Default` or `Workspace Default` and you just pass your time zone to the property whereas with the  two custom properties we would need 2 because I do not know what label the consumer will provide so we would need two properties one for the label and one for the timezone which they want to use. It would be nice to use property somehow but I dont think this is possible.

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
gavin-everett-genesys: I think this would be a breaking change if I were to change the property names.

gavin-everett-genesys:
Since this PR has gone stale I want to provide an update on the changes made.

So we currently have `workSpaceDefault` and `localDefault` which are use-case specific. For example if you use the `workSpaceDefault` property you will specify a timezone and this will appear under the defaults section in the dropdown with the text hardcoded to `( Workspace Default ) `.

My change alters the API a small bit but does not cause a breaking change. I have introduced two new properties `customDefault` and `customDefaultLabel`. The `customDefaultLabel` property will allow the consumer to provide a label for example " (BU Timezone) " as outlined in the ticket. And then `customDefault` would be to specify a timezone.

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
```diff
@@ -20,14 +20,20 @@ export class GuxTimeZonePickerBeta {
   @Element()
   private root: HTMLElement;
 
+  @Prop({ mutable: true })
```
321gillian, line 23: How come this has changed to mutable? I don't see any other change to the value here?

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
jason-evans-genesys: Gotcha, thanks!

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
jason-evans-genesys: Ah nevermind

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
```diff
@@ -143,14 +148,11 @@ export class GuxTimeZonePickerBeta {
...
-    }
-    if (this.localDefault) {
-      defaultZones.push(this.localDefault);
-    }
-    return defaultZones;
+    return [
+      this.workspaceDefault,
+      this.localDefault,
+      this.customDefault
+    ].filter(zone => !!zone);
```
jason-evans-genesys, line 155: what does this line do? `].filter(zone => !!zone);`? For my own understanding. Are you trying to filter out empty values from workspaceDefault, localDefault, and customDefault?

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
```diff
@@ -167,18 +169,16 @@ export class GuxTimeZonePickerBeta {
...
-          'localAndWorkspaceDefault'
-        )}`;
+      if (this.workspaceDefault === this.localDefault) {
+        option.defaultZone = `${this.i18n('localAndWorkspaceDefault')}`;
       } else if (option.value === this.workspaceDefault) {
         option.defaultZone = `${this.i18n('workspaceDefault')}`;
       } else if (option.value === this.localDefault) {
         option.defaultZone = `${this.i18n('localDefault')}`;
       }
+      if (option.value === this.customDefault) {
```
jason-evans-genesys, line 179: The zone list is user defined, right? Wondering if the user would ever have an extra space or even a typo in the props: defaultZone and workspaceDefault and then the exact match check here would fail? Maybe that's not our problem and would be on the user?

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
gavin-everett-genesys: Ye so basically if a user has provided a timezone for those properties it will filter those out out say for example you have `gux-timezone-picker-beta` setup like,

```
  <gux-time-zone-picker-beta
    value="Etc/GMT+1"
    workspace-default="Etc/GMT"
    local-default="America/Detroit"
  > 
```

The funtion would then filter out `Etc/GMT` and `America/Detroit` and store those in an array which will be used to display under the default zones list.

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
gavin-everett-genesys: Thats a good question I am not sure as of yet as there was no validation incorporated for the initial properties which were `workspace-default` and `local-default`. I will wait till we are happy with this api change and review this.

321gillian:
Tests failing, might just need to update snapshot

packages/genesys-spark-components/src/components/beta/gux-date-time/tests/__snapshots__/gux-date-time.spec.ts.snap
```diff
@@ -331,7 +331,7 @@ exports[`gux-date-time #render using non regional dates #render different dateti
 exports[`gux-date-time #render using non regional dates #render different datetime formats eg(short,medium,full,long) should render component as expected (2) 16`] = `
 <gux-date-time-beta datetime="2022-07-07T13:35:30.100Z" format="medium">
   <mock:shadow-root>
-    7. jul. 2022, 13:35:30
+    7. juli 2022, 13:35:30
```
gavin-everett-genesys, line 334: I think the failure is due to this not sure why it was picked up but will investigate it.

321gillian (approved):

jason-evans-genesys:
@gavin-everett-genesys I noticed there's no keyboard navigation way to get to the default zone area in the time-zone-picker dropdown. Is this expected?

Screen Shot 2024-06-17 at 10 21 11 AM

katie-bobbe-genesys (changes requested):
As Jason mentioned, the default zones do not seem to be keyboard accessible. We should resolve this issue before merging

gavin-everett-genesys:
> @gavin-everett-genesys I noticed there's no keyboard navigation way to get to the default zone area in the time-zone-picker dropdown. Is this expected?
> 
> Screen Shot 2024-06-17 at 10 21 11 AM

I think this exists in the live version aswell. I will work on a fix in this PR 👍

gavin-everett-genesys:
> As Jason mentioned, the default zones do not seem to be keyboard accessible. We should resolve this issue before merging

Updated to use `gux-option-group-beta` to display `gux-option` elements.

packages/genesys-spark-components/src/components/stable/gux-icon/readme.md
```diff
@@ -10,7 +10,7 @@
...
...| "fa/arrow-right-to-line-regular" \| "fa/arrow-right-to-line-solid" \| "fa/arrow-rotate-left-regular" \| "fa/arrow-turn-down-right-regular" \| "fa/arrow-up-from-line-regular" \| "fa/arrow-up-left-regular" \| "fa/arrow-up-regular" \| "fa/arrow-up-right-and-arrow-down-left-from-center-regular" \| "fa/arrow-up-right-from-square-regular" \| "fa/arrow-up-right-regular" \| "fa/arrows-down-to-people-regular" \| "fa/arrows-maximize-regular" \| "fa/arrows-minimize-regular" \| "fa/arrows-rotate-regular" \| "fa/arrows-rotate-solid" \| "fa/arrows-to-circle-regular" \| "fa/arrows-up-down-regular" \| "fa/asterix-regular" \| "fa/at-regular" \| "fa/award-solid" \| "fa/backward-fast-regular" \| "fa/backward-regular" \| "fa/backward-step-regular" \| "fa/badge-check-solid" \| "fa/ban-outline" \| "fa/ban-regular" \| "fa/bars-regular" \| "fa/bell-on-regular" \| "fa/bell-regular" \| "fa/bell-slash-regular" \| "fa/binoculars-regular" \| "fa/block-brick-regular" \| "fa/bold-regular" \| "fa/bolt-regular" \| "fa/book-regular" \| "fa/bookmark-regular" \| "fa/border-all-regular" \| "fa/box-archive-regular" \| "fa/box-regular" \| "fa/brackets-square-regular" \| "fa/briefcase-regular" \| "fa/browser-regular" \
```
jordanstith15, line 13: With this many icons, maybe we should discuss whether the automatic script for the icon name attribute is worth it. Feels like it just makes the readme less readable

jordanstith15 (approved):

packages/genesys-spark-components/src/components/beta/gux-time-zone-picker/gux-time-zone-picker.tsx
```diff
@@ -202,25 +202,33 @@ export class GuxTimeZonePickerBeta {
     });
   }
 
-  private renderDefaultsList(): JSX.Element {
+  private renderDefaultsList(): JSX.Element | null {
```
jason-evans-genesys, line 205: For some reason having a type of JSX.Element or null feels off to me here. I feel like it should be only one type. Is there a way to render an empty JSX.Element maybe instead of null?

katie-bobbe-genesys (commented):
This looks good to me. Just make sure to resolve conflicts in the icons readme before merging.

katie-bobbe-genesys (approved):

PR #353, opened 2024-03-04
Fix(tooltip): fixed overflow-wrap for long words 

daragh-king-genesys:


packages/genesys-spark-components/src/components/beta/gux-tooltip-base/gux-tooltip-base.scss
```diff
@@ -44,6 +44,12 @@
     color: var(--gse-ui-tooltip-light-iconColor);
   }
 
+  ::slotted(*) {
+    width: 100%;
+    overflow-wrap: break-word;
```
jordanstith15, line 49: Was this tested with asian languages? I've had weird things when trying to to wrap Japanese characters as each character is considered a word.

packages/genesys-spark-components/src/components/stable/gux-tooltip/gux-tooltip.tsx
```diff
@@ -219,7 +219,9 @@ export class GuxTooltip {
           }}
           data-placement={this.placement}
         >
-          <slot />
+          <slot name="content">
```
jordanstith15, line 222: Would changing the slot to have a name be considered a breaking change? will it break apps that currently use it with no slot name?

packages/genesys-spark-components/src/components/beta/gux-tooltip-base/gux-tooltip-base.scss
jordanstith15: I tested it with Japanese. It looks good in the new beta tooltip, but the stable tooltip still overflows the tooltip container.

packages/genesys-spark-components/src/components/stable/gux-tooltip/gux-tooltip.tsx
daragh-king-genesys: Yes but by falling back to a default slot in the new named slot I think it makes it a non breaking change.

packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/gux-tooltip.tsx
```diff
@@ -110,7 +108,11 @@ export class GuxTooltip {
           tooltipId={this.id}
           ref={el => (this.baseTooltip = el)}
         >
-          <slot />
+          <span slot="content">
+            <slot name="content">
+              <slot />
```
daragh-king-genesys, line 113: @thomas-c-dillon not sure if falling back to a default slot is a good idea here or not. Will discuss at standup.

packages/genesys-spark-components/src/components/beta/gux-tooltip-base/gux-tooltip-base.scss
daragh-king-genesys: It works in the stable one too but you can't style text nodes so you have to slot an element. That is why I am moving over to the named slot as that does not allow you to slot a text node.

packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/gux-tooltip.tsx
thomas-c-dillon: This makes sense syntactically. I can see where it is confusing though. Alternatively, as this is the beta version, we could just choose not to support the default slot here but that will require a larger version bump when we replace the stable version.

Alternatively, alternatively, we could make another component specifically for wrapping text node children that applies typography styles including word break (the sibling of gux-truncate).

daragh-king-genesys: > we could just choose not to support the default slot here but that will require a larger version bump when we replace the stable version.

I suggest we leave it like I have it for now and depending on when we migrate the beta version to stable we can leave or change the api. before v5 leave the api in v5 remove the default slot.

jordanstith15 (approved):

PR #354, opened 2024-03-04
chore(icon): add regular and restricted icons

321gillian:
GDS-2424


packages/genesys-spark-components/src/components/stable/gux-icon/example.html
```diff
@@ -5,6 +5,7 @@ <h1>gux-icon</h1>
     <gux-tab tab-id="1-1">Usage</gux-tab>
     <gux-tab tab-id="1-2">Current icons</gux-tab>
     <gux-tab tab-id="1-3">Legacy icons</gux-tab>
+    <gux-tab tab-id="1-4">Restricted icons</gux-tab>
```
321gillian, line 8: I also considered making the 3rd tab 'Legacy & Restricted' as they are both kind of discouraged. Let me know if you'd prefer that

packages/genesys-spark-components/src/components/stable/gux-icon/gux-icon.types.ts
```diff
@@ -592,6 +595,12 @@ export type GuxIconIconName =
   | 'refresh'
   | 'reset'
   | 'resource'
+  | 'restricted/fa/building-user-regular'
+  | 'restricted/fa/calendar-circle-user-regular'
+  | 'restricted/fa/compass-drafting-regular'
+  | 'restricted/fa/grid-2-regular'
+  | 'restricted/fa/headset-regular'
+  | 'restricted/fa/screwdriver-wrench-regular'
```
321gillian, line 603: My thinking on this was that it seems important to flag the restriction in the icon name but also open to other views here.

jason-evans-genesys (approved):

PR #358, opened 2024-03-05
fix(gux-modal): added space around content slot

jason-evans-genesys:
Closes https://inindca.atlassian.net/browse/COMUI-2690

I added 5px of padding to the right and left of the content slot. Doing this allows focusable elements to not be cut off on the left and right sides. I realize we could also ask the client to add spacing on their end instead of doing it inside the gux-modal component. Thoughts on adding spacing like I did and if so what kind of value makes sense? I feel like 5px is arbitrary. Should we use a token?

^^ **Update**: Daragh pushed a commit that makes the above not relevant anymore since the initial padding change I made had changed the design which we do not want.

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-modal/gux-modal.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-modal/gux-modal.scss
@@ -18,7 +18,8 @@ slot[name='start-align-buttons']::slotted(:not(button, gux-button-slot)) {
   dialog {
     box-sizing: border-box;
     justify-content: space-between;
-    border: 1px solid #b4bccb; // TODO: COMUI-2300
+    padding: 0;
+    border: none;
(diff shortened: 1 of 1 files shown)


daragh-king-genesys:
@jason-evans-genesys your change was changing the design (adding extra padding) so I had a quick look at how we could fix the issue without changing the design. It ended up easier to commit what I found than commenting it in the PR.

jason-evans-genesys:
> @jason-evans-genesys your change was changing the design (adding extra padding) so I had a quick look at how we could fix the issue without changing the design. It ended up easier to commit what I found than commenting it in the PR.

Thanks for your help! Your change works great. Tested it and seems to work.

jason-evans-genesys:
@daragh-king-genesys Is this PR good to be approved now?

daragh-king-genesys (approved):

PR #359, opened 2024-03-06
feat(avatar): add focus styled examples

321gillian:
COMUI-2665


packages/genesys-spark-components/src/components/beta/gux-avatar/example.html
```diff
@@ -470,121 +470,149 @@ <h2>Notifications</h2>
...
-  <button class="gux-avatar-button gux-focus-ring-small-focused-visible">
-    <gux-avatar-beta
-      size="xsmall"
-      name="Conor Darcy"
-      presence="available"
-      badge
-      ring
-    ></gux-avatar-beta>
-  </button>
+  <gux-avatar-focusable-beta>
```
321gillian, line 480: I don't know if it's better to have this component within avatar folder and only have this example or to have a separate set of examples etc for the focusable avatar.

packages/genesys-spark-components/src/components/beta/gux-avatar-focusable/gux-avatar-focusable.tsx
```diff
@@ -0,0 +1,46 @@
...
+        'An anchor tag or button tag must be slotted into gux-avatar-focusable'
+      );
+    }
+
+    if (!avatarElement) {
+      logWarn(this.root, 'Slotted element must contain a gux-avatar');
+    }
+
+    if (['small', 'xsmall'].includes(avatarElement.size)) {
+      slottedElement.classList.add('gux-focus-ring-small-focused-visible');
```
daragh-king-genesys, line 32: I think this could be done in a simpler way.

```
  render(): JSX.Element {
    return (
      <div class={`gux-size-${this.avatarElement.size}`}>
        <slot></slot>
      </div>
    ) as JSX.Element;
  }
```
```
.gux-size-small,
.gux-size-xsmall {
  &:focus-visible {
    @include focus.gux-focus-ring-small;
  }
}

....
```

packages/genesys-spark-components/src/components/beta/gux-avatar-focusable/gux-avatar-focusable.tsx
```diff
@@ -0,0 +1,41 @@
...
+      logWarn(
+        this.root,
+        'An anchor tag or button tag must be slotted into gux-avatar-focusable'
+      );
+    }
+  }
+
+  render(): JSX.Element {
+    return (
+      <Host class={`gux-size-${this.avatarElement?.size ?? 'large'}`}>
```
daragh-king-genesys, line 36: Using `gux-table` as a reference I think this should be
```
<Host gs-avatar-size={this.avatarElement?.size ?? 'large'}>
```

packages/genesys-spark-components/src/components/beta/gux-avatar-focusable/gux-avatar-focusable.light.scss
```diff
@@ -0,0 +1,15 @@
...
+  }
+
+  &.gux-size-small,
+  &.gux-size-xsmall {
+    a:focus-visible,
+    button:focus-visible {
+      @include focus.gux-focus-ring-small;
+    }
+  }
+}
```
daragh-king-genesys, line 15: I think it would be safer to do all the styling in the light dom. Is the a reason we shouldn't do that?

321gillian: I was wondering the same thing when I was updating it. I wasn't sure if I should only have the styles that need to be in the light dom in the light dom file. Updated now

packages/genesys-spark-components/src/components/beta/gux-avatar-focusable/gux-avatar-focusable.tsx
321gillian: Matt explained this to me so no need to respond.

321gillian:
Hi @daragh-king-genesys,

Is this ok to go ahead? It will likely change again with the photo elements but for now? 

Thanks,
Gillian

daragh-king-genesys (approved):

PR #360, opened 2024-03-07
fix(datepicker): first keypress after focus should be treated as new input

gavin-everett-genesys:
first keypress after focus should be treated as new input

To be **backported**

[COMUI-2616](https://inindca.atlassian.net/browse/COMUI-2616)

[COMUI-2616]: https://inindca.atlassian.net/browse/COMUI-2616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-datepicker/gux-datepicker.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-datepicker/gux-datepicker.tsx
@@ -162,7 +162,7 @@ export class GuxDatepicker {
    * Tracks the amount of key presses when focusing on the input.
    */
   @State()
-  pressedKeys: Set<string> = new Set();
+  pressedKeys: Array<string> = [];
 
   @Watch('value')
   watchValue() {
@@ -246,6 +246,8 @@ export class GuxDatepicker {
           );
           this.setIntervalRange(previousIntervalRange);
           this.setCursorRange();
+          this.pressedKeys.length = 0;
+
           break;
         }
         case 'ArrowRight': {
@@ -257,6 +259,8 @@ export class GuxDatepicker {
           );
           this.setIntervalRange(nextIntervalRange);
           this.setCursorRange();
+          this.pressedKeys.length = 0;
+
           break;
         }
         default:
@@ -318,6 +322,7 @@ export class GuxDatepicker {
   @OnClickOutside({ triggerEvents: 'mousedown' })
   onClickOutside() {
     this.active = false;
+    this.pressedKeys.length = 0;
   }
 
   /*********  Event Handlers  **********/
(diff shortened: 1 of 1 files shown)


packages/genesys-spark-components/src/components/stable/gux-datepicker/gux-datepicker.tsx
```diff
@@ -325,6 +331,7 @@ export class GuxDatepicker {
   }
 
   onInputFocusOut() {
+    this.pressedKeys.clear();
```
daragh-king-genesys, line 334: I think you may want to do this in `setIntervalRange` not here

packages/genesys-spark-components/src/components/stable/gux-datepicker/gux-datepicker.tsx
```diff
@@ -377,7 +385,7 @@ export class GuxDatepicker {
       ) {
         this.typeYearValue(currentSectionValue, event.key);
       } else {
-        if (this.canSetDate(inputNumber)) {
+        if (this.canSetDate(inputNumber) && this.pressedKeys.size == 2) {
```
jason-evans-genesys, line 388: This might not be necessary but what if we did `this.pressedKeys.size >= 2` as a fail safe just in case something happened and somehow pressedKeys was more than 2 at this point in the code during execution for whatever reason - it'd prob never happen or is very unlikely, but maybe something could get out of whack during program execution.

gavin-everett-genesys: Yeah I think this is a good idea and fixes and issue we were looking at where we were getting stuck in a broken state when trying to type `11`

gavin-everett-genesys:
I think this operates well now. There is another issue which is currently in the live docs aswell which I will raise a ticket for for example if you go to the `year` field and type a year less than 4 character for example `12/01/932` and then start typing in the month or day field it will cause a broken state.

Ticket: https://inindca.atlassian.net/browse/COMUI-2715

packages/genesys-spark-components/src/components/stable/gux-datepicker/gux-datepicker.tsx
```diff
@@ -312,6 +322,7 @@ export class GuxDatepicker {
   @OnClickOutside({ triggerEvents: 'mousedown' })
   onClickOutside() {
     this.active = false;
+    this.pressedKeys.length = 0;
```
jason-evans-genesys, line 325: Should we instead do `this.pressedKeys = []`?

jason-evans-genesys (approved):

PR #365, opened 2024-03-08
feat(content-search): Fix truncation issues for Safari and Firefox

321gillian:


katie-bobbe-genesys:
I would recommend checking the truncation behavior in Firefox and Safari as well as Chrome. I am no longer seeing the reported bug in Chrome, but I see it in Firefox and Safari.

321gillian:
Updated again to fix issue in Safari and Firefox

packages/genesys-spark-components/src/components/stable/gux-content-search/gux-content-search.tsx
```diff
@@ -110,9 +110,7 @@ export class GuxContentSearch {
           <gux-icon decorative icon-name="search"></gux-icon>
         </div>
 
-        <gux-truncate max-lines="1">
-          <slot />
-        </gux-truncate>
+        <slot />
```
daragh-king-genesys, line 113: This is a big change. Can you explain why it is needed. Why is it wrong?

321gillian: gux-truncate as providing a tooltip which based on my testing on Chrome, Firefox and Safari wasn't having any positive impact on the component except in Safari where the tooltip would show on hover and be empty. 
Screenshot 2024-03-11 at 16 16 32

The styling on the slotted input already has truncation within it which from what I could see works better than the addition of gux-truncate here. I think by nesting the input within a gux-truncate the styling on the input was no longer having effect. Also, gux-truncate wasn't really working in terms of truncating the text. 

I couldn't really see what positives the presence of truncate was providing to the component vs not having it.

katie-bobbe-genesys: It does look like the `gux-truncate` isn't working as expected in the current version of the spark components and is not showing a tooltip. The update made in this PR seems to be an improvement on the truncation behavior.

However, I think to fully address this issue, we would want to take a closer look at `gux-truncate` and see why it is not providing a tooltip and truncation as expected here. I think we would want to have `gux-truncate` handle the truncation here and show a tooltip on hover/on focus. If there is too much complexity to this issue and we don't want to look into this in this PR please create a separate ticket to capture this issue.

321gillian: I've replaced gux-truncate. The issue seems to relate to the use of this class which is applied when the max-lines prop is present on gux-truncate. Since the input is a one-line set up, I think we can leave it out. 

```
.gux-truncate-multi-line {
  white-space: normal;

  .gux-truncate-slot-container {
    display: -webkit-box;
    -webkit-box-orient: vertical;
    -webkit-line-clamp: 1;
  }
}
```

321gillian: Still need to do some checking around tooltips here, will revert when have more information

321gillian: After a bunch of further investigation here, I have some more info but the same conclusion. Details below and in the ticket. @daragh-king-genesys @katie-bobbe-genesys  can you review and let me know if this is ok? 

### Summary

Gux-truncate wraps the input in gux-content search. Gux-truncate is not configured to work with a slotted input at this time. Removing the gux-truncate wrapper on the input will fix this issue.  

### Why doesn’t gux-truncate support a slotted input?

 Gux-truncate relies on a slot container’s scrollwidth > its offsetWidth. Since the input is within a flex layout here, and has a width of 100%, its scrollWidth and offsetWidth are equal. 

Typically, an element which is slotted within a gux-truncate would truncate when the text spills outside of the container width. In the case of an input component such as this search field, the field does not expand with its contents. This means that the scrollWidth and offsetWidth remain equal even when the input value has is overflowing. 

As an aside, the function to get the content for the tooltip doesn’t check for input value so if there was a change in the scrollWidth, the tooltip still wouldn’t work. 

### What would it take to get gux-truncate to support a slotted input? 

In order to get gux-content-search to work with gux-truncate in its current form, it would be necessary to do some calculation on the expected width of the content and then try to predict when the tooltip should show or not. This seems like it would be buggy and excessive for the component as it is. 

### How does this compare to native input type=search? 

I compared this aspect of the gux-content-search to [basic HTML input type='search' ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search)and there doesn’t seem to be a precedence there for truncation and tooltip. I also looked as material design and Adobe Spectrum for design system comparisons. Similarly, a tooltip was not present. 

### Conclusion

I think the best way to proceed here is to remove gux-truncate from the component and keep using the slotted(input) styles to apply an ellipsis to the overflow input content. The ellipsis is still a nice feature to have in there.

packages/genesys-spark-components/src/components/stable/gux-content-search/gux-content-search.scss
```diff
@@ -16,6 +16,7 @@
   flex-basis: 100%;
   flex-shrink: 1;
   width: 100%;
+  min-width: 20px;
```
katie-bobbe-genesys, line 19: This should have a token, maybe add to GDS-2451

packages/genesys-spark-components/src/components/stable/gux-content-search/gux-content-search.tsx
katie-bobbe-genesys: This seems reasonable to me. From what I see this should not be a breaking change.

packages/genesys-spark-components/src/components/stable/gux-content-search/gux-content-search.scss
321gillian: Updated that ticket to include this ask.

packages/genesys-spark-components/src/components/stable/gux-content-search/gux-content-search.tsx
gavin-everett-genesys: I think this change makes sense to me and wont be breaking as the truncation does not work anyways on any browser so `text-overflow:ellipsis` should suffice for now. I think we should create a ticket for to investigate how to alter `gux-truncate` to support this

gavin-everett-genesys (approved):

packages/genesys-spark-components/src/components/stable/gux-content-search/gux-content-search.tsx
321gillian: https://inindca.atlassian.net/browse/COMUI-2782 Ticket created to capture this

PR #366, opened 2024-03-11
fix(tokens): remove fixedWidth for tabs token

gavin-everett-genesys:
remove fixedWidth for tabs token

[COMUI-2714](https://inindca.atlassian.net/browse/COMUI-2714)

### DIFF
--- a/packages/genesys-spark-tokens/data/tokens.json
+++ b/packages/genesys-spark-tokens/data/tokens.json
@@ -8962,10 +8962,6 @@
           "value": "{size.xl}",
           "type": "sizing"
         },
-        "fixed_width": {
-          "value": "{size.3xl}",
-          "type": "sizing"
-        },
         "horizontal": {
           "fixed_height": {
             "value": "{size.xl}",


packages/genesys-spark-tokens/data/tokens.json
```diff
@@ -8962,10 +8962,6 @@
           "value": "{size.xl}",
           "type": "sizing"
         },
-        "fixed_width": {
-          "value": "{size.3xl}",
-          "type": "sizing"
-        },
```
daragh-king-genesys, line 8968: Bit confused here. I would have assumed there would have to be corresponding snapshot changes to go with this change. However the tests passed so I am suspicious that there is something wrong with our tests on github.

gavin-everett-genesys: I forgot to run the `new-tokens-data` command which is why the snapshots never updated but I have now. But the tests should probably fail I would assume 🤔

daragh-king-genesys: Had a quick look. The github action is calling `test.ci` that calls `npm run test.ci --workspaces --if-present`.
Not every package has a `test.ci` command, the tokens package being one of them. We should add a `test.ci` script to any package that does not have it. My guess is something like `"test.ci": "npm run test",` in any package that has a `test` npm script but no `test.ci` script.

daragh-king-genesys: Let me know if you want to look into the test issue in this PR or create a seperate ticket for it. If you are doing it in a new ticket I will +1 this PR.

gavin-everett-genesys: Will take a look at it in this PR 👍

gavin-everett-genesys: Actually I will do it in a separate PR  so if it occurs again in the future.

321gillian (approved):

PR #374, opened 2024-03-12
feat(avatar): photo change overlay

321gillian:
Need to update with tokens later today but the rest is ready to review.


packages/genesys-spark-components/src/components/beta/gux-avatar/gux-avatar.tsx
```diff
@@ -134,6 +154,19 @@ export class GuxAvatar {
     ) as JSX.Element;
   }
 
+  private renderCameraIcon(): JSX.Element {
+    if (this.size === 'large' && this.changePhoto) {
```
gavin-everett-genesys, line 158: I am wondering does the size equal large check matter ? I know monse said it was designed specifically for large and the reasoning being is that is only possible to change your profile photo in the profile area of Genesys cloud where the image is large but I dont see why we should limit it to large.

packages/genesys-spark-components/src/components/beta/gux-avatar/gux-avatar.tsx
```diff
@@ -165,6 +198,11 @@ export class GuxAvatar {
     return `${this.name}`;
   }
 
+  private onAvatarClick(): void {
+    if (!this.changePhoto) return;
+    this.edit.emit();
+  }
```
gavin-everett-genesys, line 204: I think this could be simplified to 

`if(this.changePhoto) {
this.edit.emit();
}
`

to avoid the negation.
`

packages/genesys-spark-components/src/components/beta/gux-avatar/gux-avatar.tsx
321gillian: Hi Gavin, I've spoken to UX about the large avatar change photo thing again based on your comment. Their point of view is that the change photo icon on a smaller avatar represents bad UX so they'd rather not have it there.

packages/genesys-spark-components/src/components/beta/gux-avatar/gux-avatar.tsx
```diff
@@ -71,6 +79,18 @@ export class GuxAvatar {
...
+   * Show photo change overlay on hover, only available for large avatars
+   */
+  @Prop()
+  changePhoto: boolean = false;
+
+  /**
+   * Occurs when the item has been clicked when changePhoto is true
+   */
+  @Event()
+  edit: EventEmitter<string>;
```
321gillian, line 92: Not sure if there's a better name for this. Suggestions welcome

conor-darcy: editPhoto?

packages/genesys-spark-components/src/components/beta/gux-avatar/gux-avatar.tsx
```diff
@@ -182,9 +219,12 @@ export class GuxAvatar {
           [`gux-${this.presence}`]: this.ring || this.badge,
           [`gux-${this.size}`]: true,
           'gux-ring': this.ring,
-          [`gux-accent-${this.getAccent()}`]: true
+          [`gux-accent-${this.getAccent()}`]: true,
+          'gux-change-photo': this.changePhoto && this.size === 'large'
         }}
+        onClick={() => this.onAvatarClick()}
       >
+        {this.renderCameraIcon()}
```
conor-darcy, line 227: Should there be a button element somewhere if its clickable?

packages/genesys-spark-tokens/snapshot/json/gse-ui-dataDisplay.json
```diff
@@ -386,7 +390,15 @@
...
       }
     },
-    "focus": "#7B88F7"
+    "focus": "#7B88F7",
+    "add-change-image": {
+      "hover-modal": {
+        "shroud-size": "84px"
+      },
+      "icon": {
+        "size": "16px"
```
conor-darcy, line 399: There is an icon.size.sm or something like that.

conor-darcy:
Have a think about semantics & accessibility. Screenreader users need to have some info announced like "click to change photo" & keyboard users should see the icon on focus.

packages/genesys-spark-components/src/components/beta/gux-avatar/example.html
```diff
@@ -149,13 +157,15 @@ <h2>Dynamic Example</h2>
         const updateBadge = document.getElementsByName('update-badge')[0];
         const updateRing = document.getElementsByName('update-ring')[0];
         const updateNotifications = document.getElementsByName('update-notifications')[0];
+        const updateChangePhoto = document.getElementsByName('update-photo')[0];
```
daragh-king-genesys, line 160: `const updateChangePhoto = document.querySelector('[name=update-photo]');`

packages/genesys-spark-components/src/components/beta/gux-avatar/gux-avatar.scss
```diff
@@ -42,6 +46,36 @@
...
+      z-index: var(--gse-semantic-zIndex-showFocus);
+      display: inline-block;
+      color: var(--gse-ui-avatar-hoverModal-foregroundColor);
+      transform: translate(-50%, -50%);
+    }
+
+    .gux-content::after {
+      position: absolute;
+      top: 4px;
+      left: 4px;
```
daragh-king-genesys, line 65: What are these `4px` for?

packages/genesys-spark-components/src/components/beta/gux-avatar/gux-avatar.tsx
daragh-king-genesys, line 92: `guxchangephoto: EventEmitter<void>;`

packages/genesys-spark-components/src/components/beta/gux-avatar/gux-avatar.tsx
```diff
@@ -134,6 +154,19 @@ export class GuxAvatar {
...
   }
 
+  private renderCameraIcon(): JSX.Element {
+    if (this.changePhoto && this.size === 'large') {
+      return (
+        <gux-icon
+          class="gux-change-photo-icon"
+          icon-name="fa/camera-solid"
+          size="small"
+          decorative
```
daragh-king-genesys, line 164: something off here. this is `decorative` but I don't see how a screen reader user will be informed about the option to change photo. There is no button etc... that will be read to the user.

packages/genesys-spark-tokens/snapshot/json/gse-ui-dataDisplay.json
321gillian: I think it's simpler to remove these sizes and use the icon prop for size. I'll check with Monse about updating this

321gillian:
@daragh-king-genesys @conor-darcy @gavin-everett-genesys 

At this point, the component will have both an alt tag on the image and when the change photo is shown the screenreader icon text. The screenreader will read out something like 'Change profile photo John Doe (available)' when combined with an image presence and name. 

That seems a bit dodgy. 

Is it worth considering making src a prop on the component so the alt tag can be managed? 
In that case, we'd need to figure out whether to call the alt/label attribute alt or label. 
The accessibility text could then be displayed a bit more sensibly for this version of the avatar.

packages/genesys-spark-components/src/components/stable/gux-icon/icons/fa/camera-solid.svg
katie-bobbe-genesys, line 1: I know this is a design decision, but when I see a camera icon I assume that the control will activate a camera. Would a "gallery" like image make more sense here like 
image

katie-bobbe-genesys:
The main issue I see here is that this component is not accessible to a keyboard user

packages/genesys-spark-components/src/components/stable/gux-icon/icons/fa/camera-solid.svg
321gillian: I spoke with Monse about this. She said they just went with a camera for now because that's what's currently there but would look at creating a UX ticket to review the icon later if that's ok?

321gillian:
> @daragh-king-genesys @conor-darcy @gavin-everett-genesys
> 
> At this point, the component will have both an alt tag on the image and when the change photo is shown the screenreader icon text. The screenreader will read out something like 'Change profile photo John Doe (available)' when combined with an image presence and name.
> 
> That seems a bit dodgy.
> 
> Is it worth considering making src a prop on the component so the alt tag can be managed? In that case, we'd need to figure out whether to call the alt/label attribute alt or label. The accessibility text could then be displayed a bit more sensibly for this version of the avatar.

Updating this with a reference to current accessibility info
Screenshot 2024-04-10 at 12 22 48

321gillian:
> The main issue I see here is that this component is not accessible to a keyboard user

I've added some elements to it to make it keyboard navigable. Can you review it again and suggest further improvements if these aren't sufficient?

321gillian:
Updated to describe the avatar with aria-label targeted to change photo, instead of the existing alt tag.

packages/genesys-spark-components/src/components/beta/gux-avatar/example.html
```diff
@@ -614,6 +624,46 @@ <h3>Anchor Tag</h3>
     </a>
   </gux-avatar-focusable-beta>
 </div>
+<h2>Change Photo (large only)</h2>
+<gux-avatar-beta name="Simon Whelehan" change-photo>
+  <img slot="image" src="https://i.pravatar.cc/300" alt="Simon Whelehan" />
+</gux-avatar-beta>
+<gux-avatar-beta
```
katie-bobbe-genesys, line 631: The other instances of `gux-avatar` follow this API:
```
 <gux-avatar-focusable-beta>
    <a href="#"> OR <button>
      <gux-avatar-beta
        size="xsmall"
        name="Conor Darcy"
        presence="available"
        badge
        ring
      ></gux-avatar-beta>
    </a> OR </button>
  </gux-avatar-focusable-beta>
  ```
The API for the `gux-avatar-beta` component with the property `change-photo` should not be significantly different, in my opinion. Also, we need to prevent `gux-avatar-beta` with `change-photo` attribute being slotted into `gux-avatar-focusable-beta.

I would propose that the API for the `gux-avatar` photo change overlay would look something like this:

```
 <gux-avatar-focusable-beta>
    <gux-avatar-change-photo-beta>
      <gux-avatar-beta
        size="large"
        name="Conor Darcy"
        presence="available"
        badge
        ring
      ></gux-avatar-beta>
    </gux-avatar-change-photo-beta>
  </gux-avatar-focusable-beta>
  ```
Where `gux-avatar-change-photo-beta` is a button component that has the change photo event, etc. Let me know if this makes sense or if you have other opinions. I'm not as familiar with the component as you are so I might be missing something.

321gillian: I think this is really good suggestion. I've redone the PR to apply it this way.

packages/genesys-spark-components/src/components/beta/gux-avatar-change-photo/gux-avatar-change-photo.tsx
```diff
@@ -0,0 +1,73 @@
...
+})
+export class GuxAvatarChangePhoto {
+  private i18n: GetI18nValue;
+
+  @Element() root: HTMLElement;
+
+  @Event()
+  guxchangephoto: EventEmitter<void>;
+
+  @Listen('keydown')
```
katie-bobbe-genesys, line 33: Since the onclick handler is being used on a `button` the keyboard events are already handled by the native button element. You can remove this `onKeydown` code block

packages/genesys-spark-components/src/components/beta/gux-avatar/example.html
```diff
@@ -1,6 +1,5 @@
 <h1>gux-avatar-beta</h1>
-
-<h2>Image</h2>
+<!-- <h2>Image</h2>
```
katie-bobbe-genesys, line 2: Should this get uncommented or removed?

katie-bobbe-genesys (approved):
This looks good to me, just either remove or uncomment out the code in the example.html file

packages/genesys-spark-components/src/components/beta/gux-avatar/example.html
321gillian: done

package-lock.json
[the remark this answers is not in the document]
321gillian: oh gross, that should've have happened

PR #383, opened 2024-03-14
fix(date-picker): first keypress after focus should be treated as new input

gavin-everett-genesys:
first keypress after focus should be treated as new input

[COMUI-2616](https://inindca.atlassian.net/browse/COMUI-2616)

[COMUI-2616]: https://inindca.atlassian.net/browse/COMUI-2616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-datepicker/gux-datepicker.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-datepicker/gux-datepicker.tsx
@@ -158,6 +158,12 @@ export class GuxDatepicker {
   @State()
   active: boolean = false;
 
+  /**
+   * Tracks the amount of key presses when focusing on the input.
+   */
+  @State()
+  pressedKeys: Array<string> = [];
+
   @Watch('value')
   watchValue() {
     this.updateDate();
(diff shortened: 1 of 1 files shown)


packages/genesys-spark-components/src/components/stable/gux-datepicker/gux-datepicker.tsx
```diff
@@ -158,6 +158,12 @@ export class GuxDatepicker {
   @State()
   active: boolean = false;
 
+  /**
+   * Tracks the amount of key presses when focusing on the input.
+   */
+  @State()
+  pressedKeys: Array<string> = [];
```
321gillian, line 165: Is there a reason to track the event keys in this array rather than just increment a number?

gavin-everett-genesys: So basically the `pressedKeys` array is being used to track the amount of keypresses on the input. For example if you focus on the `day` field of the input and press `1` you expect `01` then focus out of the input and focus on it again and you press `1` its expected to be `01` again.

Another scenario if you go into the day input and press `1` you expect `01` and then press `1` again expect `11`.

Basically in the V3 version of the datepicker this is not the case so we are trying to get it as close to the native date picker as possible.

If you have any other questions let me know.

321gillian (approved):

PR #388, opened 2024-03-19
chore(utils): remove duplicate genesys-spark utils

katie-bobbe-genesys:


packages/genesys-spark-components/genesys-spark/src/utils/get-closest-element.js
katie-bobbe-genesys, line 1: These js files are being created when I run `npm run build`, is this expected behavior? I removed the script that copies the genesys-spark-utils into `genesys-spark-components` so is this just the default behavior of rollup?

daragh-king-genesys: I had a look at this and think I have an answer. Will discuss at/after standup.

katie-bobbe-genesys:
Will reopen a new PR for the proof of concept for using a utils package

PR #389, opened 2024-03-21
feat(button-slot): allow the slotting of links in gux-button-slot

321gillian:
COMUI-2746

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-button-slot/example.html
+++ b/packages/genesys-spark-components/src/components/stable/gux-button-slot/example.html
@@ -246,6 +246,41 @@
   </div>
 </div>
 
+<h2>Slotted Link</h2>
+<div class="container">
+  <gux-button-slot>
+    <a href="#">Default</a>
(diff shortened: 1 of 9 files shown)


katie-bobbe-genesys:
Semantically, a button is different from a link. I'm almost thinking it might make more sense to create a new component, `gux-anchor-button` that just slots an anchor and provides the style of the button. This way so that consumers are not using anchors and buttons interchangeably. What do you think?

thomas-c-dillon:
> Semantically, a button is different from a link. I'm almost thinking it might make more sense to create a new component, `gux-anchor-button` that just slots an anchor and provides the style of the button. This way so that consumers are not using anchors and buttons interchangeably. What do you think?

I agree with this if UX wants this change to go forward. I still caution against any links that look like buttons at all.

321gillian:
I've started a conversation with the UX team about how to approach these. Will revert when I have more.

321gillian:
Closing this as it needs further consideration before going anywhere else

PR #390, opened 2024-03-21
feat(date-time): use current time in date-time example

321gillian:
GDS-2446

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-date-time/example.html
+++ b/packages/genesys-spark-components/src/components/beta/gux-date-time/example.html
@@ -5,33 +5,40 @@
 <h3>Default</h3>
 <gux-date-time-beta></gux-date-time-beta>
 
-<h3>Short</h3>
-<gux-date-time-beta
-  datetime="2022-01-01T14:48:00.000Z"
-  format="short"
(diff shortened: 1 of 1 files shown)


packages/genesys-spark-components/src/components/beta/gux-date-time/example.html
```diff
@@ -5,33 +5,40 @@ <h2>Date and Time</h2>
...
+      '</div>'
+    ].join(''));
+    timeContainer.appendChild(html);
+  });
+})()"
+>
+  #time-container h3 {
+    text-transform: capitalize;
+  }
+</style>
```
jordanstith15, line 44: is there a way to do this without making the whole component js generated? Like just doing the append to the components themselves?
Doing it this way feels like it takes away from the point of the docs site: seeing how the component is used in html in the code preview (and secondarily being able to compare it to the actual DOM). It makes it harder for consumers to understand how the component works.

I also am not quite understanding where the timestamp variable is used. It doesn't look like it's added to anything

jordanstith15 (changes requested):

jordanstith15 (approved):

PR #392, opened 2024-03-22
chore(label): draft API for feedback

321gillian:
[Figma link](https://www.figma.com/file/JKbHmcf4nUF6C7Pj8M6MpY/Spark%3A-Flare-4.0---Core?type=design&node-id=19542%3A1538&mode=dev)

COMUI-2739

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-label/example.html
+++ b/packages/genesys-spark-components/src/components/beta/gux-label/example.html
@@ -1,0 +1,27 @@
+<!--
+    GUX-LABEL
+    Props:
+    fieldStatus - indicates if field is required, optional or neither
+    actionIcon - displays clickable icon, limited to question mark or info icon
+
+    Behaviour:
+    actionIcon click emits guxactioniconclick
+
+    Styling:
+    Text is not truncated, can be single or multiple lines.
+
+    Attributes:
+    for - standard label attribute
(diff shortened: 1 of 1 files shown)


packages/genesys-spark-components/src/components/beta/gux-label/example.html
```diff
@@ -0,0 +1,27 @@
...
+    Global attributes
+
+    Warnings:
+    Placing inputs inside labels gives an implicit for to the label. In this case, a warning may be logged if the element not considered appropriate within a label. This could be for anchors and buttons etc.
+
+-->
+
+<gux-form-field-text-like>
+  <input slot="input" type="text" name="t-1" value="Sample text" />
+  <gux-label slot="label" field-status="“required”" action-icon="“info”"
```
katie-bobbe-genesys, line 24: Do we need a new component for this? We already provide accessibility and field status (required) to labels that are slotted into our form field components. We can add other needed functionality such as action-icon to the form field components if needed. At first glance this doesn't really seem to be adding any extra utility and is adding complexity. Could you provide some more context as to why we need this new `gux-label` component?

321gillian: Apologies, I tried to find the existing component but totally missed it. I'll work with that one

PR #397, opened 2024-03-27
feat(modal): allow modal to be reopened

conor-darcy:
On the docs site, the modal can't be reopened once closed as closing doesn't change the this.open prop.

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-modal/gux-modal.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-modal/gux-modal.tsx
@@ -84,7 +84,7 @@ export class GuxModal {
     switch (event.key) {
       case 'Escape':
         event.preventDefault();
-        this.dialogElement.close();
+        this.open = false;
         return;
     }
   }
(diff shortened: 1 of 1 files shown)


katie-bobbe-genesys:
I took a look and the change looks fine to me, because `dialogElement.close()` is being called when the `open` state is changed in the `syncOpenState()`. You can mark this PR as ready for review so default reviewers can also take a look.

packages/genesys-spark-components/src/components/stable/gux-modal/gux-modal.tsx
```diff
@@ -84,7 +84,7 @@ export class GuxModal {
     switch (event.key) {
       case 'Escape':
         event.preventDefault();
-        this.dialogElement.close();
+        this.open = false;
```
gavin-everett-genesys, line 87: This is a nitpick but could you just call `this.onDismissHandler()` here instead as it does the same thing ?

gavin-everett-genesys (approved):

PR #398, opened 2024-03-28
fix(form-field-number): add default transparent border to arrow butto…

thomas-c-dillon:
…ns to reserve space for hover

COMUI-2767


321gillian (approved):
Can confirm

PR #405, opened 2024-04-02
fix(gux-accordion): allowing user to highlight text

jason-evans-genesys:
Closes https://inindca.atlassian.net/browse/COMUI-2763

The user was not able to highlight/select text in the accordion content and header, which is now fixed.

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-accordion/gux-accordion-section/gux-accordion-section.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-accordion/gux-accordion-section/gux-accordion-section.scss
@@ -1,7 +1,4 @@
 :host {
-  -webkit-user-select: none;
-  user-select: none;
-
   &:first-child {
     .gux-header {
       border-top: var(--gse-ui-accordion-wrapper-dividerBorder-width)


packages/genesys-spark-components/src/components/stable/gux-accordion/gux-accordion-section/gux-accordion-section.scss
```diff
@@ -1,7 +1,4 @@
 :host {
-  -webkit-user-select: none;
-  user-select: none;
```
thomas-c-dillon, line 3: I am cautious because this looks very deliberate (though probably only intended to affect the header part). I would confirm with UX if this should be removed or if we should add a style specifically for the content..

jason-evans-genesys: @thomas-c-dillon Here's an old ticket from 2 years ago I tracked down that mentions disabling highlighting user text selection on interactive elements. There was then a PR made for this ticket that added these 2 lines specifically: https://inindca.atlassian.net/browse/COMUI-1036 - I'll reach out to design and see how they'd like to move forward with this.

jason-evans-genesys: @thomas-c-dillon I spoke to UX and they QA'ed this and said it works as intended so I think we're good on this code change I made if you are.

thomas-c-dillon (approved):

PR #409, opened 2024-04-08
feat(selector-card): Individual simple radio selector card

321gillian:
COMUI-2667

This PR just covers the initial setup of files and simple radio button instance. Does not include grouping cards or other variants. 
Screenshot 2024-04-08 at 17 40 36
Screenshot 2024-04-08 at 17 40 42

[Figma Link](https://www.figma.com/file/JKbHmcf4nUF6C7Pj8M6MpY/Spark%3A-Flare-4.0---Core?node-id=10975%3A768&mode=dev)


packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
```diff
@@ -0,0 +1,112 @@
...
+  getSlottedInput
+} from '../../../stable/gux-form-field/gux-form-field.service';
+import { GuxSelectorCardVariant } from './gux-selector-card.types';
+import { calculateInputDisabledState } from '@utils/dom/calculate-input-disabled-state';
+import { onInputDisabledStateChange } from '@utils/dom/on-input-disabled-state-change';
+import { preventBrowserValidationStyling } from '@utils/dom/prevent-browser-validation-styling';
+
+@Component({
+  styleUrl: 'gux-selector-card.scss',
+  tag: 'gux-selector-card',
```
gavin-everett-genesys, line 15: I think the tag should be `gux-selector-card-beta`

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
```diff
@@ -0,0 +1,112 @@
...
+})
+export class GuxSelectorCard {
+  private input: HTMLInputElement;
+  private disabledObserver: MutationObserver;
+
+  @Element()
+  private root: HTMLElement;
+
+  @Prop()
+  variant: GuxSelectorCardVariant = 'simple';
```
gavin-everett-genesys, line 26: Does this property do anything ?

gavin-everett-genesys: From what I see in figma if its `descriptive` it will be left aligned and have a description area.

321gillian: I decided to just include the simple card styling at this point and leave the descriptive variant until a later card just to keep the PR smaller.

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  @State()
+  private disabled: boolean;
+
+  @State()
+  private checked: boolean;
+
+  componentWillLoad(): void {
+    this.setInput();
+    trackComponent(this.root);
```
gavin-everett-genesys, line 36: You can probably track the variant here also 
`trackComponent(this.root , { variant: this.variant })`

gavin-everett-genesys:
Will this only support radio buttons or will it support checkboxes also ?

321gillian:
> Will this only support radio buttons or will it support checkboxes also ?

My thoughts on the order or PRs for this is: 
1. Simple radio selector cards
2. Groupable simple radio selector cards
3. Checkbox simple cards
4. Descriptive cards

This PR is just for part 1

gavin-everett-genesys:
What is the difference between just using a `gux-card` and slotting what you want where you would not be constrained to specific implementation details over using the `gux-selector-card`  or am I missing something ?

daragh-king-genesys:
Just to add to @gavin-everett-genesys comment about `gux-card` we should check with UX if they see `gux-selector-card` and `gux-card` as related. At the moment I do think this functionality deserves its own component but maybe we should be using `gux-card` internally to keep consistency between the two components if UX sees them as related (from a design perspective)

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/example.html
```diff
@@ -0,0 +1,89 @@
...
+      <label slot="label" for="fourth">A longer card label example</label>
+      <input
+        slot="input"
+        id="fourth"
+        type="radio"
+        name="example4"
+        value="card-label"
+      />
+      <gux-icon slot="icon" icon-name="fa/diamond-regular" decorative />
+    </gux-selector-card-beta>
```
daragh-king-genesys, line 73: This api looks ok to me.

321gillian:
> Just to add to @gavin-everett-genesys comment about `gux-card` we should check with UX if they see `gux-selector-card` and `gux-card` as related. At the moment I do think this functionality deserves its own component but maybe we should be using `gux-card` internally to keep consistency between the two components if UX sees them as related (from a design perspective)

I spoke with Memo about this. While he did see it as related, the style is quite different in gux-card. It has 24px padding and 16px border-radius so quite different from gux-selector-card design. Different enough that I would say other than having the word 'card' in common, they don't really cross-over. 

Memo summarised it this way: 
> gux-cards is a plain surface component that you can put stuff on it
> and selector card is an input that looks like a card

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
```diff
@@ -0,0 +1,150 @@
...
+  validateFormIds,
+  getSlottedInput
+} from '../../../stable/gux-form-field/gux-form-field.service';
+import { GuxSelectorCardVariant } from './gux-selector-card.types';
+import { calculateInputDisabledState } from '@utils/dom/calculate-input-disabled-state';
+import { onInputDisabledStateChange } from '@utils/dom/on-input-disabled-state-change';
+import { preventBrowserValidationStyling } from '@utils/dom/prevent-browser-validation-styling';
+import { logWarn } from '@utils/error/log-error';
+import { setSlotAriaDescribedby } from '../../../stable/gux-form-field/gux-form-field.service';
+
```
gavin-everett-genesys, line 14: I think you should document your slots here like ,

```
/**
 * @slot label - Required slot for label
 * etc ...
 */
```

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  @Listen('click')
+  onClick(e: MouseEvent): void {
+    e.stopPropagation();
+    this.setChecked();
+  }
+
+  private setInput(): void {
+    this.input = getSlottedInput(this.root, 'input[slot="input"]');
```
gavin-everett-genesys, line 57: I think this would have to target specifically `radio` and `checkbox`  I think as it cant be a text input or anything like that ?

gavin-everett-genesys: If you update this function to be ,

`this.input = getSlottedInput(this.root, 'input[type="radio"][slot="input"]')` then the function `validateInputSlot` can be removed because if the input passed to the slot is not a type of `radio` then a warning message will be issued which is outlined in the `getSlottedInput` util.

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
```diff
@@ -0,0 +1,150 @@
...
+      this.input,
+      (disabled: boolean) => {
+        this.disabled = disabled;
+      }
+    );
+
+    validateFormIds(this.root, this.input);
+  }
+
+  private validateInputSlot(slottedElement: HTMLInputElement): void {
```
gavin-everett-genesys, line 75: This function can be removed if you make the changes to `setInput` in my other comment.

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
321gillian: Thanks for the tip! Made those changes now

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
```diff
@@ -0,0 +1,143 @@
...
+      if (slottedTagName === 'LABEL') {
+        return;
+      }
+      logWarn(this.root, `You must slot an label in the label slot.`);
+    }
+  }
+
+  private setChecked(): void {
+    if (this.disabled) {
+      return;
```
gavin-everett-genesys, line 97: I think to avoid this disabled check you could style the host element like with `pointer-events:none` when the gux-disabled class is applied. I think something like this is done in other components.

321gillian: Updated to apply this approach.

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  //TODO: COMUI-2784 Descriptive variant styling
+  @Prop()
+  variant: GuxSelectorCardVariant = 'simple';
+
+  @State()
+  private disabled: boolean;
+
+  @State()
+  private checked: boolean;
```
gavin-everett-genesys, line 43: Is this state needed? If its only used for the styling would `this.input.checked` work ?

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.scss
```diff
@@ -0,0 +1,192 @@
...
+      var(--gse-ui-selectorCard-selected-border-style)
+      var(--gse-ui-selectorCard-selected-border-color);
+  }
+}
+
+:host(.gux-disabled) {
+  cursor: not-allowed;
+
+  .gux-content {
+    pointer-events: none;
```
gavin-everett-genesys, line 29: I think you might also want to add 

```
  -webkit-user-select: none; /* Safari */
  user-select: none;
```

This is just to prevent the selection of text when it is disabled.

321gillian: done

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
321gillian: this didn't work for me.

gavin-everett-genesys: Discussed with gillian I dont think this will work actually because this.input.checked when changed does not cause a re-render so it wont work so with the checked state when it changes it causes a re-render.

gavin-everett-genesys (approved):

daragh-king-genesys (approved):

PR #419, opened 2024-04-17
fix(month-picker): Fixed tabbing issues

jason-evans-genesys:
✅ Closes: COMUI-1222

There was an issue with the tab order in this component. The ticket for this PR has a more detailed list of steps to check to confirm that the component tabbing functionality behaves correctly. You can test the component here:

https://apps.inindca.com/common-ui-docs/genesys-webcomponents/feature/COMUI-1222/#gux-month-picker

***Note**: I removed the focus trap code in the gux-month-calendar component since we just discussed as a team that the user should not be trapped in a component when pressing the tab key, since they'll need to get to outside areas like the address bar of the browser, etc. We will need to figure out the expected behavior if the user hits the tab key on the last element of the popover component. I will create a follow up ticket to investigate this.


packages/genesys-spark-components/src/components/beta/gux-month-picker/gux-month-calendar/gux-month-list/gux-month-list-item/gux-month-list-item.tsx
```diff
@@ -50,7 +50,6 @@ export class GuxMonthListItem {
           <button
             class={{ 'gux-selected': this.selected }}
             type="button"
-            tabIndex={-1}
```
jason-evans-genesys, line 53: It seems like this code I removed was initially here for a reason I'd imagine so I'm going to track down when and why this code was added in the first place.

katie-bobbe-genesys: It has a tab index of -1 because it was designed to use arrow navigation instead of tabbing through each of the months. Unless we have a reason to change the behavior here, I think the tabindex of -1 is intended behavior

jason-evans-genesys:
@katie-bobbe-genesys @gavin-everett-genesys thoughts on what I mentioned in the updated description of this PR?

gavin-everett-genesys:
> @katie-bobbe-genesys @gavin-everett-genesys thoughts on what I mentioned in the updated description of this PR?

Is the issue only present in firefox ? Is the issue that if you are on the right arrow button in the header and press tab it does not go back into the calendar ?

jason-evans-genesys:
> > @katie-bobbe-genesys @gavin-everett-genesys thoughts on what I mentioned in the updated description of this PR?
> 
> Is the issue only present in firefox ? Is the issue that if you are on the right arrow button in the header and press tab it does not go back into the calendar ?

Yeah, that's the current problem in firefox.

jason-evans-genesys:
> > @katie-bobbe-genesys @gavin-everett-genesys thoughts on what I mentioned in the updated description of this PR?
> 
> Is the issue only present in firefox ? Is the issue that if you are on the right arrow button in the header and press tab it does not go back into the calendar ?

@gavin-everett-genesys  I just pushed some code changes that handle moving focus from the next year arrow in the header to the selected month that looks to work in chrome and firefox.

jason-evans-genesys:
@321gillian @gavin-everett-genesys @katie-bobbe-genesys this PR is now ready for review again. Thank you.

katie-bobbe-genesys:
I see a few issues with this on the feature branch:
1. Tabbing outside of the open month picker does not close the month picker. I would expect that the month picker should close when tabbing outside of the month picker
2. Tab to the forward arrow and press enter to go to another year using the keyboard. Next, try to tab forward and you will find that you are stuck on the `>` forward button

jason-evans-genesys:
> https://apps.inindca.com/common-ui-docs/genesys-webcomponents/feature/COMUI-1222

@katie-bobbe-genesys do you mind if I tackle closing the month picker on focus out in a separate ticket to keep it isolated from this PR? Reason being is that I have another popover PR open that I would borrow code from to close the month-picker and wanted to close out that other PR and get the implementation all squared away with it before doing it in this component. Would that be ok? This is the other popover PR I'm referencing: https://github.com/MyPureCloud/genesys-spark/pull/444

jason-evans-genesys:
> I see a few issues with this on the feature branch:
> 
> 1. Tabbing outside of the open month picker does not close the month picker. I would expect that the month picker should close when tabbing outside of the month picker
> 2. Tab to the forward arrow and press enter to go to another year using the keyboard. Next, try to tab forward and you will find that you are stuck on the `>` forward button

Thanks @katie-bobbe-genesys

1. Do you mind if I tackle closing the month picker on focus out in a separate ticket to keep it isolated from this PR? Reason being is that I have another popover PR open that I would borrow code from to close the month-picker and wanted to close out that other PR and get the implementation all squared away with it before doing it in this component. Would that be ok? This is the other popover PR I'm referencing: [#444](https://github.com/MyPureCloud/genesys-spark/pull/444)
2. Put a fix in for this issue

packages/genesys-spark-components/src/components/beta/gux-month-picker/gux-month-calendar/gux-month-calendar.tsx
```diff
@@ -66,6 +67,18 @@ export class GuxMonthCalendar {
...
   locale: string;
 
+  @State()
+  previewValue: string;
+
+  /**
+   * Controls hiding and showing the month-calendar
+   */
+  @Prop({ mutable: true })
+  expanded: boolean = false;
```
jason-evans-genesys, line 77: @katie-bobbe-genesys Thoughts on my implementation of closing the calendar on focus out? I realize line 77 introduces a new prop called `expanded` and without the client using this the calendar will not open, which would be a breaking change. I'm trying to figure out a way to toggle visibility of the calendar without adding a new prop.

packages/genesys-spark-components/src/components/beta/gux-month-picker/gux-month-calendar/gux-month-calendar.tsx
```diff
@@ -94,6 +107,36 @@ export class GuxMonthCalendar {
...
+  }
+
+  @Listen('focusout')
+  onFocusout(event: FocusEvent): void {
+    const focusIsOutsideComponent = !this.root.contains(
+      event.relatedTarget as Node
+    );
+    if (focusIsOutsideComponent) {
+      this.expanded = false;
+      this.closeCalendar.emit();
```
jason-evans-genesys, line 136: @katie-bobbe-genesys This is how i notify the parent that the calendar is closed. I do this so that the parent can be told of the state of the calendar's visibility since the parent tracks that itself.

jason-evans-genesys:
> I see a few issues with this on the feature branch:
> 
> 1. Tabbing outside of the open month picker does not close the month picker. I would expect that the month picker should close when tabbing outside of the month picker
> 2. Tab to the forward arrow and press enter to go to another year using the keyboard. Next, try to tab forward and you will find that you are stuck on the `>` forward button

@katie-bobbe-genesys This is now ready for review again. Those 2 issues should hopefully be now addressed with the recent code changes I made. Thank you.

jason-evans-genesys:
> The keyboard functionality looks good, but I've found a few regressions when testing the feature with a mouse. I've captured a screen recording of the issues below.
> 
> 1. hovering over the selected month closes the calendar
> 2. clicking the calendar icon does not toggle the calendar open and closed. it just opens it each time it is clicked
> 
>  Screen.Recording.2024-05-24.at.11.17.18.mov

Thanks @katie-bobbe-genesys, I pushed fixes for this and this PR is now ready for review again.

katie-bobbe-genesys:
@jason-evans-genesys I am still seeing the "hovering over the selected month closes the calendar" in the feature branch. Maybe the conflicts in `gux-icon/readme.md` need to be resolved for the feature branch to be updated.

jason-evans-genesys:
> @jason-evans-genesys I am still seeing the "hovering over the selected month closes the calendar" in the feature branch. Maybe the conflicts in `gux-icon/readme.md` need to be resolved for the feature branch to be updated.

Hey @katie-bobbe-genesys so I looked into this more and this issue is actually still present. This seems to be happening because there is this code in `gux-month-list-item.tsx` below that will cause a focusOut event in the parent `gux-month-picker.tsx` component to trigger which will then close the popover:

```
if (parentList && parentList.shadowRoot.activeElement === null) {
      this.root.blur();
      parentList.focus();
    }
```

The root cause for this is that we are setting a tabIndex of `-1` for the month list item button so `document.activeElement` is not set because the browser isn't notified of the element being focused even though we called `target.focus();` in `gux-month-calendar.tsx`

Any ideas?

jason-evans-genesys:
@katie-bobbe-genesys The issue where hovering over the selected month in the calendar popover would close the popover is now resolved and can be tested here: https://apps.inindca.com/common-ui-docs/genesys-webcomponents/feature/COMUI-1222/#gux-month-picker

katie-bobbe-genesys:
image
I cannot tab forward when I am on this element (forward month arrow) . I would expect pressing `tab` would move focus to the calendar.

jason-evans-genesys:
> image I cannot tab forward when I am on this element (forward month arrow) . I would expect pressing `tab` would move focus to the calendar.

@katie-bobbe-genesys  Weird, it's working for me in chrome and firefox. Which browser are you on?

packages/genesys-spark-components/src/components/stable/gux-table/gux-sort-control/tests/__snapshots__/gux-sort-control.e2e.ts.snap
katie-bobbe-genesys, line 1: I wouldn't expect these snapshots to be different. Can you run the tests locally and make sure these snapshot changes for unrelated components are not included?

jason-evans-genesys: Great catch. I ran the tests locally and fixed the snapshot changes. Should be all good now.

packages/genesys-spark-components/src/components/beta/gux-month-picker/gux-month-calendar/gux-month-calendar.tsx
```diff
@@ -163,6 +193,9 @@ export class GuxMonthCalendar {
 
   private changeYear(increment: number): void {
     this.year = (parseInt(this.year) + increment).toString();
+    const month = getYearMonthObject(this.value).month;
```
katie-bobbe-genesys, line 196: this logic does not work in the case when the month picker has no initial set value. For a component without a set value, try changing the month by clicking the `>` to change the month. On the feature branch you will see in the console that there is an error being logged, because there is no set value

jason-evans-genesys: I added a fix here to disable the next/previous year buttons if `this.value` is empty.

jason-evans-genesys:
> LGTM 👍 , Aside from the tabbing stuff, I think the `.gux-display` class needs an `align-items:center` styling to align the `year` and `month` inputs.

Great catch. I just pushed a fix to align the `year` and `month` inputs.

packages/genesys-spark-components/src/components/beta/gux-month-picker/gux-month-calendar/gux-month-calendar.tsx
katie-bobbe-genesys: @jason-evans-genesys It doesn't make sense to me to disable the next/previous year buttons if `this.value` is empty. Having an empty value should not constrain a user to only choosing a month in the current year.

katie-bobbe-genesys (changes requested):
It doesn't make sense to me to disable the next/previous year buttons if `this.value` is empty. Having an empty value should not constrain a user to only choosing a month in the current year. Let me know if I'm missing something here or if there is UX guidance pointing to this behavior

katie-bobbe-genesys:
The keyboard focus gets lost when there is a min/max date set. When a user uses a keyboard to select the forward or backward arrows to change the year, and the next year is disabled, there is not a way to focus into the calendar. 

https://github.com/user-attachments/assets/fd13c9be-4656-492d-9dd6-4a31938ca3cd

jason-evans-genesys:
> The keyboard focus gets lost when there is a min/max date set. When a user uses a keyboard to select the forward or backward arrows to change the year, and the next year is disabled, there is not a way to focus into the calendar.
> Screen.Recording.2024-07-16.at.14.41.07.mov

@katie-bobbe-genesys pushed a fix for this.

katie-bobbe-genesys (approved):

PR #420, opened 2024-04-17
chore(regional-dates): backport to v3

katie-bobbe-genesys:


packages/genesys-spark-components/src/genesys-spark-utils/get-closest-element.ts
```diff
@@ -0,0 +1,24 @@
+//https://inindca.atlassian.net/browse/COMUI-2673
```
321gillian, line 1: Might be worth adding a bit of context to this. Also we should consider marking these type of comments as TODO: so we can use the highlighting feature in VSCode. https://marketplace.visualstudio.com/items?itemName=wayou.vscode-todo-highlight

packages/genesys-spark-components/src/genesys-spark-utils/intl.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  if (locale != undefined) {
+    return new Intl.DateTimeFormat(locale as string, options);
+  } else {
+    const userLocale = determineDisplayLocale();
+    return new Intl.DateTimeFormat(userLocale, options);
+  }
+}
+/**
```
321gillian, line 29:
```suggestion
}

/**
```

packages/genesys-spark/src/utils/intl.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  if (locale != undefined) {
+    return new Intl.DateTimeFormat(locale as string, options);
+  } else {
+    const userLocale = determineDisplayLocale();
+    return new Intl.DateTimeFormat(userLocale, options);
+  }
+}
+/**
+ * Provides an [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat)
```
321gillian, line 30:
```suggestion
}

/**
 * Provides an [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat)
```

packages/genesys-spark-components/src/genesys-spark-utils/get-closest-element.ts
321gillian: Ah ok, there's already a couple in v3 and a whole bunch in v4 formatted this way. Should we advise against that then?

daragh-king-genesys (approved):

321gillian (approved):

PR #424, opened 2024-04-22
feat(form) - addition of form component (FEEDBACK)

gavin-everett-genesys:
Open to feedback on this API,

Here are some questions I had :

1. Could the `gux-form-footer` be used for more purposes outside of `gux-form` at the moment I just have it as an internal component within `gux-form` but maybe it has a more general purpose to be a standalone component to be used elsewhere in other components.
2. Need to discuss with design on how `vertical` and `horizontal` stacking will work within fieldsets.
3. Within the example I did not include and example of the checkbox group as I am awaiting that to be completed.

I will add tests once we are happy with the API.


packages/genesys-spark-components/src/components/beta/gux-form/example.html
```diff
@@ -0,0 +1,50 @@
+<gux-form-beta>
+  <form>
```
321gillian, line 2: This is more just a question for my own understanding. Why create a gux-form-beta that's just a slot rather than having it contain a form tag?

gavin-everett-genesys: If I was to to put the form tag inside of `gux-form-beta` developers would not be able to access the the form tag to perform actions they may need to do as it would be in the `shadowDOM`.

packages/genesys-spark-components/src/components/beta/gux-form/gux-form-footer/gux-form-footer.tsx
```diff
@@ -0,0 +1,36 @@
...
+  placement: GuxFormFooterPlacement = 'page-desktop';
+
+  componentWillLoad(): void {
+    trackComponent(this.root, { variant: this.placement });
+  }
+
+  render(): JSX.Element {
+    return (
+      <footer
+        class={{
```
gavin-everett-genesys, line 28: Open to opinions on this should we have the footer element in the light DOM or have it shadowDOMed. I left it here for now as I had all the stylings setup and wanted to make sure. I am not sure if devs can do anything directly with the footer element so encapsulating might be fine ?

packages/genesys-spark-components/src/components/beta/gux-form/gux-form-description/gux-form-description.tsx
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+@Component({
+  styleUrl: 'gux-form-description.scss',
+  tag: 'gux-form-description',
+  shadow: true
+})
+export class GuxFormDescription {
+  render(): JSX.Element {
+    return (<slot></slot>) as JSX.Element;
```
daragh-king-genesys, line 14: I think because we are using `::slotted(*)` we can't use the default slot. That type of slot will accept a text node which that css selector will not target.
We should discuss this at standup as it has become a common issue.

packages/genesys-spark-components/src/utils/dom/wrap-text-nodes.ts
```diff
@@ -0,0 +1,12 @@
+export function wrapTextNodes(element: HTMLElement): void {
```
gavin-everett-genesys, line 1: Want to know peoples thoughts on creating a helper function which will wrap text nodes in a span. I am currently using this in `gux-form-heading` with this function you will be able to pass both elements and text nodes at the same time and the styling will apply.

MattCheely: This creates the span in the light DOM, right? If so, I don't think we can support this, as our changes will be in conflict with what the application is trying to render. I think that's one of our rules that we can't make exceptions to - the component's aren't allowed to change the structure of the light DOM, as it's owned by the hosting app/page.

gavin-everett-genesys: Thats true I was thinking there will be caveats with this solution but trying to avoid the named slot route. Maybe logging a warning is another option but will continue to look. As of now there is now selector to style text nodes.

packages/genesys-spark-components/src/components/beta/gux-form/gux-form-description/gux-form-description.tsx
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+@Component({
+  styleUrl: 'gux-form-description.scss',
+  tag: 'gux-form-description',
+  shadow: true
+})
+export class GuxFormDescription {
+  render(): JSX.Element {
+    return (<slot name="description"></slot>) as JSX.Element;
```
gavin-everett-genesys, line 14: I chose the route of named slots. I dont  think there is many solutions to solve this problem. I also considered the the logging of an error/warning but it would sort of be the same thing as you would want them to slot a element for the styles to apply correctly.  Open to opinions on this.

packages/genesys-spark-components/src/components/beta/gux-form/example.html
```diff
@@ -0,0 +1,50 @@
...
+    <header>
+      <gux-form-heading><h1>Form Heading</h1></gux-form-heading>
+      <gux-form-description
+        ><p>
```
daragh-king-genesys, line 6: Do we have to slot a 'p' or could we use a 'span'?

packages/genesys-spark-components/src/components/beta/gux-form/gux-form-footer/gux-form-footer.tsx
daragh-king-genesys: As at the moment the `header` element is in the light dom I think we should do the same with the `footer` element.

gavin-everett-genesys: By this do you mean to get rid of the `gux-form-footer` completely or just remove the `footer` from the shadowDOM and slot it instead ?

gavin-everett-genesys: I removed it from the shadowDOM so now it can just be slotted to the `gux-form-footer`

gavin-everett-genesys: The `gux-form-footer` will have to stay though I think due to there being different placements.

packages/genesys-spark-components/src/components/beta/gux-form/gux-form.light.scss
```diff
@@ -0,0 +1,97 @@
...
+        font-size: var(
+          --gse-semantic-forms-formHeader-text-fontSize
+        ) !important;
+        font-weight: var(
+          --gse-semantic-forms-formHeader-text-fontWeight
+        ) !important;
+        line-height: var(
+          --gse-semantic-forms-formHeader-text-lineHeight
+        ) !important;
+        color: var(--gse-semantic-forms-formHeader-foregroundColor) !important;
```
321gillian, line 26: Why do we need to use !important for all of these?

gavin-everett-genesys: I think using `!important` is the only way to overide the default stylings of for example a `h1`. 
Screenshot 2024-06-12 at 09 32 36
If I am not to use `!important` the default `h1` will take precedence as shown in the image above.

daragh-king-genesys (approved):

daragh-king-genesys:
Notify UX that this is now available for review

packages/genesys-spark-components/src/components/beta/gux-form/gux-form.light.scss
321gillian: Ah ok, cool. I had a look there and it looks to me like the style from the gux-form-heading component for slotted(*) takes precedence over this. Do both sets of styles have different purposes?

gavin-everett-genesys: The `::slotted(*)` styles are just specific to this design for the forms. The other `h1` styles are for if you were to use a `h1` in a component for example like the card component or something so they are aligned with the spark default styles.

321gillian: Sorry, what I'm trying to ask is if we have this coming from the component gux-form-header as in the image do we also need to have them in gux-form.light? 
Screenshot 2024-06-12 at 14 02 41

gavin-everett-genesys: Oh yeah, so the reasoning why we have this styling in two places is because the `:slotted(*)` selector in `gux-form-heading` cannot target text nodes so for example if you do something like,

`<gux-form-heading>Hello</gux-form-heading>` 

the `::slotted(*)` wont be able to target it. I think its a know issue but the only way around it is to also style it in the `gux-form.light.scss` file so we can cover both if the user slots a text node or slots and element.

321gillian: ah ok, that makes sense.

PR #425, opened 2024-04-22
feat(form-fields): Form label font-weight change and addition of optional indicator

321gillian:
I've updated this now with all of the examples, snapshot updates, prop updates etc for the form field label change. 

Should change on: 

- form-field-color
- form-field-dropdown
- form-field-number
- form-field-phone
- form-field-range
- form-field-search
- form-field-select
- form-field-text-like
- form-field-textarea
- form-field-time-picker
- form-field-time-zone-picker

As far as I recall, most of these will use gux-form-field-label and gux-form-field-legend so it comes via that plus added props.


gavin-everett-genesys:
Will have to figure out which one will take precedence over the other for example will required take precedence over optional etc if both properties are supplied to the component ?

In you current working example if we supply both the optional indicator and required nothing seems to happen Id assume required would take precedence but I think thats a UX question.

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.tsx
```diff
@@ -90,9 +96,12 @@ export class GuxFormFieldColor {
       <GuxFormFieldContainer labelPosition={this.computedLabelPosition}>
         <GuxFormFieldLabel
           position={this.computedLabelPosition}
-          required={this.required}
+          required={
+            this.indicatorMark === 'optional-only' ? false : this.required
+          }
```
gavin-everett-genesys, line 101: Based on my comment about which property should take precedence below I think this could just stay the same as before like `this.required=this.required` and then inside `renderOptionalIndicator` the conditional would be `if(this.indicatorMark && !this.required` , the required would take precedence over the indicatorMark if both were present. Let me know what you think.

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.tsx
```diff
@@ -45,6 +48,9 @@ export class GuxFormFieldColor {
   @Prop()
   labelPosition: GuxFormFieldLabelPosition;
 
+  @Prop()
+  indicatorMark: GuxFormFieldIndicatorMark = 'required-only';
```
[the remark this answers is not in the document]
jason-evans-genesys: If `required` on the input is true will that means `indicatorMark` is also always true after this change is made? I'm just trying to make sure I understand all of the logic here and what the purpose of both `required` on the input and `indicatorMark` being used instead of only just `required` on the input?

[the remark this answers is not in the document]
321gillian: I discussed this with Daragh previously and we agreed that while it kind of looks obvious to make this a boolean property, doing so would make it much more difficult if an additional style needed to be implemented further down the line. 

Since 'required-only' is the default, this will not impact anyone currently using this component but will also support the scenario where the (optional) mark is shown. 

The required on the input will have a bearing on the display depending on what's chosen for the indicator mark. 

Here's a rough list of scenarios: 
input === 'required' && indicator-mark === 'optional-only' then do not show asterisk
input !== 'required' && indicator-mark === 'optional-only' then show (optional)

input === 'required' && indicator-mark === 'required-only' then show asterisk
input !== 'required' && indicator-mark === 'required-only' then  do not show (optional)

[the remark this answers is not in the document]
jason-evans-genesys: Makes sense, thanks!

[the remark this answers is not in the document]
jason-evans-genesys: Is there any way to document those scenarios somewhere in the component so it's more obvious?

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.tsx
```diff
@@ -90,9 +96,12 @@ export class GuxFormFieldColor {
       <GuxFormFieldContainer labelPosition={this.computedLabelPosition}>
         <GuxFormFieldLabel
```
daragh-king-genesys, line 97: I think we should add a `indicatorMark` property to this component and keep the required and optional stuff together.

321gillian: Moved that stuff into the label component

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.tsx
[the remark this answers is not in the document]
321gillian: @jason-evans-genesys I added a comment to the prop in the field and I think it's maybe a little clearer in the updated implementation now. Let me know if you think I could add in some more comments and stuff here

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.tsx
```diff
@@ -45,6 +48,14 @@ export class GuxFormFieldColor {
   @Prop()
   labelPosition: GuxFormFieldLabelPosition;
 
+  /**
+   * Field indicator mark which can show *, (optional) or blank
+   * Defaults to required, this shows * for required fields and blank for optional
```
jason-evans-genesys, line 53: Thanks for adding these comments. Super helpful! Is there any way we can we modify the wording on line 53 slightly? 

Maybe:

`Defaults to required, this shows * for required fields and blank for optional` -> `Defaults to required. The component will show * for required fields and blank for optional`

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.tsx
```diff
@@ -45,6 +48,14 @@ export class GuxFormFieldColor {
...
+   * When set to optional, shows (optional) for optional and blank for required.
```
jason-evans-genesys, line 54: Maybe a slight change to be:

`When set to optional, the component will display (optional) for optional and blank for required.`

packages/genesys-spark-components/src/components/stable/gux-form-field/functional-components/gux-form-field-label/gux-form-field-label.tsx
```diff
@@ -1,25 +1,32 @@
...
 
 interface GuxFormFieldLabelProps {
   position: GuxFormFieldLabelPosition;
   required: boolean;
+  indicatorMark?: GuxFormFieldIndicatorMark;
 }
 
 export const GuxFormFieldLabel: FunctionalComponent<GuxFormFieldLabelProps> = (
-  { position, required },
+  { position, required, indicatorMark = 'required' },
```
jason-evans-genesys, line 15: thoughts on setting the default value of `required` on line 11 to make it more clear?

packages/genesys-spark-components/src/i18n/translations/en.json
```diff
@@ -99,6 +99,9 @@
     "increment": "Increment",
     "decrement": "Decrement"
   },
+  "gux-form-field-optional-indicator": {
+    "optional": "optional"
```
jason-evans-genesys, line 103: Does optional need to uppercase or is lowercase expected?

321gillian: good question, it's lowercase by design

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.tsx
```diff
@@ -45,6 +48,14 @@ export class GuxFormFieldColor {
   @Prop()
   labelPosition: GuxFormFieldLabelPosition;
 
+  /**
+   * Field indicator mark which can show *, (optional) or blank
+   * Defaults to required. When set to required, the component will display * for required fields and blank for optional
+   * When set to optional, the component will display (optional) for optional and blank for required.
+   */
+  @Prop()
+  indicatorMark: GuxFormFieldIndicatorMark;
```
gavin-everett-genesys, line 57: I dont think this currently defaults to `required`. I think it would have to be `indicatorMark: GuxFormFieldIndicatorMark = 'required'`

321gillian: The default is set in the label component so I don't think it's needed here.

321gillian:
> We Will have to figure out which one will take precedence over the other for example will required take precedence over optional etc if both properties are supplied to the component ?
> 
> In you current working example if we supply both the optional indicator and required nothing seems to happen Id assume required would take precedence but I think thats a UX question.

The idea is that if the form has majority optional fields then indicate only required. 

If the form has majority required fields, then the option is there to highlight only the optional ones.

gavin-everett-genesys:
> > We Will have to figure out which one will take precedence over the other for example will required take precedence over optional etc if both properties are supplied to the component ?
> > In you current working example if we supply both the optional indicator and required nothing seems to happen Id assume required would take precedence but I think thats a UX question.
> 
> The idea is that if the form has majority optional fields then indicate only required.
> 
> If the form has majority required fields, then the option is there to highlight only the optional ones.

So for example if I was to use this markup,

 ` <gux-form-field-number indicator-mark="optional">
    <input slot="input" type="number" name="lp-1" value="10" required />
    <label slot="label">Indicator Mark</label>
  </gux-form-field-number>`

Where the input has a `required`  attribute but the `gux-form-field-number` element sets the `indicator-mark` to optional nothing will show ?

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/gux-form-field-color.tsx
gavin-everett-genesys: Ahh okay makes sense, is there a reason for setting it within the functional component ? I guess it fine as there is a comment explaining that it is set to `required` by default.

321gillian: Didn't really have a preference but I updated it to take the default from the component instead.

321gillian:
Closing this for now until the first one is done

PR #429, opened 2024-04-24
fix(table-toolbar): fix initial glitchy render

gavin-everett-genesys:
I am not 100% happy with this solution but it will prevent the initial glitchy render on-load when the component is trying to figure out what layout to use. I wasnt sure should I dig deep into a re-design of how it renders it as its probably a big enough refactor potentially.

Let me know your thoughts.
https://inindca.atlassian.net/browse/COMUI-2200
✅ Closes: COMUI-2200

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.scss
@@ -1,4 +1,15 @@
-@use '~genesys-spark/src/style/mixins.scss';
+@mixin gux-toolbar-prevent-initial-flicker-workaround {
+  opacity: 0;
+  animation-name: show-toolbar;
+  animation-delay: 0.5s;
+  animation-fill-mode: forwards;
+
+  @keyframes show-toolbar {
+    to {
+      opacity: 1;
+    }
+  }
+}
 
 :host {
   display: flex;
@@ -8,7 +19,7 @@
   width: 100%;
   max-height: var(--gse-ui-dataTableItems-tableToolbar-height);
 
-  @include mixins.gux-toolbar-prevent-initial-flicker-workaround;
+  @include gux-toolbar-prevent-initial-flicker-workaround;
 
   .search-filter-container {
     min-width: 260px;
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.scss
```diff
@@ -5,6 +5,10 @@
   justify-content: space-between;
   width: 100%;
   max-height: var(--gse-ui-dataTableItems-tableToolbar-height);
+  opacity: 0;
+  animation-name: show-toolbar;
+  animation-delay: 0.5s;
+  animation-fill-mode: forwards;
```
daragh-king-genesys, line 11: Not a huge fan of this approach either but it is better what we have now so worth doing I think.
I do think we should move this into a mixin so that we can name it as this is a workaround and someone in the future will wonder why it was added.

packages/genesys-spark/src/style/_mixins.scss
```diff
@@ -49,3 +49,16 @@
...
+  animation-name: show-toolbar;
+  animation-delay: 0.5s;
+  animation-fill-mode: forwards;
+
+  @keyframes show-toolbar {
+    to {
+      opacity: 1;
+    }
+  }
+}
```
321gillian, line 64: Just for my own learning, if this is only used in one place why is it a mixin?

gavin-everett-genesys: @daragh-king-genesys  suggested above to make it mixin so that future devs will be able to know what it is for because at first glance you probably would be wondering why there is a delay on showing the toolbar.

gavin-everett-genesys: You can reproduce this glitchy initial render if you go to the live docs and go to the condensed layout or even the icon only mode for the toolbar. When you refresh the page you will see a flicker for a split second while the component is trying to calculate the correct layout to use.

321gillian: Ah ok, how would we qualify what gets included in general mixins vs what is a mixin declared in a component scss file? 

Might be worth adding a comment with a ref to the ticket somewhere just since it's not that intuitive, like you were saying.

gavin-everett-genesys: Actually good spot I am not sure why I put it in the general mixins :) I will just define in in the component file

katie-bobbe-genesys (approved):

321gillian (approved):

daragh-king-genesys (approved):

PR #433, opened 2024-04-25
fix(icon): Reset icons updated last week and map old names to deprecated icons [WIP]

321gillian:
This resets 3 icons that changed visually last week. 

The fa/microsoft svg was changed to resemble what seems to be the fa/windows icon
play-regular and file-regular changed slightly to look like 'control-play' and 'file' deprecated icons rather than the current (as I understand it from FA versions), I replaced these with what was there prior to the change. 
I also mapped the old names to their deprecated equivalents. 

# Previous changes (which this will revert)
## play-regular 
Screenshot 2024-04-25 at 12 10 45

## file-regular
Screenshot 2024-04-25 at 12 08 30

## microsoft
Screenshot 2024-04-25 at 12 07 48


daragh-king-genesys:
Will have to discuss this with UX. AFAIK all these changes were requested by UX.

321gillian:
Closing this on the basis that the other icon work is happening

PR #438, opened 2024-04-29
fix(toggle): Add option to click on label

321gillian:
✅ Closes: COMUI-2807

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-toggle/gux-toggle.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-toggle/gux-toggle.tsx
@@ -165,16 +165,15 @@ export class GuxToggle {
             'gux-disabled': this.disabled || this.loading
           }}
         >
-          <div class="gux-toggle-input">
+          <div class="gux-toggle-input" onClick={this.toggle.bind(this)}>
             <gux-toggle-slider
               checked={this.checked}
               disabled={this.disabled || this.loading}
               guxAriaLabel={this.getAriaLabel()}
-              onClick={this.toggle.bind(this)}
-              onKeyDown={this.onKeydown.bind(this)}
               labelId={
                 this.checkedLabel && this.uncheckedLabel ? this.labelId : ''
               }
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-toggle/gux-toggle.tsx
```diff
@@ -165,16 +165,15 @@ export class GuxToggle {
...
             <gux-toggle-slider
               checked={this.checked}
               disabled={this.disabled || this.loading}
               guxAriaLabel={this.getAriaLabel()}
-              onClick={this.toggle.bind(this)}
-              onKeyDown={this.onKeydown.bind(this)}
               labelId={
                 this.checkedLabel && this.uncheckedLabel ? this.labelId : ''
               }
+              onKeyDown={this.onKeydown.bind(this)}
```
conor-darcy, line 176: @321gillian a bit pedantic, but I would keep the 2 events together on the same element if possible just so we know that they are both there.

conor-darcy (approved):

packages/genesys-spark-components/src/components/stable/gux-toggle/gux-toggle.tsx
jordanstith15: @conor-darcy If the point is to allow more clickable area than just the toggle-slider (can be useful in some situations), I'm not sure if it's possible. The key down would be used with keyboard navigation and would need to be focused on the gux-toggle-slider to work afaik.

jordanstith15 (approved):

packages/genesys-spark-components/src/components/stable/gux-toggle/gux-toggle.tsx
321gillian: @conor-darcy As @jordanstith15 said, in this case the keydown is linked to the focusable gux-toggle-slider. The click applies to both items in the gux-toggle-input.

PR #444, opened 2024-04-30
fix(popover): Closing popover on focus out

jason-evans-genesys:
✅ Closes: COMUI-1636

For now, you can paste this for the example page content to test the functionality added for this PR.

To test this PR, you can move focus into an element in the popover using the tab key to get there, and then move focus out of the popover by using the tab key or shift + tab and you should see the popover become hidden when you move focus out of the popover.

```
<h1>gux-popover</h1>
<div class="scroll-container">
  <div class="scroll-interior">
    <div id="popover-target">Example Element</div>
    <gux-popover
      id="popover-example"
      position="top"
      for="popover-target"
      is-open
    >
      <span slot="title">Title</span>
      <div class="popover-content-wrapper">
        <div>Popover Content</div>
        <div class="popover-content">
          <gux-button accent="ghost" onclick="notify(event)">
            Action 3
          </gux-button>
          <gux-button accent="secondary" onclick="notify(event)">
            Action 2
          </gux-button>
          <gux-button accent="primary" onclick="notify(event)">
            Action 1
          </gux-button>
        </div>
      </div>
    </gux-popover>
  </div>
</div>

<style
  onload="(function () {
  const popoverTarget = document.getElementById('popover-target');
  const popoverExample = document.getElementById('popover-example');

  popoverTarget.addEventListener('mouseover', () => {
    popoverExample.isOpen = true;
  });
})()"
>
  .scroll-container {
    width: 900px;
    height: 300px;
    margin: auto;
    overflow: scroll;
    border: 1px solid #aaa;
  }

  .scroll-interior {
    position: relative;
    width: 600px;
    height: 600px;
    background-color: #fff;
  }

  .popover-content-wrapper {
    display: flex;
    flex-direction: column;
    gap: 16px;
  }

  .popover-content {
    display: flex;
    flex-direction: row;
    gap: 8px;
    width: 280px;
  }

  #popover-target {
    position: absolute;
    top: 250px;
    left: 250px;
    display: inline-block;
    padding: 10px 15px;
    background-color: #fdfdfd;
    border: 1px #e4e9f0;
    box-shadow: 0 0 2px fade(#202937, 24%);
  }
</style>
```

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
@@ -86,6 +86,20 @@ export class GuxPopover {
     }
   }
 
+  @Listen('focusout')
+  onFocusout(event: FocusEvent): void {
+    if (!this.closeOnClickOutside && this.displayDismissButton) {
+      return;
+    }
+
+    const focusIsOutsidePopover = !this.root.contains(
+      event.relatedTarget as Node
+    );
+    if (focusIsOutsidePopover) {
+      this.dismiss();
+    }
+  }
+
   @OnClickOutside({ triggerEvents: 'mousedown' })
   checkForClickOutside(event: MouseEvent) {
     const clickPath = event.composedPath();


packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
```diff
@@ -86,6 +86,13 @@ export class GuxPopover {
     }
   }
 
+  @Listen('focusout')
```
katie-bobbe-genesys, line 89: The existing behavior is that if both `close-on-click-outside` and `display-dismiss-button`  are set to `true`, then clicking outside of the popover does cause the popover to close. With this change added to close on focusout, this behavior is no longer preserved (clicking outside is focusing out) I think this could be considered a breaking change?


image

jason-evans-genesys: @katie-bobbe-genesys great catch. I pushed code to try and handle this scenario. The new code I wrote should handle these scenarios:

1. `close-on-click-outside` is `false` and `display-dismiss-button` is `true`
- User tabs out of popover -> popover closes
- User clicks out of popover -> popover does not close
2. `close-on-click-outside` is `true`
- User tabs out of popover -> popover closes
- User clicks out of popover -> popover closes
3. `close-on-click-outside` is `false` and `display-dismiss-button` is `false`
- User tabs out of popover -> popover closes
- User clicks out of popover -> popover closes

katie-bobbe-genesys:
```
close-on-click-outside is false and display-dismiss-button is true
User tabs out of popover -> popover closes
User clicks out of popover -> popover does not close
```
I may be mistaken, but I was thinking the behavior should be that if `close-on-click-outside` is false and `display-dismiss-button` is true:
User tabs out of popover -> popover does not close
User clicks out of popover -> popover does not close

If this is the case, I believe we just need to check `if !close-on-click-outside and display-dismiss-button` then do not close on focus out.  In all other cases, we want to close on focus out. I don't think we need to add the check for whether `tab` is pressed or not. Let me know if I'm missing something

jason-evans-genesys: @katie-bobbe-genesys  Great catch! I pushed up a fix for this just now.

jason-evans-genesys: > ```
> close-on-click-outside is false and display-dismiss-button is true
> User tabs out of popover -> popover closes
> User clicks out of popover -> popover does not close
> ```
> 
> I may be mistaken, but I was thinking the behavior should be that if `close-on-click-outside` is false and `display-dismiss-button` is true: User tabs out of popover -> popover does not close User clicks out of popover -> popover does not close
> 
> If this is the case, I believe we just need to check `if !close-on-click-outside and display-dismiss-button` then do not close on focus out. In all other cases, we want to close on focus out. I don't think we need to add the check for whether `tab` is pressed or not. Let me know if I'm missing something

@katie-bobbe-genesys This is now ready for review again. Thank you.

PR #446, opened 2024-05-07
(form-footer) - tokens

gavin-everett-genesys:


packages/genesys-spark-tokens/snapshot/css/gse-ui.css
```diff
@@ -294,4 +294,23 @@
...
+  --gse-ui-formFooter-page-mobile-gap: 16px;
+  --gse-ui-formFooter-separator-color: #c6c8ce;
+  --gse-ui-formFooter-separator-width: 1;
+  --gse-ui-formFooter-separator-style: solid;
+  --gse-ui-formFooter-sideSheet-desktop-bottomPadding: 16px;
+  --gse-ui-formFooter-sideSheet-desktop-horizontalPadding: 24px;
+  --gse-ui-formFooter-sideSheet-desktop-gap: 16px;
+  --gse-ui-formFooter-modal-desktop-bottomPadding: 16px;
+  --gse-ui-formFooter-modal-desktop-horizontalPadding: 32px;
+  --gse-ui-formFooter-modal-desktop-gap: 32px;
```
conor-darcy, line 311: I don't think this is the right place for the page & modal tokens and I think the desktop & mobile tokens are not ideal either.
Should page/modal tokens be specific to those components or patterns?
`--gse-ui-formFooter-modal-desktop-bottomPadding`
should be
`--gse-ui-modal-formFooter-paddingBottom`
or possibly
`--gse-ui-modal-form-footer-paddingBottom`

Here's the [component naming convention](https://intranet.genesys.com/pages/viewpage.action?spaceKey=PureCloud&title=COMUI-1733%3A+Design+token+naming+conventions):
`--[domain]-[set]-[component]-[sub-component]-[variant]-[state]-[type]-[category]`

For me, the footer is a subcomponent of the form. But if it needs to change based on its grandparent than the grandparent should apply the style if possible.

Maybe this needs a wider conversation.

gavin-everett-genesys: @JackieYada

JackieYada: I understand where the suggestion of putting form footer under modal comes from but I don't think this is the ideal place as it's not actually only for modals. Modals will be using the form footer component. There is 3 types of formfooter for desktop (modal, sidesheet and page) and one for mobile, that's why we separated like this. 
I could move the form footer under forms and inputs which will be --gse-formandinputs-formFooter-modal-desktop-bottomPadding

conor-darcy (approved):

PR #455, opened 2024-05-15
feat(table): improving table row hover

conor-darcy:
Additional CSS rule to make the table row hover on the darker rows more robust & difficult to override.

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-table/gux-table.light.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-table/gux-table.light.scss
@@ -18,12 +18,6 @@ gux-table {
       color: var(--gse-ui-dataTableItems-cell-foregroundColor);
       text-align: left;
 
-      &:hover {
-        background-color: var(
-          --gse-ui-dataTableItems-cell-hoverBackgroundColor
-        );
-      }
-
       gux-icon {
         color: var(--gse-ui-dataTableItems-cell-icon-foregroundColor);
       }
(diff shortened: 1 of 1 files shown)


conor-darcy:
@gavin-everett-genesys Yeah, not sure why line 21 that is there so I thought I'd just leave it.
But I think they are processing their CSS, maybe minimising it & its combining duplicate declarations into a single rule.

Maybe we should do that on our end & deliver minimised CSS but that's a conversation for another day.

gavin-everett-genesys:
I am not sure we should be adding something if we are not 100% sure it fixes the issue. Maybe bring it up in standup.

packages/genesys-spark-components/src/components/stable/gux-table/gux-table.light.scss
```diff
@@ -101,6 +101,12 @@ gux-table {
```
daragh-king-genesys, line 101: This could be simplified a small bit more to
```
tr {
  background-color: var(--gse-ui-dataTableItems-cell-defaultBackgroundColor);

  &:hover {
    background-color: var(--gse-ui-dataTableItems-cell-hoverBackgroundColor);
  }

  &:nth-child(2n) {
    background-color: var(--gse-ui-dataTableItems-cell-altBackgroundColor);
  
    &:hover {
      background-color: var(--gse-ui-dataTableItems-cell-hoverBackgroundColor);
    }
  }
}
```
The second hover block does look odd/redundant so we might have to add a comment to stop it getting removed in the future.

gavin-everett-genesys (approved):

321gillian (approved):

PR #468, opened 2024-05-21
feat(selector-card): Add descriptive variant for radio selector card

321gillian:
[Figma Link](https://www.figma.com/design/JKbHmcf4nUF6C7Pj8M6MpY/Spark%3A-Flare-4.0---Core?node-id=10975-2034&m=dev)

✅ Closes: COMUI-2784


packages/genesys-spark-components/src/components/beta/gux-selector-cards/example.html
```diff
@@ -31,3 +31,62 @@ <h2>Simple Selector Card Group</h2>
     <gux-icon slot="icon" icon-name="fa/play-regular" decorative />
   </gux-selector-card-beta>
 </gux-selector-cards-beta>
+
+<h2>Descriptive Selector Card Group</h2>
+<gux-selector-cards-beta class="example2">
+  <gux-selector-card-beta variant="descriptive">
+    <label slot="label" for="first">First</label>
+    <input slot="input" id="first" type="radio" name="example2" />
+    <span slot="content"
```
daragh-king-genesys, line 40: I think `description` might be a better slot name

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.scss
```diff
@@ -189,3 +189,61 @@ $gux-icon-radio-checked: "data:image/svg+xml,%3Csvg viewBox='0 0 16 16' xmlns='h
     @include mixins.gux-sr-only;
   }
 }
+
+:host(.gux-descriptive) {
```
daragh-king-genesys, line 193: I think it is a bad idea to have these classes applied to the host element.
Is there a reason we need to use the `Host` element?

daragh-king-genesys (commented):
LGTM just a few nit picks :)

packages/genesys-spark-components/src/components/beta/gux-selector-cards/example.html
321gillian: Done

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
```diff
@@ -91,24 +94,26 @@ export class GuxSelectorCard {
...
+          <gux-truncate class="gux-label" max-lines={2}>
             <slot name="label" />
           </gux-truncate>
           <slot name="input" />
-          {this.renderContent()}
+          {this.renderDescription()}
           {this.renderBadge()}
         </div>
-      </Host>
+      </div>
```
321gillian, line 116: @daragh-king-genesys @gavin-everett-genesys I have a problem with this. As it currently stands, I have the input running over everything with a `position: absolute;`. The idea of this was to avoid adding a `checked` state property which has to be updated in JavaScript based on the input. This is so the card border can change color appropriately. 

However, doing it this way blocks the tooltip for truncated text from showing because the stacking context gives priority to the `input`. 

Do either of you know of any way I can make the tooltips available without going back to a `checked` state type thing (to support changing the card border when input is selected)? I've looked at a bunch of stuff but nothing is jumping out at me so far. Any thoughts or rough guess appreciated.

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.scss
```diff
@@ -118,74 +106,146 @@ $gux-icon-radio-checked: "data:image/svg+xml,%3Csvg viewBox='0 0 16 16' xmlns='h
...
+        height: var(--gse-ui-icon-size-lg);
+        margin-bottom: var(--gse-ui-selectorCard-descriptive-gap);
+      }
     }
   }
 
-  .gux-screenreader {
-    @include mixins.gux-sr-only;
+  &.gux-disabled {
+    cursor: not-allowed;
```
gavin-everett-genesys, line 241: I think `pointer-events:none` will be needed here as you can still select a `disabled` radio button.

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
gavin-everett-genesys: Will check this out 👍

gavin-everett-genesys: If you just apply a higher `z-index` to the truncate component would this work without anything breaking for example a `z-index:2` potentially ?

321gillian: I spent so long thinking about that and not getting the z-index to work. I think it should work now. Thanks Gavin!

gavin-everett-genesys:
Just one small issue see image below, 

Screenshot 2024-05-27 at 09 13 43

You can see when hovering over the second span that the text overlaps other text on the frame. I think currently the two spans inside have the same `z-index` but I dont think that will work for some reason. I am not sure if the second span needs a `z-index` or maybe it needs to be higher potentially.

gavin-everett-genesys:
One last issue I should have picked this up in my first review 🙈, 
Screenshot 2024-05-28 at 10 33 36

I think the truncation is not working on the label here as the `gux-truncate` styling is nested under the `.gux-description` class so that standard radio button variant cannot access the styling I think just move it from there and all should work.

321gillian:
> One last issue I should have picked this up in my first review 🙈, Screenshot 2024-05-28 at 10 33 36
> 
> I think the truncation is not working on the label here as the `gux-truncate` styling is nested under the `.gux-description` class so that standard radio button variant cannot access the styling I think just move it from there and all should work.

Gavin, you are the MVP for this. Thanks for spotting that!

gavin-everett-genesys (approved):

PR #471, opened 2024-05-21
chore(tooltip): Added tests for gux-tooltip-title component

jason-evans-genesys:
✅ Closes: COMUI-1408

Let me know if there are any other tests you think I should write for `gux-truncate` and `gux-tooltip-title`.

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-tooltip-title/tests/gux-tooltip-title.e2e.ts
+++ b/packages/genesys-spark-components/src/components/stable/gux-tooltip-title/tests/gux-tooltip-title.e2e.ts
@@ -1,0 +1,107 @@
+import { E2EPage, newE2EPage } from '@stencil/core/testing';
+import { a11yCheck } from '../../../../test/e2eTestUtils';
+
+async function newNonrandomE2EPage(
+  {
+    html
+  }: {
+    html: string;
+  },
+  lang: string = 'en'
+): Promise<E2EPage> {
+  const page = await newE2EPage();
+
+  await page.evaluateOnNewDocument(() => {
+    Math.random = () => 0.5;
+  });
+  await page.setContent(`<div lang=${lang}>${html}</div>`);
+  await page.waitForChanges();
+  await page.addScriptTag({
+    path: '../../node_modules/axe-core/axe.min.js'
+  });
+  await page.waitForChanges();
+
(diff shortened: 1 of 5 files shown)


jason-evans-genesys:
> Any reason you're going for just e2e tests alone here?

I was actually going to add a comment about this specifically today. I've been trying to work on writing spec tests but it seems like in order to write tests to show the tooltip you need to nest the gux-tooltip-text component within another one, and I think this.root within gux-tooltip-text has to be a gux-tooltip-text tag, and because of this it doesn't seem like I'm able to feasibly nest a gux-tooltip-text component within another one for a spec test. But correct me if I'm wrong. I'll ask the team in standup tomorrow.

packages/genesys-spark-components/src/components/stable/gux-tooltip-title/tests/gux-tooltip-title.e2e.ts
```diff
@@ -0,0 +1,43 @@
...
+  await page.waitForChanges();
+  await page.addScriptTag({
+    path: '../../node_modules/axe-core/axe.min.js'
+  });
+  await page.waitForChanges();
+
+  return page;
+}
+
+describe('gux-tooltip-title', () => {
```
jason-evans-genesys, line 27: I'm going to add some tests to check for tooltip visibility

jason-evans-genesys:
> Any reason you're going for just e2e tests alone here?

It looks like some of the tests to test for tooltip visibility wouldn't work in a spec file without breaking the gux-tooltip-text component at runtime. I saw that sometimes components in the spark repo don't have a spec file and rely solely on a e2e test file. I thought perhaps this could be one of those components where write all tests within an e2e test file made sense.

packages/genesys-spark-components/src/components/stable/gux-truncate/gux-truncate.tsx
```diff
@@ -86,7 +86,10 @@ export class GuxTruncate {
         }}
       >
         <span
-          class="gux-truncate-slot-container"
+          class={{
+            'gux-overflow-hidden': this.needsTruncation(),
```
jason-evans-genesys, line 90: I believe this is backwards compatible.

321gillian (approved):

PR #473, opened 2024-05-23
feat(tokens): splitted JSON

daragh-king-genesys:


packages/genesys-spark-tokens/data/layout & container.json
```diff
@@ -0,0 +1,255 @@
+{
```
gavin-everett-genesys, line 1: Is this a valid file name `layout & container.json` ?

daragh-king-genesys: valid yes. A good idea no :). I will get it changed in future PRs

PR #479, opened 2024-05-24
feat(dropdown): Options for selected option flexibility 

321gillian:
This is a discussion PR for a way to try and make the gux-option a little bit more flexible without impacting the text shown for the dropdown selected item. 

There were some challenges with the current approach. One was an issue the other day with the subtext for gux-option (which is undone here). The other also involved an attempt as adding supplementary info via tooltip. In both cases, because we currently just grab the gux-option textContent, the selected option in the dropdown didn't look correct. 

Any thoughts welcome. 

✅ Closes: COMUI-2736

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
@@ -20,6 +20,7 @@ import { onInputDisabledStateChange } from '@utils/dom/on-input-disabled-state-c
 import { afterNextRender } from '@utils/dom/after-next-render';
 import { trackComponent } from '@utils/tracking/usage';
 import { OnMutation } from '@utils/decorator/on-mutation';
+import { getSlotTextContent } from '@utils/dom/get-slot-text-content';
 
 import translationResources from './i18n/en.json';
 
@@ -421,13 +422,10 @@ export class GuxDropdown {
 
   private renderOption(option: HTMLGuxOptionElement): JSX.Element {
     let optionText = option.textContent;
-    if (hasSlot(option, 'subtext')) {
-      const subtext = option.querySelector('[slot=subtext]');
-      optionText = optionText.substring(
-        0,
-        optionText.length - subtext.textContent.length
-      );
+    if (hasSlot(option, 'text')) {
+      optionText = getSlotTextContent(option, 'text');
     }
+
     return (
       <gux-truncate ref={el => (this.truncateElement = el)}>
         {optionText}
--- a/packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
@@ -422,8 +422,15 @@ export class GuxDropdown {
 
   private renderOption(option: HTMLGuxOptionElement): JSX.Element {
     let optionText = option.textContent;
+    // TODO: COMUI-2909 Look at removing this subtext clause
     if (hasSlot(option, 'text')) {
       optionText = getSlotTextContent(option, 'text');
+    } else if (hasSlot(option, 'subtext')) {
+      const subtext = option.querySelector('[slot=subtext]');
+      optionText = optionText.substring(
+        0,
+        optionText.length - subtext.textContent.length
+      );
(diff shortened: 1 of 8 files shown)


gavin-everett-genesys:
I guess one idea would be to have a new `option-type` within the `option-types.ts` file currently we have `gux-option` and `gux-option-icon`. Maybe the creation of a `gux-option-tooltip` so that the tooltip will always display on hover and the truncation component would not be used you just do potentially `text-overflow:ellipsis`  which would prevent two tooltips from displaying in the scenario of long text.

I guess from the example provided they would have tooltips on all the `gux-options` but then there would be scenarios where the text will truncate and then you will be left with two tooltips.

I think using `subtext` would be the ideal option but they mentioned that there list would be increase in size over time and would double the size of the list.

321gillian:
> I guess one idea would be to have a new `option-type` within the `option-types.ts` file currently we have `gux-option` and `gux-option-icon`. Maybe the creation of a `gux-option-tooltip` so that the tooltip will always display on hover and the truncation component would not be used you just do potentially `text-overflow:ellipsis` which would prevent two tooltips from displaying in the scenario of long text.
> 
> I guess from the example provided they would have tooltips on all the `gux-options` but then there would be scenarios where the text will truncate and then you will be left with two tooltips.
> 
> I think using `subtext` would be the ideal option but they mentioned that their list would be increase in size over time and would double the size of the list.

I think the option with tooltip is not a pattern that we want to recommend. 

However, currently we determine the selected option text in a dropdown by getting option.textContent. This can result in supplementary information being concatenated with the option text. This limits the flexibility of the component. 

The proposal is just to allow the user to optionally identify the text that should be shown when option is selected.

packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
```diff
@@ -421,13 +422,10 @@ export class GuxDropdown {
 
   private renderOption(option: HTMLGuxOptionElement): JSX.Element {
     let optionText = option.textContent;
-    if (hasSlot(option, 'subtext')) {
-      const subtext = option.querySelector('[slot=subtext]');
-      optionText = optionText.substring(
-        0,
-        optionText.length - subtext.textContent.length
-      );
```
MattCheely, line 429: I think removing this will change the behavior for existing components with a `subtext` slot but no text slot, which could break deployed apps. So we probably need to preserve this in that specific case, unless I've missed something.

MattCheely (commented):
This approach looks good to me, but we need to make sure we don't impact existing implementations that predate the new `text` slot.

packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
321gillian: Added this back in, and put in a ticket for v5 to review

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option/gux-option.tsx
```diff
@@ -87,6 +87,7 @@ export class GuxOption {
       >
         <div class="gux-option-wrapper">
           <gux-truncate ref={el => (this.truncateElement = el)}>
+            <slot name="text" />
             <slot />
```
gavin-everett-genesys, line 91: Do both these slots serve the same purpose ? I know the default slot cannot be changed as it would be a breaking change ?

321gillian: yeah, I think they generally do but it keep the existing behaviour working as expected

katie-bobbe-genesys:
Would it be possible to review this ticket during collaboration time tomorrow? I'm not completely clear on what bug is being fixed by adding this `text` slot. If there is content in the text slot and the default slot, both are still being shown in the option:
image

321gillian:
> Would it be possible to review this ticket during collaboration time tomorrow? I'm not completely clear on what bug is being fixed by adding this `text` slot. If there is content in the text slot and the default slot, both are still being shown in the option: image

I don't think we'll need to. I conducted another few checks on this today and I think because of the truncation on the option, this isn't really a sensible change to make to the dropdown.

PR #487, opened 2024-06-03
Engage UI 8216 support container chart width

caitlanspronk:
Simple change to add width: 100% to the embed container for visualizations.


packages/genesys-spark-chart-components/src/components/beta/gux-visualization/gux-visualization.scss
```diff
@@ -17,3 +17,7 @@ text {
 .mark-rule.role-axis-domain line {
   stroke: #2e394c;
 }
+
+.chart-container {
```
daragh-king-genesys, line 21: I know this is annoying and not necessary (as this class is in the shadow dom) but we have a convention of prefixing all classes with `gux-`.

daragh-king-genesys (approved):

PR #489, opened 2024-06-04
Feature/missing cta group token single file

conor-darcy:
Adding to the old single file tokens.
Will create another PR for the multi-file tokens.


packages/genesys-spark-tokens/data-old/tokens.json
```diff
@@ -8232,6 +8232,12 @@
           }
         }
       }
+    },
+    "cta_group": {
```
jason-evans-genesys, line 8236: Just for my own understanding, what is is cta group exactly?

conor-darcy: @jason-evans-genesys "Call to action group". Basically a wrapper for buttons to provide consistent positioning.

PR #490, opened 2024-06-06
docs(tooltip-beta): fixing invalid html

conor-darcy:
A div cannot be nested inside a paragraph, replacing with span.


conor-darcy:
> Should we update the markup we use in the tests as well or does that matter?

Yeah, it'll do no harm @321gillian

321gillian (approved):

gavin-everett-genesys (approved):

gavin-everett-genesys:
Something I was looking at within this component which is unrelated to this issue is,

`    <span slot="content">
            <slot name="content">
              <slot />
            </slot>
          </span>
` . Is this valid ?

PR #491, opened 2024-06-06
chore(form-field-checkbox): Design changes for error message and icon

jason-evans-genesys:
-Error message icon change
-Error message icon and text container gap change
-Error message icon and text container alignment when text is multi-line - error icon now is positioned at the top of the container instead of the middle. To test this you can add multi-line text in the example page for a component like form-field-checkbox like in the screenshot below:
Screen Shot 2024-06-06 at 12 47 49 PM

✅ Closes: COMUI-2766


jason-evans-genesys:
@daragh-king-genesys @thomas-c-dillon  I noticed the snapshot tests are failing which I'll fix now.

daragh-king-genesys (approved):

PR #494, opened 2024-06-07
[WIP]feat(selector-cards): Checkbox variants added

321gillian:
✅ Closes: COMUI-2783


packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.scss
```diff
@@ -3,17 +3,29 @@
...
+$gux-icon-checkbox-unchecked: "data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' %3E%3Cpath d='M13.7143 0C14.975 0 16 1.02321 16 2.28571V13.7143C16 14.975 14.975 16 13.7143 16H2.28571C1.02321 16 0 14.975 0 13.7143V2.28571C0 1.02321 1.02321 0 2.28571 0H13.7143ZM13.7143 1.71429H2.28571C1.97 1.71429 1.71429 1.97 1.71429 2.28571V13.7143C1.71429 14.0286 1.97 14.2857 2.28571 14.2857H13.7143C14.0286 14.2857 14.2857 14.0286 14.2857 13.7143V2.28571C14.2857 1.97 14.0286 1.71429 13.7143 1.71429Z' /%3E%3C/svg%3E";
+$gux-icon-checkbox-checked: "data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' %3E%3Cpath d='M2.28571 0C1.025 0 0 1.025 0 2.28571V13.7143C0 14.975 1.025 16 2.28571 16H13.7143C14.975 16 16 14.975 16 13.7143V2.28571C16 1.025 14.975 0 13.7143 0H2.28571ZM12.0357 6.32143L7.46429 10.8929C7.12857 11.2286 6.58571 11.2286 6.25357 10.8929L3.96786 8.60714C3.63214 8.27143 3.63214 7.72857 3.96786 7.39643C4.30357 7.06429 4.84643 7.06071 5.17857 7.39643L6.85714 9.075L10.8214 5.10714C11.1571 4.77143 11.7 4.77143 12.0321 5.10714C12.3643 5.44286 12.3679 5.98571 12.0321 6.31786L12.0357 6.32143Z' /%3E%3C/svg%3E";
```
gavin-everett-genesys, line 7: I know these are already being used in `gux-form-field-checkbox.icons.scss` I was wondering is it possible to centralise them if they are the same svgs as the `form-field-checkbox` ones or maybe you could import that file here potentially ? Just incase we get another component that needs checkboxes and we have multiple files with the same svgs. Maybe some of the mixins from that file could also carry over if they are relatively the same.

gavin-everett-genesys:
Looks good to me 👍  I just had a few comments,

1. There is a small issue with the text in both the `radio` and `checkbox` types. There is extra whitespace around the label text which prevents you from selecting the deselecting the radio/checbox selector card if you are to click in that area. Screenshot 2024-06-10 at 10 03 51
2. Active styling does not seem to be applied to both checkbox and radio button should it be a dark shade of purple if active.
3. Will error states be included in a different PR. I seen one section in the design that had it but was not sure.
4. I think maybe logging a warning if a user provides an incorrect input other than the radio or checkbox types.

321gillian:
> Looks good to me 👍 I just had a few comments,
> 
> 1. There is a small issue with the text in both the `radio` and `checkbox` types. There is extra whitespace around the label text which prevents you from selecting the deselecting the radio/checbox selector card if you are to click in that area. Screenshot 2024-06-10 at 10 03 51
> 2. Active styling does not seem to be applied to both checkbox and radio button should it be a dark shade of purple if active.
> 3. Will error states be included in a different PR. I seen one section in the design that had it but was not sure.
> 4. I think maybe logging a warning if a user provides an incorrect input other than the radio or checkbox types.

Hi Gavin, 

Thanks for this feedback, it's super helpful. 
1. I made the gux-truncate fit-content so this should be ok now. 
2. Active styling was there but I think the way the styles were nested it wasn't showing up if you use the elements toggle for :active in Chrome. I've adjusted that so it should work now. 
3. Error states will be handled in a different ticket but I'll create a Jira to follow-up with design. 
4. The warning is already handled by getSlottedInput helper function

gavin-everett-genesys:
> > Looks good to me 👍 I just had a few comments,
> > 
> > 1. There is a small issue with the text in both the `radio` and `checkbox` types. There is extra whitespace around the label text which prevents you from selecting the deselecting the radio/checbox selector card if you are to click in that area. Screenshot 2024-06-10 at 10 03 51
> > 2. Active styling does not seem to be applied to both checkbox and radio button should it be a dark shade of purple if active.
> > 3. Will error states be included in a different PR. I seen one section in the design that had it but was not sure.
> > 4. I think maybe logging a warning if a user provides an incorrect input other than the radio or checkbox types.
> 
> Hi Gavin,
> 
> Thanks for this feedback, it's super helpful.
> 
> 1. I made the gux-truncate fit-content so this should be ok now.
> 2. Active styling was there but I think the way the styles were nested it wasn't showing up if you use the elements toggle for :active in Chrome. I've adjusted that so it should work now.
> 3. Error states will be handled in a different ticket but I'll create a Jira to follow-up with design.
> 4. The warning is already handled by getSlottedInput helper function

Thanks for this, I just did a quick re-review and noticed two things,

1. Will events be added to this component currently there are no events emitted ?
2. The truncation does not seem to work currently, I think its something related to the `maxLines` property. Probably has to be a lesser value.

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.scss
321gillian: I had a go at this but I feel like there's probably some ways to improve it.

321gillian:
> Thanks for this, I just did a quick re-review and noticed two things,
> 
> 1. Will events be added to this component currently there are no events emitted ?
> 2. The truncation does not seem to work currently, I think its something related to the `maxLines` property. Probably has to be a lesser value.

1. There's an input and change events on selector cards. I have added some markup to the html file to show that. Is that enough? 

2. I think the truncation should be ok in the updated version. Good catch

gavin-everett-genesys:
There is just one last small issue from re-reviewing. The tooltip does not seem to appear on truncation for the `description` slot.

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card-icons.scss
```diff
@@ -0,0 +1,98 @@
+@use '~genesys-spark/dist/scss/checkbox-icons.scss';
+@use '~genesys-spark/dist/scss/radio-icons.scss';
+
+@mixin radio-unchecked {
+  @include radio-icons.radio-unchecked;
+
+  background-color: var(
+    --gse-ui-selectorCard-default-selectedIndicator-unselected-foregroundColor
+  );
```
gavin-everett-genesys, line 9: Is `background-color` the only property that is causing this file to be needed ? Would it be better to delete this file and just put the `background-color` under the mixin in the `gux-selector-card.scss` file like ,

`::slotted(input[type='radio']:not(:checked))::before {
  background-color: var(
    --gse-ui-selectorCard-default-selectedIndicator-unselected-foregroundColor
  );
  @include gux-selector-card-icons.radio-unchecked;
}`

packages/genesys-spark-components/src/components/beta/gux-selector-cards/gux-selector-card/gux-selector-card.tsx
```diff
@@ -100,7 +100,6 @@ export class GuxSelectorCard {
           [`gux-${this.variant}`]: true,
           'gux-disabled': this.disabled
         }}
-        onClick={() => this.input.setAttribute('checked', 'true')}
```
321gillian, line 103: I removed this because I think it wasn't having an impact except when the description was clicked. I looked at applying a click handler to the description solely. However, this causes some issues. 

The checked attribute can't be relied on to set the state of the checkbox programmatically. It causes issues when the user interacts with the component. 

Triggering a click on the input when the description is clicked will set the checkbox to the correct state. However, the element will not receive focus which is the expected behaviour on Chrome and Firefox. While it's possible to just add some focus programmatically also, this doesn't match with the expected input behaviour on Safari. 

At this point, I think it's probably a case of either have the tooltip and clicking the description does nothing OR no tooltip and the description truncates but clicking on it selects the card. What do you think @gavin-everett-genesys @daragh-king-genesys ?

gavin-everett-genesys: Yeah I think the `onClick` should be removed . I could not find any differences between how the `label` is setup and how the `description` is setup and the label seems to be working fine. I think it must be possible to have that when you click the description it will focus and also have the ability to show the tooltip. It may be a `zindex` issue. I am not 100% sure.

gavin-everett-genesys:
Was there a reason why you removed the `truncation/tooltip` as now the truncation does not work on the simple selector card. And for the descriptive it does truncate but no tooltip is showing. Is this the same issue we were talking about the last day where it would not select the descriptive checkbox if you clicked on the description ? If so I can take a look to see if theres a way around it.

321gillian:
> Was there a reason why you removed the `truncation/tooltip` as now the truncation does not work on the simple selector card. And for the descriptive it does truncate but no tooltip is showing. Is this the same issue we were talking about the last day where it would not select the descriptive checkbox if you clicked on the description ? If so I can take a look to see if theres a way around it.

The issue with the truncation was that for keyboard only navigation there wasn't really a neat way to provide the truncated information to the user. I discussed this with the team at stand-up and we looked at the content guidelines in https://spark.genesys.com/latest/components/forms-and-inputs/selector-card/usage-MG5p2JxQ which specify not having the text any longer. At that point, it was agreed to leave out the tooltip since the label and description length are not supposed to run any longer.

gavin-everett-genesys:
I think there is just one styling issue should the `checkbox` and `radio` buttons color be a different colour. In the figma designs they are greyish colour.

Screenshot 2024-08-26 at 11 46 03

321gillian:
> I think there is just one styling issue should the `checkbox` and `radio` buttons color be a different colour. In the figma designs they are greyish colour.
> 
> Screenshot 2024-08-26 at 11 46 03

this is new to me. I'll follow-up with design

321gillian:
> I think there is just one styling issue should the `checkbox` and `radio` buttons color be a different colour. In the figma designs they are greyish colour.
> 
> Screenshot 2024-08-26 at 11 46 03

This was an error in Figma, fixed now

daragh-king-genesys (approved):

gavin-everett-genesys:
I think this issue which was recently raised may be resolved by this PR https://inindca.atlassian.net/browse/COMUI-3071

PR #496, opened 2024-06-10
Feature/gds 2472 tags color and size

conor-darcy:


packages/genesys-spark-components/src/components/stable/gux-tag/example.html
```diff
@@ -36,6 +36,42 @@ <h2>Disabled</h2>
...
+    <gux-tag emphasis="subtle" accent="10">Subtle 10</gux-tag>
+    <gux-tag emphasis="subtle" accent="invalid">Subtle Invalid</gux-tag>
+  </div>
+  <div>
+    <gux-tag emphasis="strong" accent="1">Strong 1</gux-tag>
+    <gux-tag emphasis="strong" accent="2">Strong 2</gux-tag>
+    <gux-tag emphasis="strong" accent="3">Strong 3</gux-tag>
+    <gux-tag emphasis="strong" accent="4">Strong 4</gux-tag>
+    <gux-tag emphasis="strong" accent="5">Strong 5</gux-tag>
+    <gux-tag emphasis="strong" accent="6">Strong 6</gux-tag>
```
gavin-everett-genesys, line 66: EDIT: I think this should be `bold` not `strong`

I know this probably wont match the figma design but I just dont know if there is a need for `bold` as `bold` is basically the default. So if you just provide a `gux-tag` component like,

`<gux-tag accent="1"> Accent 1 </gux-tag>` 

It will apply the `gux-emphasis-bold` class but it does not do anything. I was wondering if the if you could just have something like,

`@Prop()
 subtle: boolean = false;`

and then assign that class if it has been applied to the component without the need for strong ? Sort of like what has been done it badge.

`class={{gux-subtle: 'this.subtle'}}`


I guess it does make sense to have `bold` as the tokens are now called `bold` 🤔

packages/genesys-spark-components/src/components/stable/gux-tag/example.html
```diff
@@ -36,6 +36,42 @@ <h2>Disabled</h2>
...
+<h2>Size</h2>
+<div class="tags">
+  <gux-tag size="small">Small</gux-tag>
+  <gux-tag size="large">Large</gux-tag>
+</div>
+
+<h2>Emphasis</h2>
+<div class="tags">
+  <div>
+    <gux-tag emphasis="subtle" accent="1">Subtle 1</gux-tag>
```
daragh-king-genesys, line 48: Would `variant` be a more flexible property name?

packages/genesys-spark-components/src/components/stable/gux-tag/example.html
gavin-everett-genesys: Yeah now that I look at it again I think something like the `bold` property will probably lock us down to that specific variant in the case of more variants being added in the future it maybe be a bit of hassle.

conor-darcy: @gavin-everett-genesys you're right regarding 'strong', that was a typo on my part.
UX have confirmed that they think a 3rd 'minimal' option may be needed so we should avoid a boolean.

packages/genesys-spark-components/src/components/stable/gux-tag/example.html
conor-darcy: @daragh-king-genesys Variant doesn't sit well with UX, It has a particular meaning to designers in Figma & the feel 'emphasis' has more meaning in this case.

Badge has the opposite default value so we can't use a 'bold' prop however they feel they could need a 3rd option in the future & don't want to use a boolean.

packages/genesys-spark-components/src/components/stable/gux-tag/example.html
```diff
@@ -36,6 +36,42 @@ <h2>Disabled</h2>
   <gux-tag disabled accent="invalid">Invalid</gux-tag>
 </div>
 
+<h2>Size</h2>
+<div class="tags">
+  <gux-tag size="small">Small</gux-tag>
+  <gux-tag size="large">Large</gux-tag>
+</div>
+
+<h2>Emphasis</h2>
```
gavin-everett-genesys, line 45: We also have an `Active` section in the example file which are basically the same tags as the `bold` under the emphasis section. I am not sure how you will differentiate the two or maybe its fine.

conor-darcy: @gavin-everett-genesys yes, it's confusing. I've updated the docs to me more clear.

gavin-everett-genesys (approved):

PR #501, opened 2024-06-12
feat(tokens): Created themed styles

daragh-king-genesys:
Created themed styles

GDS-2540


321gillian:
One thing I noticed. Running `npm run new-tokens-data` generates a set of snapshots that aren't yet included here.

gavin-everett-genesys:
LGTM 👍 everything builds fine :)  I was getting these warnings about token collisions when running `npm run new-tokens-data` I am not sure if they are a cause for concern,
Screenshot 2024-06-14 at 10 53 08

PR #503, opened 2024-06-13
fix(calendar): addition of fixed height to calendar

gavin-everett-genesys:
addition of fixed height to calendar

COMUI-2915

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-calendar/gux-calendar.less
+++ b/packages/genesys-spark-components/src/components/stable/gux-calendar/gux-calendar.less
@@ -69,7 +69,7 @@
   .gux-content {
     display: flex;
     align-items: flex-start;
-    height: 190px;
+    height: 196px;
     padding: 20px 24px;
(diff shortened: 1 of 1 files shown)


packages/genesys-spark-components/src/components/stable/gux-calendar/gux-calendar.less
```diff
@@ -69,6 +69,7 @@
   .gux-content {
     display: flex;
     align-items: flex-start;
+    height: 190px;
```
gavin-everett-genesys, line 72: I think this height suits better for the `v3` version in `v4` the token is 234px but it does not look good in `v3` as the paddings and stuff are different.

321gillian: Screenshot 2024-06-14 at 10 39 27

I think the single calendar is getting cut a little too short here. 196px seems like it might work better there.

321gillian (approved):

PR #514, opened 2024-06-26
docs(icon): splitting icons into multiple pages

conor-darcy:
Too many icons on a single page takes too long to render (if they even render at all).
So I've split the icons into multiple pages.
Pages placed inside a 'docs' folder are generated & can be linked to.

Not perfect but it's an improvement.
Maybe other components could do with having sub-pages.
I will take another look at improving this in the near future, maybe an improved nav or adding them to the menu.

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-icon/docs/current.html
+++ b/packages/genesys-spark-components/src/components/stable/gux-icon/docs/current.html
@@ -2,7 +2,10 @@
 
 <nav>
   <ul>
-    <li><a href="gux-icon-current.html">Current icons</a></li>
+    <li><a href="gux-icon.html">Usage Guidelines</a></li>
+    <li>
+      <a href="gux-icon-current.html" aria-current="page">Current icons</a>
+    </li>
     <li><a href="gux-icon-spark.html">Spark icons</a></li>
     <li><a href="gux-icon-restricted.html">Restricted icons</a></li>
     <li><a href="gux-icon-legacy.html">Legacy icons</a></li>
--- a/packages/genesys-spark-components/src/components/stable/gux-icon/docs/current.html
+++ b/packages/genesys-spark-components/src/components/stable/gux-icon/docs/current.html
@@ -1,0 +1,50 @@
+<h1>gux-icon</h1>
+
+<nav>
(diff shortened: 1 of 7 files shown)


packages/genesys-spark-components/src/components/stable/gux-icon/docs/current.html
```diff
@@ -0,0 +1,50 @@
+<h1>gux-icon</h1>
+
+<nav>
+  <ul>
+    <li><a href="gux-icon-current.html">Current icons</a></li>
+    <li><a href="gux-icon-spark.html">Spark icons</a></li>
+    <li><a href="gux-icon-restricted.html">Restricted icons</a></li>
+    <li><a href="gux-icon-legacy.html">Legacy icons</a></li>
+    <li><a href="gux-icon-deprecated.html">Deprecated icons</a></li>
```
daragh-king-genesys, line 9: I would add a link for the root/example.html page
```
<li><a href="gux-icon.html">Usage Guidelines</a></li>
```

packages/genesys-spark-components/src/components/stable/gux-icon/docs/current.html
```diff
@@ -0,0 +1,50 @@
+<h1>gux-icon</h1>
+
+<nav>
+  <ul>
+    <li><a href="gux-icon-current.html">Current icons</a></li>
```
daragh-king-genesys, line 5: For completeness you should add a `aria-current="page"` attribute

gavin-everett-genesys:
LGTM 👍  Would be nice if we could have an option like `Load more...` or some sort of infinite scrolling but there is probably accessibility problems associated with it and may be harder to find icons if the user does `CTRL + F` to look for icons.

conor-darcy:
> LGTM 👍 Would be nice if we could have an option like `Load more...` or some sort of infinite scrolling but there is probably accessibility problems associated with it and may be harder to find icons if the user does `CTRL + F` to look for icons.

Yes Gavin, we can keep that in mind for the future.

gavin-everett-genesys (approved):

daragh-king-genesys (approved):

PR #518, opened 2024-06-27
chore(popover): Added role and arialDescribeby to popover and popover-list [WIP]

jason-evans-genesys:
This PR is on hold until this other PR is merged: https://github.com/MyPureCloud/genesys-spark/pull/627

Added role and arialDescribeby to popover and popover-list. You can use apple voiceover to confirm that when moving focus or opening a popover on the screen that the screenreader will say the popover element is a dialog and will read the content elements aloud. 

✅ Closes: COMUI-798

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
@@ -35,7 +35,7 @@ export class GuxPopoverList {
   private popupElement: HTMLDivElement;
   private arrowElement: HTMLDivElement;
   private cleanupUpdatePosition: ReturnType<typeof autoUpdate>;
-  private id: string = randomHTMLId('gux-tooltip');
+  private id: string = randomHTMLId('gux-popover-list');
 
   @Element()
   private root: HTMLElement;
--- a/packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
@@ -205,6 +205,8 @@ export class GuxPopoverList {
           'gux-hidden': !this.isOpen,
           'gux-popover-wrapper': true
         }}
+        role="dialog"
+        aria-describedby="gux-popover-content"
         data-placement
       >
         <div
--- a/packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
@@ -17,7 +17,6 @@ import {
   Placement,
   shift
 } from '@floating-ui/dom';
-import { randomHTMLId } from '@utils/dom/random-html-id';
 import { OnClickOutside } from '@utils/decorator/on-click-outside';
 import { trackComponent } from '@utils/tracking/usage';
 import { findElementById } from '@utils/dom/find-element-by-id';
@@ -35,7 +34,6 @@ export class GuxPopoverList {
   private popupElement: HTMLDivElement;
   private arrowElement: HTMLDivElement;
   private cleanupUpdatePosition: ReturnType<typeof autoUpdate>;
-  private id: string = randomHTMLId('gux-popover-list');
 
   @Element()
   private root: HTMLElement;
@@ -207,7 +205,6 @@ export class GuxPopoverList {
           'gux-popover-wrapper': true
         }}
         role="dialog"
-        aria-describedby={this.id}
         data-placement
       >
         <div
@@ -219,7 +216,7 @@ export class GuxPopoverList {
             onClick={this.dismiss.bind(this)}
           ></gux-dismiss-button>
         )}
-        <div id={this.id} class="gux-popover-content">
+        <div class="gux-popover-content">
           <slot />
         </div>
       </div>
--- a/packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
@@ -17,7 +17,7 @@ import {
   Placement,
   shift
 } from '@floating-ui/dom';
-
+import { randomHTMLId } from '@utils/dom/random-html-id';
 import { OnClickOutside } from '@utils/decorator/on-click-outside';
 import { trackComponent } from '@utils/tracking/usage';
 import { findElementById } from '@utils/dom/find-element-by-id';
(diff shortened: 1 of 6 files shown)


packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
```diff
@@ -205,6 +205,8 @@ export class GuxPopoverList {
           'gux-hidden': !this.isOpen,
           'gux-popover-wrapper': true
         }}
+        role="dialog"
+        aria-describedby="gux-popover-content"
```
daragh-king-genesys, line 209: I am not sure about the dialog role for a popover. Did you have a reference for using this role for a popover?

daragh-king-genesys, line 209: `aria-describedby` takes a list of ids not a class. You will have to give the element an auto generate id and then reference it here.

jason-evans-genesys: It's a good point. I see in our spark repo that we use gux-popover more like a dialog since the user clicking on an element triggers the popover to show. However, do clients using spark use the popover more like a tooltip where it shows on hover?

jason-evans-genesys: Fixed

katie-bobbe-genesys: It's important to keep in mind that this ticket was created `December 15, 2021 at 3:49 PM`. A lot has changed since then and what's written in these older tickets and comments is not necessarily always still relevant. 

The three issues documented in the ticket were:
_1. Content in popover is only accessible on mouse hover_

The example page should probably be updated to show the popover showing on button press instead of hover. The [Spark Popover documentation](https://spark.genesys.com/latest/components/feedback/popovers/usage-wA2Lqhzx) now states that 'popovers are triggered on click and persistent'. We should update the example page to highlight this use case.

_2. Consider using `role="tooltip"` for popover_

This ticket was written before the `dialog` role was available to use in all browsers. I don't believe the `tooltip` role is usable here, since the design system now suggests that the popover should be triggered "on click".  Tooltips are supposed to be triggered on hover, generally. Since `role=tooltip` is probably not appropriate, we need to do more research to confirm whether or not `dialog` is an appropriate role for this component. 

_3. Consider trying `aria-describedby` to link popover target and popover target_ 

This is trying to say that the **target** (button or something that opens the popover) should have the aria-describedby attribute linking to the popover content (id on the content)

katie-bobbe-genesys (changes requested):
see comment above

packages/genesys-spark-components/src/components/stable/gux-popover-list/gux-popover-list.tsx
jason-evans-genesys: @katie-bobbe-genesys Correct me if I'm wrong, but will the client need to remember to add `aria-describeby` to the target element that opens the popover? I could add it to the target element in the popover's example page but wanted to clarify this.

packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
```diff
@@ -251,6 +251,7 @@ export class GuxPopover {
           'gux-hidden': !this.isOpen,
           'gux-popover-wrapper': true
         }}
+        role="dialog"
```
daragh-king-genesys, line 254: I don't think a popover is a dialog. Have a read of this and see if you agree or not https://bootcamp.uxdesign.cc/popups-dialogs-tooltips-and-popovers-ux-patterns-2-939da7a1ddcd

packages/genesys-spark-components/src/components/stable/gux-popover/example.html
```diff
@@ -1,17 +1,26 @@
 <h1>gux-popover</h1>
+
 <div class="scroll-container">
   <div class="scroll-interior">
-    <div id="popover-target">Example Element</div>
+    <gux-button
+      id="popover-target"
+      accent="primary"
+      aria-describedby="popover-content-1"
```
daragh-king-genesys, line 8: This does not make sense to me. Why would an element be described by its popover content?

packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
jason-evans-genesys: @daragh-king-genesys  I agree with you. After researching more I don't think dialog is the right choice. It seems like the dialog role would be a good choice if we had a smaller window that was blocking in some sort to direct user to that dialog and make it the main focus of the page perhaps. popover-list seems to be a little more passive than a dialog and not as blocking. I'm trying to figure out what would be a better choice than dialog. 

Some potential role options I found are:

1) tooltip - I'm not sure about this one. So at first I thought tooltips were good for additional content that would show on hover, but from researching it looks like you can have tooltips that show on click too. I don't think tooltip makes sense for us because we're always going to show a list of some sort, and not just additional text, which is what I'd think the tooltip role would be good for.
1) list - I think this is more of a general list (e.g. `ul`) that has plain text in each item
1) listbox - this seems to be used for elements like a `select` where you have different options to choose from. Could be used potentially but we won't always necessarily have selectable list items and could have just plain text.
1) menu - this seems to be for actionable list items (e.g. clicking a button to trigger an action). I don't think this one makes sense since you can have actionable or non-actionable items in popover-list.

Since popover-list can have any sort of list elements then I'm thinking maybe a more generic `list` role might be good. What do you think?

packages/genesys-spark-components/src/components/stable/gux-popover/example.html
jason-evans-genesys: I read this as the popover content is being described by the button element. So if the user focuses on the button then the screenreader could read out additionally linked information about the popover.

Here's an example online I found for using `aria-describedby` with a button and tooltip text container that has some similarities with how I implemented `aria-describedby`:

```
   <button type="button" aria-describedby="my-tooltip">Snipping Tool</button>
    <div hidden id="my-tooltip" role="tooltip">
        It can take still screenshots of an open window,
        rectangular areas, a free-form area,
        or the entire screen.
    </div>
```

What do you think?

katie-bobbe-genesys:
I was looking into this again today and I would recommend considering a new perspective on this older ticket. I haven't had a chance to test it out yet, but it looks like there is a newly available [popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) and [popover global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover) that we might be able to leverage here. I would recommend doing some research into this new Popover API and see if this will fit our use case for popover accessibility.

PR #520, opened 2024-06-28
fix(popover): Fixed accessibility focus issues with gux-popover (FEEDBACK)

jason-evans-genesys:
Fixed some accessibility issues with the popover component COMUI-824 has more info about the 3 accessibility issues fixed in this PR)

There is now logic to focus on the popover content container element after the user opens the popover via clicking a target element. I tell the screenreader about this focused content container element with an aria-label attribute.

✅ Closes: COMUI-824

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-popover/example.html
+++ b/packages/genesys-spark-components/src/components/stable/gux-popover/example.html
@@ -17,9 +17,9 @@
       <div class="popover-content-wrapper">
         <div>Popover Content</div>
         <div class="popover-content">
-          <gux-button accent="ghost" onclick="notify(event)">
-            Action 3
-          </gux-button>
+          <gux-button-slot accent="ghost">
+            <button type="button" autofocus>Action 3</button>
+          </gux-button-slot>
           <gux-button accent="secondary" onclick="notify(event)">
             Action 2
           </gux-button>
--- a/packages/genesys-spark-components/src/components/stable/gux-popover/example.html
+++ b/packages/genesys-spark-components/src/components/stable/gux-popover/example.html
@@ -2,6 +2,11 @@
 <div class="scroll-container">
   <div class="scroll-interior">
     <div id="popover-target">Example Element</div>
+
+    <div class="open-popover">
+      <button id="open-popover">Open Popover</button>
+    </div>
+
     <gux-popover
       id="popover-example"
       position="top"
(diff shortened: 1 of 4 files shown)


packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
```diff
@@ -215,17 +218,43 @@ export class GuxPopover {
...
+    this.contentElement.focus();
+
+    // TODO: should we dynamically focus first focusable elemtn in the slotted content?
+    //
+    // // Focus the first focusable element in the slotted content
+    // const contentElementChildren = this.getRootChildren(this.contentElement);
+    // if (contentElementChildren?.length) {
+    //   const firstChild = this.getRootChildren(this.contentElement)[0] as HTMLElement;
+    //   const firstSlottedInput = firstChild.querySelector(
+    //     'gux-button gux-dropdown'
```
jason-evans-genesys, line 246: Problem with the code I have here if we want to dynamically focus the first focusable element is that what I have on line 246 is not scalable since we'd have to maintain a whitelist of focusable gux components. Maybe instead we could search for a raw input or button tag if we go with this approach.

packages/genesys-spark-components/src/components/stable/gux-popover/example.html
```diff
@@ -12,9 +17,9 @@ <h1>gux-popover</h1>
       <div class="popover-content-wrapper">
         <div>Popover Content</div>
         <div class="popover-content">
-          <gux-button accent="ghost" onclick="notify(event)">
-            Action 3
-          </gux-button>
+          <gux-button-slot accent="ghost">
+            <button type="button" autofocus>Action 3</button>
```
jason-evans-genesys, line 21: Trying to use an autofocus attribute like the gux-modal does to focus an element on popover load. This doesn't seem to be working yet so I'll keep messing with it to get this element to autofocus on popover load.

packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
```diff
@@ -215,17 +218,43 @@ export class GuxPopover {
...
       this.runUpdatePosition();
     } else if (this.cleanupUpdatePosition) {
       this.cleanupUpdatePosition();
     }
   }
 
+  private focusFirstSlottedElement(): void {
+    this.contentElement.focus();
+
+    // TODO: should we dynamically focus first focusable elemtn in the slotted content?
```
jason-evans-genesys, line 239: I'm going to delete this code.

katie-bobbe-genesys (changes requested):
The original ticket (which is several years old at this point), https://inindca.atlassian.net/browse/COMUI-824 was intended for the `gux-popup` component not the `gux-popover` component. The `gux-popup` component used to be included in the examples page, but is not any longer because we are only supporting it as a component internal to other components. 

I don't think that these changes in this PR are beneficial for the `gux-popover` component and I think the original ticket can be closed. `gux-popup` is no longer supported as a standalone component, and the accessibility issues such as focus management should be handled by the components consuming it internally. 

If you would like, we can go over this PR tomorrow in collab time and I can explain why some of these additions in this PR are actually counter productive and decreases accessibility of this component.

PR #524, opened 2024-07-01
chore(other): Added playground scripts in mono repo

jason-evans-genesys:
✅ Closes: COMUI-1721


package.json
```diff
@@ -28,6 +28,8 @@
     "lint-staged.root": "lint-staged",
     "lint-staged.workspaces": "npm run lint-staged --workspaces --if-present",
     "new-tokens-data": "npm run snapshot --workspace=packages/genesys-spark-tokens",
+    "playground.angular": "cd ./playground/angular-v13 && rm ./package-lock.json && npm install && npm start",
```
jason-evans-genesys, line 31: I'm going to try using a playground workspace instead of changing directory in this script

package.json
```diff
@@ -50,7 +54,6 @@
     "./packages/genesys-spark-components",
     "./packages/genesys-spark-chart-components",
     "./packages/*",
-    "./playgrounds/*",
```
jason-evans-genesys, line 53: I'm trying to get the playground workspace to work instead of using the `cd` command to change directory into a playground folder.

jason-evans-genesys: Our playground project is named `playground` and not `playgrounds` so when I changed the workspace name to be `playground` then running `npm install` on the root of the spark repo was installing angular dependencies into the root's `node_modules` folder which according to Node workspace docs is expected but it was causing some angular reference errors in the root directory. What do you all think? Does it make sense to maybe not have the playgrounds as a workspace?

katie-bobbe-genesys: I'm looking into this now, I'll follow up once I find an answer to this

jason-evans-genesys: I was able to figure this issue out and am using workspaces for the playgrounds now.

jason-evans-genesys:
@katie-bobbe-genesys Can you clarify `Also add some scripts for running a playground and the docs site simultaneously` in the description of this ticket? I added scripts to run each playground framework in the root package.json and wanted to confirm this is what we want or if I’m missing something. Thank you.

package.json
```diff
@@ -28,6 +28,10 @@
     "lint-staged.root": "lint-staged",
     "lint-staged.workspaces": "npm run lint-staged --workspaces --if-present",
     "new-tokens-data": "npm run snapshot --workspace=packages/genesys-spark-tokens",
+    "playground.angular": "rm -f ./package-lock.json && npm install && npm start --workspace=playground/angular-v13",
```
katie-bobbe-genesys, line 31: I don't think we want to remove the `package-lock.json` each time this is run

katie-bobbe-genesys:
The script that runs the playground should also start the local genesys-webcomponents dev server so that the local version of the webcomponents can be tested with the playground. Some of the playgrounds may be currently hard coded to reference certain deployed version numbers, so we may need to change the playgrounds to all reference the local version.

Another nice to have would be to have a script or an easy way to switch between testing the local version of genesys webcomponents and a certain deployed version number.  This can just make it easier when troubleshooting a bug that is reported on a certain version number.

I'm not exactly sure the best way to implement this at the moment, but I'm thinking that there should be a way to start the playground testing with the local version, and a way to start the playground testing against a deployed version.

Let me know if you would like more details or clarification on this

jason-evans-genesys:
> The script that runs the playground should also start the local genesys-webcomponents dev server so that the local version of the webcomponents can be tested with the playground. Some of the playgrounds may be currently hard coded to reference certain deployed version numbers, so we may need to change the playgrounds to all reference the local version.
> 
> Another nice to have would be to have a script or an easy way to switch between testing the local version of genesys webcomponents and a certain deployed version number. This can just make it easier when troubleshooting a bug that is reported on a certain version number.
> 
> I'm not exactly sure the best way to implement this at the moment, but I'm thinking that there should be a way to start the playground testing with the local version, and a way to start the playground testing against a deployed version.
> 
> Let me know if you would like more details or clarification on this

@katie-bobbe-genesys I pushed some changes that will run the playgrounds concurrently with the spark process. Let me know if this solution is good for us. If so, thoughts on how we should organize these individual playground scripts i made or do you think they're fine within the root dir like it is now? I could add a playground-scripts folder and organize them within there.

playground.angular.json
```diff
@@ -0,0 +1,39 @@
...
+    "command": ["npm", "run", "dev.examples"],
+    "title": "Docs - http://localhost:8080",
+    "defaultStatus": ["🗃️", "🗃️"],
+    "status": {
+      "ERROR": ["🚨", "🚨"],
+      "compiled successfully": ["🗃️", "🗃️"]
+    }
+  },
+  {
+    "command": ["npm", "run", "playground.angular.start"],
```
jason-evans-genesys, line 35: We could just run all playgrounds instead of each individual one in a separate script like I do here. Thoughts?

katie-bobbe-genesys: I think running one playground at a time would probably work best.

playground.angular.json
```diff
@@ -0,0 +1,39 @@
...
+    "title": "Angular Playground",
```
katie-bobbe-genesys, line 36: For each playground, for this title I would add the port that links to the playground. For example:
`"title": "Angular Playground - http://localhost:4200"

katie-bobbe-genesys:
So far this implementation looks pretty good to me. I would make sure that each of the playgrounds are correctly configured to work with the locally running docs. For example, the `index.html` in Angular 13 need to be updated to point to the correct CSS and Javascript links. The links are outdated from before some of the updates we did in the monorepo, so they need to be updated. Let me know if you have any questions about this and I can dig a bit deeper to try to find more info

playground/angular-v13/src/index.html
```diff
@@ -7,13 +7,13 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="icon" type="image/x-icon" href="favicon.ico" />
     <link
-      href="http://localhost:8080/genesys-webcomponents/genesys-webcomponents.css"
+      href="http://localhost:3733/dist/genesys-webcomponents/genesys-webcomponents.css"
```
jason-evans-genesys, line 10: @katie-bobbe-genesys I made this path change in each playground to get the local web components working with the playgrounds.

daragh-king-genesys (approved):

katie-bobbe-genesys:
This is looking pretty good, I pulled down this branch and tested the scripts. The only thing missing is that I think each of the playgrounds needs an `.npmrc` file that points to the public npm registry. There is already one in the `vue-3` folder.

katie-bobbe-genesys (approved):

PR #527, opened 2024-07-10
style(copy-to-clipboard): Constrain width of copy-to-clipboard component 

harter-genesys:
Set max-width to 100% for copy-to-clipboard component.  Set min-width of icon in copy-to-clipboard-component.

✅ Closes: COMUI-2961
Before:
before-copy-to-clipboard-with-truncate

After:
copy-to-clipboard-with-truncate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### Bug Fixes
+
+* **copy-to-clipboard:** contrain max-width of component and min-width of copy icon ([202c3d9](https://github.com/harter-genesys/genesys-spark/commit/202c3d931fdad696968b72296ca28105b11f9858))
+
+### [4.64.1](https://github.com/MyPureCloud/genesys-spark/compare/v4.63.1...v4.63.2) (2024-07-02)
+
 ## [4.64.0](https://github.com/MyPureCloud/genesys-spark/compare/v4.63.3...v4.64.0) (2024-07-10)
 
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "genesys-spark-components",
-  "version": "4.64.0",
+  "version": "4.64.1",
   "description": "Common webcomponents",
   "license": "MIT",
   "scripts": {
--- a/packages/genesys-spark-components/src/components/stable/gux-copy-to-clipboard/gux-copy-to-clipboard.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-copy-to-clipboard/gux-copy-to-clipboard.scss
@@ -17,11 +17,13 @@ button {
 .gux-copy-to-clipboard-wrapper {
   display: inline-flex;
   align-items: center;
+  max-width: 100%;
 
   .gux-copy-content {
     display: inherit;
     gap: var(--gse-ui-copyToClipboard-contentContainer-gap);
     align-items: inherit;
+    max-width: 100%;
     cursor: pointer;
 
     ::slotted(*) {
@@ -42,6 +44,7 @@ button {
     gux-icon[icon-name='copy'] {
       flex-shrink: 0;
       width: var(--gse-ui-icon-size-sm);
+      min-width: 22px;
       height: var(--gse-ui-icon-size-sm);
       padding: var(--gse-ui-copyToClipboard-iconContainer-padding);
       visibility: hidden;


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### Bug Fixes
+
+* **copy-to-clipboard:** contrain max-width of component and min-width of copy icon ([202c3d9](https://github.com/harter-genesys/genesys-spark/commit/202c3d931fdad696968b72296ca28105b11f9858))
```
harter-genesys, line 7: I imagine this commit needs update, but it's a bit of a chicken an egg problem since it's coming from a fork.

packages/genesys-spark-components/src/components/stable/gux-copy-to-clipboard/gux-copy-to-clipboard.scss
```diff
@@ -42,6 +44,7 @@ button {
     gux-icon[icon-name='copy'] {
       flex-shrink: 0;
       width: var(--gse-ui-icon-size-sm);
+      min-width: 22px;
```
harter-genesys, line 47: 22px was a size that I found in playing around with the icon.  Otherwise, it is fairly arbitrary. 

This is important because I've found that the icon shrinks far too small in the event the content is too big for the space its in, and the container wraps the text.

harter-genesys:
The merge conflict has to do with my question about which commit to include in the change log.  I'll wait for direction on how to fill out the changelog until addressing the merge conflict.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "genesys-spark-components",
-  "version": "4.64.0",
+  "version": "4.64.1",
```
321gillian, line 3: Again, no need to update this. Once a change it merged, the daily release auto-increments the version number

321gillian:
Hi Matt, 

Can you confirm where you saw the issue just so I can validate the change on my side? 

Thanks,
Gillian

harter-genesys:
> Hi Matt,
> 
> Can you confirm where you saw the issue just so I can validate the change on my side?
> 
> Thanks, Gillian

I saw this locally when using the copy-to-clipboard component in a table (not a gux-table if that matters).  I have html to reproduce the issue in the ticket.

packages/genesys-spark-components/src/components/stable/gux-copy-to-clipboard/gux-copy-to-clipboard.scss
harter-genesys: It keeps the icon from shrinking too small.  I've had the copy icon shrink to the size of a dot if the text for copy-to-clipboard is too big, and the container it is in wraps text. 

`min-width` just defines a minimum size for the icon so that does not happen.  Looking through the changelog, I saw this was an issue that was attempted to be addressed in a 4.0.0 beta. 
https://github.com/MyPureCloud/genesys-spark/commit/2b65c9278268aca2163c8cc10a969e8ef0658660

321gillian: Hi @harter-genesys 

Can you send on some more info about how to recreate the shrinking icon issue? I've tried altering the font-size and content of the main section of the copy-to-clipboard but I couldn't replicate it. 

Thanks,
Gillian

harter-genesys: I'm having troubles directly replicating without using the `fixtable` package our app is using for tables.  It appears the table-layout is what is causing the shrinking to occur (that in conjunction with something else).  I'm having a hard time reproducing this in the webcomponent playground.  If I'm unable to get an example, I am fine to remove this bit.
With the `table-layout` set to `fixed`:
image
With `table-layout` unset:
image

Here are the styles for the top level table element, but like I said, using those styles in the playground don't appear to be replicating the issue:
```html
<table class="table table-hover" style="margin: 0px;padding: 0px;border-width: 0px;border-style: initial;border-color: initial;table-layout: fixed;">
```

harter-genesys: @321gillian I figured it out.  For some reason, the styles for `gux-icon` are not being applied in my usage.  That `flex-shrink: 0` works to resolve the issue pasted in the screenshot.  I'm investigating why these styles are not present

harter-genesys: it looks like it was a bug fixed a couple weeks ago in 4.64.4 (right after I opened this ticket).  The selector in scss was incorrect and the styles were not being applied.  
https://github.com/MyPureCloud/genesys-spark/commit/741f6792b498d6e907b3d47beb903f1f64fdf7e4

321gillian (approved):

katie-bobbe-genesys (approved):

PR #528, opened 2024-07-10
feature(tokens): Added theming to UI tokens pt1

VirtualClover:
This is the first batch of UI tokens updates, so they can support theming


jordanstith15 (approved):

VirtualClover:
> Screenshot 2024-07-15 at 09 59 48 When I try to run this locally I get the following error from StyleDictionary

https://github.com/MyPureCloud/genesys-spark/blob/d8c7ba91f76a7f1065de5fc5a1038c25fa85d4f2/packages/genesys-spark-tokens/data/semantic/global.json#L432


Seems to be there, let me take a closer look 👀

VirtualClover:
> > Screenshot 2024-07-15 at 09 59 48 When I try to run this locally I get the following error from StyleDictionary
> 
> https://github.com/MyPureCloud/genesys-spark/blob/d8c7ba91f76a7f1065de5fc5a1038c25fa85d4f2/packages/genesys-spark-tokens/data/semantic/global.json#L432
> 
> Seems to be there, let me take a closer look 👀

Ok, so it seems the issue is caused because there's also a `focus` token on the UI/forms_and_inputs set also, I changed the one we added to focus_outline to fix our issue.

With that fix it looks that flare-light is built with no issues

Screenshot 2024-07-16 at 1 18 46 p m


But then it throws a whole lot of other errors I'm assuming is trying to build dark mode but since we don't have dark mode tokens then this happens. Is that correct @daragh-king-genesys ?

VirtualClover:
> > > Screenshot 2024-07-15 at 09 59 48 When I try to run this locally I get the following error from StyleDictionary
> > 
> > 
> > https://github.com/MyPureCloud/genesys-spark/blob/d8c7ba91f76a7f1065de5fc5a1038c25fa85d4f2/packages/genesys-spark-tokens/data/semantic/global.json#L432
> > 
> > Seems to be there, let me take a closer look 👀
> 
> Ok, so it seems the issue is caused because there's also a `focus` token on the UI/forms_and_inputs set also, I changed the one we added to focus_outline to fix our issue.
> 
> With that fix it looks that flare-light is built with no issues
> 
> Screenshot 2024-07-16 at 1 18 46 p m
> But then it throws a whole lot of other errors I'm assuming is trying to build dark mode but since we don't have dark mode tokens then this happens. Is that correct @daragh-king-genesys ? Screenshot 2024-07-16 at 1 18 40 p m

@321gillian @daragh-king-genesys Yep, that was the issue, references were fixed and it should build now :)

cc. @JackieYada

daragh-king-genesys (approved):

PR #531, opened 2024-07-15
fix(copy-to-clipboard): fix status icon styles and align icon and text

gavin-everett-genesys:
fix status icon styles and align icon and text

[✅ Closes: COMUI-2965](https://inindca.atlassian.net/browse/COMUI-2965)

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-copy-to-clipboard/gux-copy-to-clipboard.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-copy-to-clipboard/gux-copy-to-clipboard.scss
@@ -5,7 +5,7 @@
     );
   }
 
-  gux-icon[icon-name='copy'] {
+  gux-icon[icon-name='fa/copy-regular'] {
     visibility: visible;
   }
 }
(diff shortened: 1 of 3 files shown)


packages/genesys-spark-components/src/components/stable/gux-copy-to-clipboard/gux-copy-to-clipboard.scss
```diff
@@ -5,7 +5,7 @@
     );
   }
 
-  gux-icon[icon-name='copy'] {
+  gux-icon[icon-name='fa/copy-regular'] {
```
gavin-everett-genesys, line 8: There seems to be some duplicate icons, for example `copy` and `fa-copy-regular` are the same visually.

jordanstith15: Are we switching to Font Awesome icons then? I thought we were prioritizing ours. Might be something to ask UX

gavin-everett-genesys: Confirmed in standup to use `fa`

jordanstith15 (approved):

PR #532, opened 2024-07-15
feat(side sheet): side sheet component

conor-darcy:
API feedback wanted please for the non-modal version:
https://www.figma.com/design/JKbHmcf4nUF6C7Pj8M6MpY/Spark%3A-Flare-4.0---Core?node-id=28482-782&m=dev


packages/genesys-spark-components/src/components/beta/gux-side-sheet/example.html
```diff
@@ -0,0 +1,73 @@
+<div class="side-sheet-content">
+  <h1>Side Sheet</h1>
+</div>
+
+<gux-side-sheet-beta size="medium">
+  <gux-icon slot="icon" icon-name="fa/diamond-regular" size="medium"></gux-icon>
+  <gux-side-sheet-heading-beta slot="heading" level="3">
+    Side sheet title
+  </gux-side-sheet-heading-beta>
```
gavin-everett-genesys, line 9: I dont think this component needs the `beta` tag tbh since its just styling the heading ? Also I am sort of confused by this component. I cannot find anywhere in the designs where it specifies the use of heading tags.

Could you not just have a component that has a default slot and the user can slot whatever they like eg a heading tag and you style whatever element is passed with the same styling like,


```
::slotted(*) {
  padding: 0;
  margin: 0;
  font-family: var(--gse-ui-dynamicSideSheet-heading-fontFamily);
  font-size: var(--gse-ui-dynamicSideSheet-heading-fontSize);
  font-weight: var(--gse-ui-dynamicSideSheet-heading-fontWeight);
  line-height: var(--gse-ui-dynamicSideSheet-heading-lineHeight);
  color: var(--gse-ui-dynamicSideSheet-headerColor);
}
```

so regardless of what the consumer slots it will always get the same styling so the api might look like,
`<gux-side-sheet-heading slot="heading"><h1>This is a heading</h1> </gux-side-sheet-heading>`

There may be an issue if a user was to slot just a text node as `::slotted(*)` cannot target that but you could just create a `light.scss` file and do it there like,

```
gux-side-sheet-beta {

gux-side-sheet-heading {
   ... same styles as from the component

}
}
```

I may have missed something from the design so let me know.

packages/genesys-spark-components/src/components/beta/gux-side-sheet/example.html
```diff
@@ -0,0 +1,73 @@
+<div class="side-sheet-content">
+  <h1>Side Sheet</h1>
+</div>
+
+<gux-side-sheet-beta size="medium">
```
gavin-everett-genesys, line 5: I think you should include examples of `small` and `large` also.

packages/genesys-spark-components/src/components/beta/gux-side-sheet/gux-side-sheet.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  render(): JSX.Element {
+    return (
+      <div class={`gux-side-sheet gux-side-sheet-${this.size}`}>
+        <header>
+          <div class="gux-side-sheet-icon">
+            <slot name="icon" />
+          </div>
+          <slot name="heading" />
+          <gux-dismiss-button />
```
daragh-king-genesys, line 29: `gux-dismiss-button` does not have any functionality by default you will have to add a click handler and dismiss functionality

packages/genesys-spark-components/src/components/beta/gux-side-sheet/gux-side-sheet.tsx
```diff
@@ -0,0 +1,40 @@
...
+})
+export class GuxSideSheet {
+  @Element()
+  private root: HTMLElement;
+
+  @Prop()
+  size: GuxSideSheetSize = 'small';
+
+  componentWillLoad(): void {
+    trackComponent(this.root);
```
daragh-king-genesys, line 18: we usually track stuff like the size as a varient.

packages/genesys-spark-components/src/components/beta/gux-side-sheet/gux-side-sheet.tsx
```diff
@@ -0,0 +1,40 @@
...
+  tag: 'gux-side-sheet-beta',
+  styleUrl: 'gux-side-sheet.scss',
+  shadow: true
+})
+export class GuxSideSheet {
+  @Element()
+  private root: HTMLElement;
+
+  @Prop()
+  size: GuxSideSheetSize = 'small';
```
daragh-king-genesys, line 15: If this is just the non modal version I think the sizing is wrong, there should only be small and medium

daragh-king-genesys (commented):
API wise my biggest question is how much control we want of this component. At the moment it is mainly just a presentational component and most of the layout stuff is left up to the app using the component.
I think we will need to see an implementation of the modal version too before we nail down an api.

packages/genesys-spark-components/src/components/beta/gux-side-sheet/gux-side-sheet.tsx
gavin-everett-genesys, line 29: I dont think the `gux-dismiss-button` is currently positioned correctly. I think it should be positioned `absolute` with a `top` and `right` of `12px`

packages/genesys-spark-components/src/components/beta/gux-side-sheet/gux-side-sheet.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <div class={`gux-side-sheet gux-side-sheet-${this.size}`}>
+        <header>
+          <div class="gux-side-sheet-icon">
+            <slot name="icon" />
+          </div>
+          <slot name="heading" />
+          <gux-dismiss-button />
+        </header>
+        <div class="gux-side-sheet-content">
+          <slot name="content" />
```
gavin-everett-genesys, line 32: Its not really outlined in the design so probably a UX question but there is an area above the `content` slot for  side-sheet description of some sort. Wondering if there is a need for a `description` slot.

packages/genesys-spark-components/src/components/beta/gux-side-sheet/gux-side-sheet.scss
```diff
@@ -0,0 +1,58 @@
...
+  & > header {
+    display: flex;
+    flex: 0 0 auto;
+    gap: var(--gse-ui-dynamicSideSheet-header-interiorGap);
+    align-items: center;
+    padding: var(--gse-ui-dynamicSideSheet-header-paddingTop)
+      var(--gse-ui-dynamicSideSheet-header-paddingRight)
+      var(--gse-ui-dynamicSideSheet-header-gap)
+      var(--gse-ui-dynamicSideSheet-header-paddingLeft);
+    border-bottom: 1px solid var(--gse-ui-modalSideSheet-dividerColor);
```
gavin-everett-genesys, line 35: Token needed here.

packages/genesys-spark-components/src/components/beta/gux-side-sheet/gux-side-sheet.scss
```diff
@@ -0,0 +1,58 @@
...
+    font-family: var(--gse-ui-dynamicSideSheet-description-fontFamily);
+    font-size: var(--gse-ui-dynamicSideSheet-description-fontSize);
+    font-weight: var(--gse-ui-dynamicSideSheet-description-fontWeight);
+    line-height: var(--gse-ui-dynamicSideSheet-description-lineHeight);
+  }
+
+  & > footer {
+    flex: 0 0 auto;
+    padding: var(--gse-ui-dynamicSideSheet-footer-padding);
+    border-top: 1px solid var(--gse-ui-modalSideSheet-dividerColor);
```
gavin-everett-genesys, line 56: Token needed here

packages/genesys-spark-components/src/components/beta/gux-side-sheet/example.html
conor-darcy: @gavin-everett-genesys
**"I cannot find anywhere in the designs where it specifies the use of heading tags."**
UX just specifies a heading but we should allow for different heading levels so people can create the correct document outline for accessibility, so we shouldn't hardcode a heading level like in the modal.

**"default slot and the user can slot whatever they like eg a heading tag"**
Yes, that would be ideal & I was originally slotting in  h1, h2, h3...tags but there are global styles on the h* tags that are taking precedence. I'd like to avoid light dom styles here if at all possible.

**"issue if a user was to slot just a text node "**
There will have to be some tag present if even just to be able to apply the slot attribute so I don't think you can just insert a text node without some sort of wrapper.

Ping me if you have a better idea & we can tinker with it.

packages/genesys-spark-components/src/components/beta/gux-side-sheet/example.html
conor-darcy: @gavin-everett-genesys yes, 100%, just looking for API feedback at the moment.

packages/genesys-spark-components/src/components/beta/gux-side-sheet/gux-side-sheet-heading.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  componentWillLoad(): void {
+    trackComponent(this.root);
+    this.HeadingTag = `h${this.level}` as GuxSideHeadingTag;
+  }
+
+  render(): JSX.Element {
+    return (
+      <this.HeadingTag>
+        <slot />
```
gavin-everett-genesys, line 35: I think we should use a named slot here.

samlimby-genesys:
Hi @conor-darcy whats the status of this component in terms of merging into the gux main library?

conor-darcy:
@samlimby-genesys just waiting for the current tokens work to be completed first as I need the tokens included there.

PR #539, opened 2024-07-18
fix(flyout-menu): Fix menu arrow position

321gillian:
✅ Closes: COMUI-2989

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-flyout-menu/gux-flyout-menu.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-flyout-menu/gux-flyout-menu.scss
@@ -27,13 +27,25 @@
 
   .gux-arrow {
     position: absolute;
-    z-index: -1;
     width: var(--gse-ui-popover-anchor-width);
     height: var(--gse-ui-popover-anchor-height);
-    background: var(--gse-ui-popover-backgroundColor);
-    border: none;
-    box-shadow: var(--gse-ui-menu-boxShadow-x) var(--gse-ui-menu-boxShadow-y)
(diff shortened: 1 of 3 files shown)


jordanstith15 (approved):

gavin-everett-genesys:
Do you know what may have caused this to happen ?

321gillian:
> Do you know what may have caused this to happen ?

This component uses tokens  --gse-ui-popover-anchor-width and --gse-ui-popover-anchor-height. These were 13px x 13px but they changed in v4.48.0 to be 18px x 9px. This was to accommodate a change to the arrow on gux-popover but I think it wasn't clear at that time the knock-on to gux-flyout-menu. 

I was talking to Dorka earlier about adding some tokens specific to the menu for anchor width and height so the dependency is clearer on the tokens side. They would reference the same core/semantic token but be more explicit. I should be able to add those in before it's merged. 

I think it's still better in this case to update the flyout-menu arrow to look at bit more like the newer popover version but ideally with its own token.

321gillian:
> > Do you know what may have caused this to happen ?
> 
> This component uses tokens --gse-ui-popover-anchor-width and --gse-ui-popover-anchor-height. These were 13px x 13px but they changed in v4.48.0 to be 18px x 9px. This was to accommodate a change to the arrow on gux-popover but I think it wasn't clear at that time the knock-on to gux-flyout-menu.
> 
> I was talking to Dorka earlier about adding some tokens specific to the menu for anchor width and height so the dependency is clearer on the tokens side. They would reference the same core/semantic token but be more explicit. I should be able to add those in before it's merged.
> 
> I think it's still better in this case to update the flyout-menu arrow to look at bit more like the newer popover version but ideally with its own token.

Adding a todo here for the extra tokens to allow Ux some time to work on them and so I can fix the menu

PR #546, opened 2024-07-22
fix(form-field-range): Fix console error on updatePosition before componentWillLoad

321gillian:
✅ Closes: COMUI-2970


gavin-everett-genesys:
I was just wondering do you have some markup that I can use to display the error ?

321gillian:
> I was just wondering do you have some markup that I can use to display the error ?

I couldn't find a reliable way. I set up an angular 18 playground and tried to do it that way but it wasn't really present. I ended up just commenting out the `setInput` in `componentWillLoad` so I could see how the defaults were used in `updatePosition`.

gavin-everett-genesys:
Seems like a timing issue as for me the `componentWillLoad` always executes before the `OnResize` decorator.

gavin-everett-genesys (approved):

PR #548, opened 2024-07-23
feat(form-fields): Add optional indicator to form field labels 

321gillian:
COMUI-2773


gavin-everett-genesys:
Changes LGTM, I think you just have two merge conflicts 👍

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-color/example.html
```diff
@@ -86,6 +86,20 @@ <h2>Examples</h2>
...
+    <gux-form-field-color indicator-mark="optional">
+      <input slot="input" type="color" name="e-1" value="#cc3ebe" />
+      <label slot="label">Default</label>
+    </gux-form-field-color>
+
+    <gux-form-field-color indicator-mark="optional">
+      <input slot="input" type="color" name="e-1" value="#cc3ebe" required />
+      <label slot="label">Default</label>
+    </gux-form-field-color>
+  </fieldset>
```
gavin-everett-genesys, line 102: Just had a quick question , if the `indicator-mark` is `optional` and `required` is set on the input does that mean nothing will display ? or should one of them take precedence.

321gillian: - if indicatorMark is 'optional' and input is 'required' then show nothing
- if indicatorMark is 'required' and input is 'required' then show asterisk
- if indicatorMark is 'optional' and input is not required then show(optional)
- if indicatorMark is 'required' and input is not required then show nothing

That's the way it works. The idea is that the default is show asterisk for required and nothing for optionals. However, if you have a set of form fields where the vast majority are required then you may decide to just indicate the optional ones instead.

gavin-everett-genesys (approved):

PR #549, opened 2024-07-24
feat(time-zone): Add gux-time-zone to v4

321gillian:
This component was added to v3 but never v4 back in September 2023. There was a request to add it to v4 and I don't see any obvious reason not to but I'm open to correction. 

✅ Closes: COMUI-2956

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-time-zone/gux-time-zone.service.ts
+++ b/packages/genesys-spark-components/src/components/beta/gux-time-zone/gux-time-zone.service.ts
@@ -1,0 +1,19 @@
+import { GuxTimeZoneIdentifier } from '../../../i18n/time-zone/types';
+import { getTimeZoneList, formatOffset } from '../../../utils/date/time-zone';
+import { GuxTimeZoneListing } from '../gux-time-zone-picker/gux-time-zone-picker.types';
+
+export function shortenZone(zone: string): string {
+  const sections = zone.split('/');
+  return sections?.pop() || zone;
+}
+
+export function getLocalizedOffset(
+  localizedUTC: string,
+  timeZoneId: GuxTimeZoneIdentifier
+): string {
+  const zoneList: GuxTimeZoneListing[] = getTimeZoneList();
+  const timeZone = zoneList.find(zone => zone.name === timeZoneId);
+
+  const formattedOffset = formatOffset(timeZone?.currentTimeOffsetInMinutes);
+  return `${localizedUTC}${formattedOffset}`;
+}
--- a/packages/genesys-spark-components/src/components/beta/gux-time-zone/gux-time-zone.tsx
+++ b/packages/genesys-spark-components/src/components/beta/gux-time-zone/gux-time-zone.tsx
@@ -1,0 +1,72 @@
+import { Component, JSX, h, Host, Prop, Element } from '@stencil/core';
+import { trackComponent } from '@utils/tracking/usage';
+import { buildI18nForComponent, GetI18nValue } from '../../../i18n';
+import translationResources from '../gux-time-zone-picker/i18n/en.json';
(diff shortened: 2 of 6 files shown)


packages/genesys-spark-components/src/components/beta/gux-time-zone/gux-time-zone.service.ts
jasonmobley, line 1: Would it be possible to implement these formatting helpers in the genesys-spark package? That would let consumers get the localized formatted time zone name whether we're on v3 or v4 of components and would allow for cases where the text is needed but can't be used as a component (e.g. in an attribute value like a `screenreader-text` or as a templated part of a larger localization string).

jasonmobley: I have a separate ticket for that, but it might be easier to implement now rather than coming back to refactor later on.

https://inindca.atlassian.net/browse/COMUI-2972

321gillian: Hi Jason, 

I'd prefer to deal with this part of the request as is before reviewing the other ticket. 

Thanks,
Gillian

katie-bobbe-genesys (approved):

packages/genesys-spark-components/src/components/beta/gux-time-zone/gux-time-zone.service.ts
```diff
@@ -0,0 +1,19 @@
+import { GuxTimeZoneIdentifier } from '../../../i18n/time-zone/types';
+import { getTimeZoneList, formatOffset } from '../../../utils/date/time-zone';
+import { GuxTimeZoneListing } from '../gux-time-zone-picker/gux-time-zone-picker.types';
+
+export function shortenZone(zone: string): string {
```
jason-evans-genesys, line 5: Is there any way to document here an example of a zone string input and how it'd be transformed to then be returned?

packages/genesys-spark-components/src/components/beta/gux-time-zone/gux-time-zone.tsx
```diff
@@ -0,0 +1,72 @@
...
+  timeZoneId: GuxTimeZoneIdentifier;
+
+  /**
+   * True to display the zone's offset from UTC
+   */
+  @Prop()
+  offset: boolean;
+
+  /**
+   * True to display surround the offset with parentheses
```
jason-evans-genesys, line 35: This phrasing is confusing to me: `True to display surround the offset with parentheses`

packages/genesys-spark-components/src/components/beta/gux-time-zone/gux-time-zone.tsx
```diff
@@ -0,0 +1,72 @@
...
+    if (this.offset) {
+      const localizedUTC = this.i18n('UTC');
+      const localizedOffset = getLocalizedOffset(localizedUTC, this.timeZoneId);
+
+      displayText = `${localizedZone} ${localizedOffset}`;
+      if (this.surroundOffset) {
+        displayText = `${localizedZone} (${localizedOffset})`;
+      }
+    }
+    return (<span>{displayText}</span>) as JSX.Element;
```
jason-evans-genesys, line 61: I feel like the code would be more testable if instead this function returned `displayText` and then the consuming code rendered the surrounding span tag.

packages/genesys-spark-components/src/components/beta/gux-time-zone/gux-time-zone.service.ts
katie-bobbe-genesys: This component has already been in the code base before, so there isn't new code to review in this PR. Any changes to the component itself (if needed) should be in another PR. This is to avoid potentially introducing bugs at this stage.

jason-evans-genesys (approved):

PR #553, opened 2024-07-25
docs(form-field-radio-group): update html file to reflect UX guidelines

katie-bobbe-genesys:
Update the example html files to match UX usage guidelines for the `form-field-checkbox-group-beta` and the `form-field-radio-group-beta` components.

**Question**: Is it enough to just update the example file to match UX guidelines? Or should we add `log warn` for combinations of uses that are not part of the UX guidelines? 

For example, a UX guideline is that the Error text slot should override the Help text slot. Should we add logic to the component to constrain users to this usage guideline?

It is my inclination to not impose strict usage guidelines within the component itself, instead we just update the examples so that users follow the best practices.


321gillian (approved):
What you've done looks good to me, also a very good question to bring up.

PR #575, opened 2024-08-27
style(modal): dynamic gux-modal footer buttons getting cut off 

gavin-everett-genesys:
Apply padding-top to modal-header instead of container. The ticket also mentions that the `dialog` element has scrollbars but these were removed in a previous PR.

✅ Closes: COMUI-2998

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-modal/gux-modal.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-modal/gux-modal.scss
@@ -35,7 +35,6 @@ slot[name='start-align-buttons']::slotted(:not(button, gux-button-slot)) {
     .gux-modal-container {
       display: flex;
       flex-direction: column;
-      padding-top: var(--gse-ui-modal-padding);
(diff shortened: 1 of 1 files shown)


321gillian:
Screenshot 2024-08-27 at 14 17 08

The padding at the bottom of the modal seems to be spilling out as well so that the amount of space between the footer and then bottom edge of the modal container doesn't look like it matches the 32px that's in Figma. Worth moving that padding as well?

gavin-everett-genesys:
> Screenshot 2024-08-27 at 14 17 08
> 
> The padding at the bottom of the modal seems to be spilling out as well so that the amount of space between the footer and then bottom edge of the modal container doesn't look like it matches the 32px that's in Figma. Worth moving that padding as well?

Updated now, its weird because if you view the spark website (https://spark.genesys.com/latest/components/feedback/modals/design-IxFyKrhe) and look at the layout and spacing its different to the figma design. Will bring this up with UX. I assume the figma design is the correct version and the website needs updating.

daragh-king-genesys (approved):

PR #576, opened 2024-08-27
fix(tooltip-beta): change tooltip-beta based on base display

321gillian:
There were some issues flagged while looking at the avatar tooltip and using axeDevTools. These indicated that the tooltip-beta element didn't have any text visible to screenreader. 

The issue seems to be that the role of "tooltip" can be present in the DOM when the tooltip content is hidden so to the tests it looks like a tooltip with no info. Added a condition here to only give this element the role of tooltip when the tooltip content is present. 

✅ Closes: COMUI-3055


packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/tests/gux-tooltip.e2e.ts
```diff
@@ -29,6 +29,8 @@ describe('gux-tooltip-beta', () => {
           'gux-tooltip-beta >>> gux-tooltip-base-beta'
         );
         expect(baseTooltip.className.split(' ')).not.toContain('gux-show');
+        await a11yCheck(page);
```
321gillian, line 32: Added this to verify that there are no issues prior to the tooltip being activated.

packages/genesys-spark-components/src/components/beta/gux-tooltip-base/gux-tooltip-base.tsx
```diff
@@ -83,6 +85,9 @@ export class GuxTooltipBase {
   @State()
   isShown: boolean = false;
 
+  @Event()
+  guxshown: EventEmitter<boolean>;
```
daragh-king-genesys, line 89: using other components as a reference I think we should have two separate events for show and hide.
see `gux-accordion-section` - `guxopened`/`guxclosed` and `gux-dropdown` - `guxexpanded`/`guxcollapsed`

packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/gux-tooltip.tsx
```diff
@@ -100,7 +110,7 @@ export class GuxTooltip {
 
   render(): JSX.Element {
     return (
-      <Host id={this.id} role="tooltip">
+      <Host id={this.id} role={this.tooltipShown ? 'tooltip' : null}>
```
daragh-king-genesys, line 113: are you sure we have to put the role on this element? Would it work if the role was on the base element?

packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/gux-tooltip.tsx
```diff
@@ -100,21 +97,18 @@ export class GuxTooltip {
 
   render(): JSX.Element {
     return (
-      <Host id={this.id} role="tooltip">
```
321gillian, line 103: I wasn't sure what this was bringing to the table for the component but I may have just missed it. Also, it meant that the ids on the tooltip-beta and the tooltip-base were the same which seemed unusual to me.

daragh-king-genesys: The `id` is needed as this component will be the one in the light dom

packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/gux-tooltip.scss
```diff
@@ -0,0 +1,7 @@
+:host {
+  display: none;
+}
+
+:host(.gux-show) {
+  display: block;
+}
```
321gillian, line 7: Mimic the display state of the base-tooltip in light dom elements so there's no tooltip role without shadow dom tooltip content.

packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/gux-tooltip.tsx
```diff
@@ -72,10 +76,34 @@ export class GuxTooltip {
     trackComponent(this.root, { variant: this.placement });
   }
 
+  componentDidLoad(): void {
+    if (!this.tooltipObserver && window.MutationObserver) {
+      this.tooltipObserver = new MutationObserver(
+        (): string => (this.role = this.baseTooltip?.role)
+      );
```
321gillian, line 83: I tried out the @OnMutation decorator but it didn't seem to pick up the role change on the base-tooltip.

packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/gux-tooltip.tsx
```diff
@@ -28,10 +28,14 @@ export class GuxTooltip {
   private root: HTMLElement;
   private baseTooltip: HTMLGuxTooltipBaseBetaElement;
   private id: string = randomHTMLId('gux-tooltip');
+  private tooltipObserver?: MutationObserver;
```
jason-evans-genesys, line 31: What happens if this is not optional?

packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/gux-tooltip.tsx
```diff
@@ -72,10 +76,34 @@ export class GuxTooltip {
...
   }
 
+  componentDidLoad(): void {
+    if (!this.tooltipObserver && window.MutationObserver) {
+      this.tooltipObserver = new MutationObserver(
+        (): string => (this.role = this.baseTooltip?.role)
+      );
+    }
+
+    if (this.tooltipObserver) {
```
jason-evans-genesys, line 86: Would everything still work if we moved lines 87-94  below line 84 so it's within the first `if` block in `componentDidLoad`, instead of having separate `if` blocks?

321gillian: grouped together now.

packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/gux-tooltip.tsx
```diff
@@ -72,10 +76,30 @@ export class GuxTooltip {
     trackComponent(this.root, { variant: this.placement });
   }
 
+  componentDidLoad(): void {
+    if (!this.tooltipObserver && this.baseTooltip) {
```
jason-evans-genesys, line 80: Will there ever be a time when `this.tooltipObserver` is true when this code block runs? Trying to think if on compnoentDidLoad if `this.tooltipObserver` is always going to be false since `this.tooltipObserver.disconnect();` is called on line 99. Maybe I'm wrong though but just wanted to check.

321gillian: I think that makes sense. I've updated it again

jason-evans-genesys (approved):

daragh-king-genesys (approved):

PR #579, opened 2024-08-29
feat(other): add browser region override support for Hong Kong and Si…

katie-bobbe-genesys:
…ngapore regions


packages/genesys-spark-components/src/genesys-spark-utils/intl.ts
```diff
@@ -84,11 +84,23 @@ export function determineDisplayLocale(
...
+    browserLocaleOverride(localeString) ||
+    (localeString.length == 2 &&
+      navigator.language?.startsWith(`${localeString}-`))
   );
 }
 
+function browserLocaleOverride(localeString: string): boolean {
+  switch (localeString.toLowerCase()) {
+    case 'zh-cn':
+      return !!navigator.language?.toLowerCase().startsWith('zh-sg');
```
gavin-everett-genesys, line 96: Would  `return  navigator.language?.toLowerCase().startsWith('zh-sg') || false` work also ? I always find it hard to parse the double negation pattern 😆

daragh-king-genesys: `Boolean(navigator.language?.toLowerCase().startsWith('zh-hk'))` is also an option :)

katie-bobbe-genesys:
I made a separate ticket for general utils refactors here COMUI-3087

PR #581, opened 2024-09-03
fix(phone-input): added fallbacks for international toll-free placeho…

thomas-c-dillon:
…lder numbers

✅ Closes: COMUI-3062


321gillian (approved):
Tested this on my local. Looks good

PR #586, opened 2024-09-05
(blank-state) - tokens removed/updated

gavin-everett-genesys:
I was just wondering when I run `new-tokens-data` should the old files be updating as-well for example the tokens will still exist in `_gse-ui-feedback.scss` or is it okay for those files to go out of sync.


321gillian:
> I was just wondering when I run `new-tokens-data` should the old files be updating as-well for example the tokens will still exist in `_gse-ui-feedback.scss` or is it okay for those files to go out of sync.

I think on the UX side they need to update the `data-old/tokens.json` file as well as the new stuff. Had a similar thing in the form field gap changes lately

VirtualClover:
cc. @MonseGarciaLimon

daragh-king-genesys:
@gavin-everett-genesys can you rebase this on main

gavin-everett-genesys:
I have resolved the conflicts on this PR now I had one question should we also be updating the `data-old/tokens.json` file also to remove the tokens from there ?

321gillian:
Screenshot 2024-10-29 at 11 26 09
Does this one need to go too?

321gillian (approved):

PR #590, opened 2024-09-06
feat(time-picker): Added min and max props (IN PROGRESS)

jason-evans-genesys:
✅ Closes: COMUI-1736

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-time-picker/example.html
+++ b/packages/genesys-spark-components/src/components/stable/gux-time-picker/example.html
@@ -45,6 +45,18 @@
 
   <h3>Has Error</h3>
   <gux-time-picker has-error></gux-time-picker>
+
+  <h3>Min and Max with 12 Hour Time Format</h3>
+  <gux-time-picker interval="15" min="3:15" max="10:45"></gux-time-picker>
+
+  <h3>Min and Max with 24 Hour Time Format</h3>
+  <gux-time-picker
+    value="16:15"
+    interval="15"
+    min="3:15"
+    max="10:45"
+    clock-type="24h"
+  ></gux-time-picker>
 </form>
 
 <h2>Languages</h2>
--- a/packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.service.ts
+++ b/packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.service.ts
@@ -12,7 +12,9 @@ import {
 
 export function getTimeDisplayValues(
   minuteInterval: GuxMinuteInterval,
-  clockType: GuxClockType
+  clockType: GuxClockType,
+  min: string,
+  max: string
 ): GuxISOHourMinute[] {
   const minuteOptions = [0, 15, 30, 45]
     .filter(option => Number.isInteger(option / minuteInterval))
@@ -22,13 +24,49 @@ export function getTimeDisplayValues(
       ? ['12', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11']
       : Array.from(Array(24).keys()).map(x => String(x).padStart(2, '0'));
 
-  return hourOptions.reduce((acc, hourOption) => {
+  const hourOptionsFormatted = hourOptions.reduce((acc, hourOption) => {
     return acc.concat(
       minuteOptions.map(
         minuteOption => `${hourOption}:${minuteOption}`
       ) as GuxISOHourMinute[]
     );
   }, [] as GuxISOHourMinute[]);
+
+  return applyHourBoundaries(clockType, min, max, hourOptionsFormatted);
+}
+
+function applyHourBoundaries(
+  clockType: GuxClockType,
+  min: string,
+  max: string,
+  hourOptions: string[]
+) {
+  let hourOptionsFiltered = [...hourOptions];
+
+  // TODO: should I validate min and/or max prop's format that's passed in by the developer or is that unnecessary?
+
+  if (min) {
+    if (clockType === '12h' && !min.includes('12:')) {
+      // If min value's hour is not 12 then remove it as an option
+      hourOptionsFiltered = hourOptions.filter(h => !h.includes('12:'));
+    }
+
+    const minAsNumber = min ? parseInt(min.replace(':', ''), 10) : 0;
+    hourOptionsFiltered = hourOptionsFiltered.filter(hour => {
+      const hourAsNumber = hour.replace(':', '');
+      return parseInt(hourAsNumber, 10) > minAsNumber;
+    });
+  }
+
+  if (max) {
+    const maxAsNumber = min ? parseInt(max.replace(':', ''), 10) : 0;
+    hourOptionsFiltered = hourOptionsFiltered.filter(hour => {
+      const hourAsNumber = hour.replace(':', '');
+      return parseInt(hourAsNumber, 10) < maxAsNumber;
+    });
+  }
+
+  return hourOptionsFiltered as GuxISOHourMinute[];
 }
 
 export function getLocaleClockType(root: HTMLElement): GuxClockType {
--- a/packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.tsx
@@ -67,6 +67,12 @@ export class GuxTimePicker {
   @Prop({ mutable: true })
   clockType: GuxClockType;
 
+  @Prop()
+  min?: string;
+
+  @Prop()
+  max?: string;
+
   @State()
   expanded: boolean = false;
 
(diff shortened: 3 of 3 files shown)


packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.service.ts
```diff
@@ -22,13 +24,49 @@ export function getTimeDisplayValues(
...
+
+function applyHourBoundaries(
+  clockType: GuxClockType,
+  min: string,
+  max: string,
+  hourOptions: string[]
+) {
+  let hourOptionsFiltered = [...hourOptions];
+
+  // TODO: should I validate min and/or max prop's format that's passed in by the developer or is that unnecessary?
```
jason-evans-genesys, line 46: Question here. What do you all think?

packages/genesys-spark-components/src/components/stable/gux-time-picker/tests/gux-time-picker.service.spec.ts
```diff
@@ -323,9 +323,9 @@ describe('gux-time-picker.service', () => {
         expectedOutput: GuxISOHourMinute[];
       }) => {
         it(`should work as expected for ${clockType} with ${minuteInterval} minute intervals`, async () => {
-          expect(getTimeDisplayValues(minuteInterval, clockType)).toStrictEqual(
-            expectedOutput
-          );
+          expect(
+            getTimeDisplayValues(minuteInterval, clockType, '0', '5:00')
```
jason-evans-genesys, line 327: I'm going to write unit tests here and will not hardcode "0" and "5:00" after that

packages/genesys-spark-components/src/components/stable/gux-time-picker/example.html
```diff
@@ -45,6 +45,18 @@ <h3>Disabled</h3>
...
   <gux-time-picker has-error></gux-time-picker>
+
+  <h3>Min and Max with 12 Hour Time Format</h3>
+  <gux-time-picker interval="15" min="12:15" max="5:30"></gux-time-picker>
+
+  <h3>Min and Max with 24 Hour Time Format</h3>
+  <gux-time-picker
+    value="16:15"
+    interval="15"
+    min="3:15"
```
321gillian, line 56: Should 24hr format be 03:15?

packages/genesys-spark-components/src/components/stable/gux-time-picker/tests/gux-time-picker.service.spec.ts
```diff
@@ -331,6 +331,71 @@ describe('gux-time-picker.service', () => {
...
+  describe('#getTimeDisplayValues with set boundaries', () => {
+    it(`should work as expected for 24h with 30 minute intervals and boundaries of 11:00-20:00`, async () => {
+      const minuteInterval = 30;
+      const clockType = '24h';
+      const min = '11:00';
+      const max = '20:00';
+      const expectedOutput = [
+        '11:30',
+        '12:00',
+        '12:30',
```
321gillian, line 343: Might just be worth adding a min of a single digit number to verify how that formatting works.

jason-evans-genesys: Great catch. This has been added now.

packages/genesys-spark-components/src/components/stable/gux-time-picker/example.html
jason-evans-genesys: Great catch. Fixed.

packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.service.ts
```diff
@@ -22,13 +24,51 @@ export function getTimeDisplayValues(
...
+  return applyHourBoundaries(clockType, hourOptionsFormatted, min, max);
+}
+
+function applyHourBoundaries(
+  clockType: GuxClockType,
+  hours: string[],
+  min?: string,
+  max?: string
+) {
+  // TODO: should I validate min and/or max prop's format that's passed in or is that unnecessary?
```
jason-evans-genesys, line 44: Question for the team here.

321gillian:
LGTM but I think worth bringing up the questions you have with the team

gavin-everett-genesys:
I was just reading up on how the `start` and `end` would work if you were crossing over midnight for example if you have a `min` of `22:00` and an `end`  of `04:00`. Apparently for the native timepicker it wraps around midnight . So it would,

- The valid time range would start at 22:00 (10:00 PM) and continue until 23:59 (11:59 PM).
- Then, it wraps around midnight, continuing from 00:00 (12:00 AM) to 04:00 (4:00 AM).

Still uncertain how it works in terms of wrapping around midnight but I think you could potentially create a validation function .I think it might be possible to convert the `start` and `end` times into a comparable format like seconds from midnight. You could then check  If `min` is less than `max` check if the  time value  falls between min and max so eg
```
value = 01:00 , start = 22:00 , end = 04:00

if ( 22:00 < 04:00 ) then
// no wrap around
      return 1:00 >= 22:00 && 1:00 <= 04:00 
else
//wrap around
    return 1:00 >= 22:00 || 1:00 <= 04:00
```


so this should return true. This is just a rough idea of what I was thinking about but not sure if we want to be thinking about the crossover.

jason-evans-genesys:
> I was just reading up on how the `start` and `end` would work if you were crossing over midnight for example if you have a `min` of `22:00` and an `end` of `04:00`. Apparently for the native timepicker it wraps around midnight . So it would,
> 
>     * The valid time range would start at 22:00 (10:00 PM) and continue until 23:59 (11:59 PM).
> 
>     * Then, it wraps around midnight, continuing from 00:00 (12:00 AM) to 04:00 (4:00 AM).
> 
> 
> Still uncertain how it works in terms of wrapping around midnight but I think you could potentially create a validation function .I think it might be possible to convert the `start` and `end` times into a comparable format like seconds from midnight. You could then check If `min` is less than `max` check if the time value falls between min and max so eg
> 
> ```
> value = 01:00 , start = 22:00 , end = 04:00
> 
> if ( 22:00 < 04:00 ) then
> // no wrap around
>       return 1:00 >= 22:00 && 1:00 <= 04:00 
> else
> //wrap around
>     return 1:00 >= 22:00 || 1:00 <= 04:00
> ```
> 
> so this should return true. This is just a rough idea of what I was thinking about but not sure if we want to be thinking about the crossover.

@gavin-everett-genesys This is ready for review again. Thanks!

packages/genesys-spark-components/src/components/stable/gux-time-picker/example.html
```diff
@@ -45,6 +45,27 @@ <h3>Disabled</h3>
 
   <h3>Has Error</h3>
   <gux-time-picker has-error></gux-time-picker>
+
+  <h3>Min of "03:15" and Max of "10:45" with 24 Hour Time Format</h3>
```
katie-bobbe-genesys, line 49: I would maybe expect that min and max are inclusive of the min and max values. For example, I would think that 3:15 and 10:45 should be included in the list?
image

katie-bobbe-genesys:
Should `min` and `max` work with the 12 hour format as well? Pressing the AM/PM button allows for selecting "out of bounds" times in 12 hour format mode

jason-evans-genesys:
> Should `min` and `max` work with the 12 hour format as well? Pressing the AM/PM button allows for selecting "out of bounds" times in 12 hour format mode

@daragh-king-genesys I think you had mentioned we shold not do the min and max for the 12 hour format but correct me if I'm wrong.

packages/genesys-spark-components/src/components/stable/gux-time-picker/example.html
jason-evans-genesys: That's a good catch. I'll make a fix for this.

katie-bobbe-genesys:
> > Should `min` and `max` work with the 12 hour format as well? Pressing the AM/PM button allows for selecting "out of bounds" times in 12 hour format mode
> 
> @daragh-king-genesys I think you had mentioned we should not do the min and max for the 12 hour format but correct me if I'm wrong.

If it is the case that 12 hour format is not supported I would suggest the following
1. do a check like `if format is not 12hour and there is min / max prop` then use the min max prop logic, otherwise ignore it
2. perhaps consider logWarn if the format is 12hr and there is min/max prop added

packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.tsx
```diff
@@ -102,6 +108,10 @@ export class GuxTimePicker {
 
     this.i18n = await buildI18nForComponent(this.root, translationResources);
     this.clockType = this.clockType || getLocaleClockType(this.root);
+
+    if (this.clockType == '12h' && (this.min || this.max)) {
+      console.error('clock type must be "24h" when using min/max props');
```
jason-evans-genesys, line 113: @katie-bobbe-genesys I had this code here already to check if the clock type is 12h but I used a console error. Should I keep this code but change to a console warn? I'll also do an `if` check like you mentioned to ignore the min/max logic in a certain scenario.

packages/genesys-spark-components/src/components/stable/gux-time-picker/example.html
jason-evans-genesys: fixed

packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.service.ts
```diff
@@ -22,13 +24,59 @@ export function getTimeDisplayValues(
...
-  return hourOptions.reduce((acc, hourOption) => {
+  const hourOptionsFormatted = hourOptions.reduce((acc, hourOption) => {
     return acc.concat(
       minuteOptions.map(
         minuteOption => `${hourOption}:${minuteOption}`
       ) as GuxISOHourMinute[]
     );
   }, [] as GuxISOHourMinute[]);
+
+  return clockType === '24h'
```
jason-evans-genesys, line 35: @katie-bobbe-genesys this is where I check if the clocktype is 24h before applying the min/max check.

katie-bobbe-genesys (approved):

PR #598, opened 2024-09-10
chore(other): updated some dependencies to latest

daragh-king-genesys:
updated some dependencies to latest

✅ Closes: COMUI-2775


jordanstith15 (changes requested):
Looks like the TimeZoneList length has changed and is breaking the tests

jordanstith15 (approved):
I do think it's a bit weird that we are testing the length of an array we largely don't control.. maybe it should be switched to check the array exists or length is greater than 0 (or the length of what we add to it ourselves)

PR #603, opened 2024-09-13
fix: gux-button accessibility with icon-only and gux-title

jordanstith15:
Closes [COMUI-3044](https://inindca.atlassian.net/browse/COMUI-3044)

[COMUI-3044]: https://inindca.atlassian.net/browse/COMUI-3044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/genesys-spark-components/src/components/beta/gux-tooltip-base/gux-tooltip-base.tsx
```diff
@@ -204,6 +213,7 @@ export class GuxTooltipBase {
         id={this.tooltipId ? undefined : this.id}
         role={this.tooltipId && !this.isShown ? undefined : 'tooltip'}
```
321gillian, line 214: Is it bad to have aria-hidden with role tooltip? I don't have any strong thoughts on it but just wondering

katie-bobbe-genesys:
Is there a reason why the following existing implementations can't be used instead? Let me know if I'm missing something but at first glance these two implementations seem sufficient:

using **gux-button**
```
<gux-button>
    <gux-icon icon-name="user-add" decorative></gux-icon>
    <gux-tooltip-beta>
      <div slot="content">Add User</div>
    </gux-tooltip-beta>
</gux-button>
```
using **gux-button-slot**
```
  <gux-button-slot>
    <button type="button" onclick="notify(event)">
    <gux-icon icon-name="user-add"></gux-icon>
    <gux-tooltip-beta>
      <div slot="content">Add User</div>
    </gux-tooltip-beta>
    </button>
  </gux-button-slot>
```

packages/genesys-spark-components/src/components/beta/gux-tooltip-base/gux-tooltip-base.tsx
jordanstith15: It didn't seem to affect anything in my testing with voiceover or in the scan with axe, so I guess it's fine?

jordanstith15:
> Is there a reason why the following existing implementations can't be used instead? Let me know if I'm missing something but at first glance these two implementations seem sufficient:
> 
> using **gux-button**
> 
> ```
> <gux-button>
>     <gux-icon icon-name="user-add" decorative></gux-icon>
>     <gux-tooltip-beta>
>       <div slot="content">Add User</div>
>     </gux-tooltip-beta>
> </gux-button>
> ```
> 
> using **gux-button-slot**
> 
> ```
>   <gux-button-slot>
>     <button type="button" onclick="notify(event)">
>     <gux-icon icon-name="user-add"></gux-icon>
>     <gux-tooltip-beta>
>       <div slot="content">Add User</div>
>     </gux-tooltip-beta>
>     </button>
>   </gux-button-slot>
> ```

@katie-bobbe-genesys I agree with you on this. This is the best option imo, and I asked about it in standups, but was told to keep the current guxTitle functionality (just with our own styled tooltip instead of the "title" attr) instead of removing it entirely.
We already use it like this in the gux-tooltip example page ¯\_(ツ)_/¯

katie-bobbe-genesys:
> > Is there a reason why the following existing implementations can't be used instead? Let me know if I'm missing something but at first glance these two implementations seem sufficient:
> > using **gux-button**
> > ```
> > <gux-button>
> >     <gux-icon icon-name="user-add" decorative></gux-icon>
> >     <gux-tooltip-beta>
> >       <div slot="content">Add User</div>
> >     </gux-tooltip-beta>
> > </gux-button>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > using **gux-button-slot**
> > ```
> >   <gux-button-slot>
> >     <button type="button" onclick="notify(event)">
> >     <gux-icon icon-name="user-add"></gux-icon>
> >     <gux-tooltip-beta>
> >       <div slot="content">Add User</div>
> >     </gux-tooltip-beta>
> >     </button>
> >   </gux-button-slot>
> > ```
> 
> @katie-bobbe-genesys I agree with you on this. This is the best option imo, and I asked about it in standups, but was told to keep the current guxTitle functionality (just with our own styled tooltip instead of the "title" attr) instead of removing it entirely. We already use it like this in the gux-tooltip example page ¯\_(ツ)_/¯

Jordan, could you tag whoever said to keep the `guxTitle` functionality in standup here. I would like to hear their opinion on this too

packages/genesys-spark-components/src/components/beta/gux-tooltip-base/gux-tooltip-base.tsx
```diff
@@ -157,8 +164,10 @@ export class GuxTooltipBase {
 
   private setForElement(): void {
     if (this.forElement) {
-      const tooltipId = this.tooltipId ? this.tooltipId : this.id;
-      this.forElement.setAttribute('aria-describedby', tooltipId);
+      if (!this.visualOnly) {
+        const tooltipId = this.tooltipId ? this.tooltipId : this.id;
+        this.forElement.setAttribute('aria-describedby', tooltipId);
+      }
```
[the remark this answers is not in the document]
jordanstith15: I tested that one too and it seems to override aria-hidden so things were reading out twice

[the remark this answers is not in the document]
321gillian: Sorry, I wrote that in a confusing way. I meant you could add `if (!this.visualOnly || Boolean(this.getAttribute('aria-hidden'))` and then it would carry through.

jordanstith15:
> > > Is there a reason why the following existing implementations can't be used instead? Let me know if I'm missing something but at first glance these two implementations seem sufficient:
> > > using **gux-button**
> > > ```
> > > <gux-button>
> > >     <gux-icon icon-name="user-add" decorative></gux-icon>
> > >     <gux-tooltip-beta>
> > >       <div slot="content">Add User</div>
> > >     </gux-tooltip-beta>
> > > </gux-button>
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > using **gux-button-slot**
> > > ```
> > >   <gux-button-slot>
> > >     <button type="button" onclick="notify(event)">
> > >     <gux-icon icon-name="user-add"></gux-icon>
> > >     <gux-tooltip-beta>
> > >       <div slot="content">Add User</div>
> > >     </gux-tooltip-beta>
> > >     </button>
> > >   </gux-button-slot>
> > > ```
> > 
> > 
> > @katie-bobbe-genesys I agree with you on this. This is the best option imo, and I asked about it in standups, but was told to keep the current guxTitle functionality (just with our own styled tooltip instead of the "title" attr) instead of removing it entirely. We already use it like this in the gux-tooltip example page ¯_(ツ)_/¯
> 
> Jordan, could you tag whoever said to keep the `guxTitle` functionality in standup here. I would like to hear their opinion on this too

@MattCheely Can we get your opinion on this? You were the one said in the EU/APAC standup to keep the guxTitle for now.

Keeping the guxTitle to not cause breaking changes makes sense to me, but should we start having our examples work like Katie's example so that is what it moves to in the future? It makes sense to me since it simplifies our code and gives the consumer more flexibility too. The guxTitle prop doesn't really make sense to have anymore imo especially since we can't use a "title" prop in a button anyways without it clashing with our styling.

MattCheely:
Sorry for the delayed response on this. We have to continue to support the `gux-title` because removing it would break backwards compatibility. It's not clear to me from reviewing the ticket if there's an actual a11y issue with `gux-title` or if it's a false positive from Axe scans related to Shadow DOM. I'd summarize my position as:

- We have to support `gux-title` until the next major release
- `gux-title` needs to be properly accessible. If these changes are required for that, we should implement them.
- If `gux-title` is currently accessible, but is flagged by axe, I'm fine with just changing the example recommendations and not altering the components.

321gillian (approved):

packages/genesys-spark-components/src/components/stable/gux-button/gux-button.tsx
```diff
@@ -92,6 +106,12 @@ export class GuxButton {
       if (child.tagName === 'GUX-ICON') {
         return true;
       }
+    } else if (
+      children.length === 2 &&
+      children[0].tagName === 'GUX-ICON' &&
+      children[1].tagName === 'GUX-TOOLTIP'
+    ) {
+      return true;
```
jordanstith15, line 114: I found out that the gux-tooltip idea did cause an issue with icon-only buttons, making them wider because they didn't have the gux-icon-only class anymore. The gux-tooltip example page uses a text button so it wasn't seen there.
This logic fixes that.

321gillian: Might want to expand this to gux-tooltip-beta also

jordanstith15: Done.
I also noticed a couple other issues that I've fixed.
1. The padding was weird on the tooltip
2. The example ran into the same accessibility issues I noticed before. Updated the example in a way that fixes it.
3. Added examples of buttons with text using a descriptive tooltip too.

321gillian: For the icon buttons, the screenreader is still reading out the 'Visual of screen reader text' for me.

321gillian: `  <gux-button onclick="notify(event)">
    <gux-icon
      icon-name="fa/diamond-regular"
      screenreader-text="This will be read by a screen reader"
    ></gux-icon>
    <gux-tooltip-beta visual-only="true" aria-hidden="true"
      ><span slot="content"
        >Visual of screen reader text</span
      ></gux-tooltip-beta
    >
  </gux-button>`

I can get the screenreader to ignore the text if I add aria-hidden here but not if I add it to the host element in gux-tooltip. I don't know why that happens

jordanstith15: That is really strange since the whole point of the `visualOnly` is that it adds `aria-hidden` to the tooltip.. is this another case where the cases are being weird?

jordanstith15: No.. aria-hidden is correctly on the tooltip.. maybe a shadowdom issue?

321gillian: yeah, it seems like like a shadow dom thing whatever it is

321gillian (approved):

PR #605, opened 2024-09-17
chore(icon): Added search functionality [FEEDBACK]

jason-evans-genesys:
✅ Closes: COMUI-2684

This is a WIP and not code complete yet but I was wondering before I move forward if I could get feedback from the team to see at a high level if this approach I took is ok. There are some things I will address in this PR soon (e.g. the search input total count value not being updated currently, grouping the icons into categories on the icon example page, etc)

This PR addresses adding a search input so that users can filter out the icons on the icon example page. Monse provided me with an icon metadata json file and she said that each icon object in that json file has a description comma separated list that we will use to match against the user's search input value - I'm doing an exact match right now but need to confirm if that's what we want with UX.

The approach I took was to inject the icon metadata json into the DOM via webpack so that the icon example page has access to the icon metadata json when filtering the icons based on the search input value. I don't know if this is the approach we want or if instead we want to make an API call to get the icon metadata JSON. I'm looking for feedback from the team on this.

I'm also wondering what the team thinks about the best way to group the icons based how they're grouped in this Figma file below. I was thinking we could ask UX if they could make some sort of spreadsheet that associates each icon name with its category and then we could just organize the icons into their own category folder within the `gux-icon/icons` folder:
Screen Shot 2024-09-17 at 12 05 17 PM

https://www.figma.com/design/JKbHmcf4nUF6C7Pj8M6MpY/Spark%3A-Flare-4.0---Core?node-id=16288-7098&node-type=frame&t=1hT1hxAMSPOHxg0L-0

### DIFF
--- a/web-apps/genesys-spark-examples/icon-metadata.json
+++ b/web-apps/genesys-spark-examples/icon-metadata.json
@@ -1,0 +1,3291 @@
+{
+  "envelope-regular": {
+    "description": "mail, email, send, sent, message, post, letter",
+    "documentationLinks": [],
+    "id": "23:5513"
+  },
+  "inbox-regular": {
+    "description": "mail, email, inbox, received, message, post, letter",
+    "documentationLinks": [],
+    "id": "23:5343"
+  },
+  "asterisk-regular": {
+    "description": "star, wildcard, bullet point, 6, emergency",
+    "documentationLinks": [],
+    "id": "3126:132136"
+  },
+  "custom-phone-low-volume-regular": {
+    "description": "call, calling, telephone, landline, talk, communication, ring, active, weak sound waves",
+    "documentationLinks": [],
+    "id": "23:5369"
+  },
+  "custom-phone-volume-regular": {
+    "description": "call, calling, telephone, landline, talk, communication, ring, active, sound waves",
+    "documentationLinks": [],
+    "id": "23:5367"
+  },
+  "custom-phone-high-volume-regular": {
+    "description": "call, calling, telephone, landline, talk, communication, ring, active, strong sound waves",
+    "documentationLinks": [],
+    "id": "23:5353"
+  },
+  "phone-slash-regular": {
+    "description": "call, telephone, landline, talk, communication, disabled, slash",
+    "documentationLinks": [],
+    "id": "23:5365"
+  },
+  "custom-phone-bars-regular": {
+    "description": "call, telephone, landline, talk, communication, history",
+    "documentationLinks": [],
+    "id": "23:5401"
+  },
+  "custom-phone-arrow-left-regular": {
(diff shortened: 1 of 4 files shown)


321gillian:
> ✅ Closes: COMUI-2684
> 
> This is a WIP and not code complete yet but I was wondering before I move forward if I could get feedback from the team to see at a high level if this approach I took is ok. There are some things I will address in this PR soon (e.g. the search input total count value not being updated currently, grouping the icons into categories on the icon example page, etc)
> 
> This PR addresses adding a search input so that users can filter out the icons on the icon example page. Monse provided me with an icon metadata json file and she said that each icon object in that json file has a description comma separated list that we will use to match against the user's search input value - I'm doing an exact match right now but need to confirm if that's what we want with UX.

Probably want to be able to search by name also although I know this is probably for after the initial POC

> The approach I took was to inject the icon metadata json into the DOM via webpack so that the icon example page has access to the icon metadata json when filtering the icons based on the search input value. I don't know if this is the approach we want or if instead we want to make an API call to get the icon metadata JSON. I'm looking for feedback from the team on this.

I am in favour of the idea of looser searching in theory. Would this then require a change to how icons are updated every time? i.e. would the metadata be updated also? 

Also, do we have any data on how devs are using the WCL icon area? Do they use the WCL to find icons or would they check anywhere else first? 

> I'm also wondering what the team thinks about the best way to group the icons based how they're grouped in this Figma file below. I was thinking we could ask UX if they could make some sort of spreadsheet that associates each icon name with its category and then we could just organize the icons into their own category folder within the `gux-icon/icons` folder: Screen Shot 2024-09-17 at 12 05 17 PM

I think if we move the folders then it'll break the current icons that are used as fa/00-regular etc. There'd have to be a workaround to map them back again I think. 

These are just some initial rough thoughts from my side.

web-apps/genesys-spark-examples/icon-metadata.json
```diff
@@ -0,0 +1,3291 @@
...
+    "description": "",
+    "documentationLinks": [],
+    "id": "32119:157"
+  },
+  "In-Line Button": {
+    "description": "",
+    "documentationLinks": [],
+    "id": "769:73609"
+  },
+  "Button": {
```
katie-bobbe-genesys, line 2526: remove other component meta data

katie-bobbe-genesys:
The examples page is a pretty low risk change because since no one is consuming this code, so I would say it's a good opportunity to experiment with new ideas and be creative.

I think using the metadata provided by UX is a good way to search, but my main concern is how we will maintain the metadata json and keep it up to date when new icons are added.  I'm not super fond of the idea of updating a JSON file each time we add a new icon, unless it's automated in some way. 

My advice would be to think of what process makes sense for updating the metadata when new icons are added. Are there any other ways to add the metadata (key words) to the icons that is maintainable long term?

PR #606, opened 2024-09-18
chore(toast): updating to use gux-cta-group

jordanstith15:
Closes [COMUI-3082](https://inindca.atlassian.net/browse/COMUI-3082)

[COMUI-3082]: https://inindca.atlassian.net/browse/COMUI-3082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jordanstith15:
I used "chore" since it doesn't affect the api for the consumer, but is that correct?

321gillian (approved):

321gillian:
> I used "chore" since it doesn't affect the api for the consumer, but is that correct?

This is a good question. I also am unsure about this. Might be one for stand-up

PR #609, opened 2024-09-20
feat(modal): using the cta-group in the modal

conor-darcy:
I've added a new 'footer' slot so as not to introduce a breaking change. Old slots will continue to work & we can remove them in V5.


packages/genesys-spark-components/src/components/stable/gux-modal/example.html
```diff
@@ -45,24 +41,17 @@ <h2>No title</h2>
   >
   <gux-modal id="example2" size="small">
     <div slot="content">This contains the modal content.</div>
-    <div slot="start-align-buttons">
-      <gux-button-slot accent="primary">
-        <button type="button">Save</button>
-      </gux-button-slot>
-      <gux-button-slot>
```
gavin-everett-genesys, line 52: I think you should maybe keep these examples here and include a separate cta group example since they will still be relevant.

packages/genesys-spark-components/src/components/stable/gux-modal/gux-modal.scss
```diff
@@ -93,6 +93,17 @@ slot[name='start-align-buttons']::slotted(:not(button, gux-button-slot)) {
...
       }
+
+      footer {
+        padding-right: var(--gse-ui-modal-padding);
+        padding-bottom: var(--gse-ui-modal-padding);
+        padding-left: var(--gse-ui-modal-padding);
+        margin-top: var(--gse-ui-modal-gap);
+
+        &.gux-no-buttons {
+          display: none;
```
gavin-everett-genesys, line 104: Is this needed here ?

packages/genesys-spark-components/src/components/stable/gux-modal/example.html
conor-darcy: These slots are to be deprecated in V5 so we don't want anybody new using them in the meantime.

packages/genesys-spark-components/src/components/stable/gux-modal/gux-modal.scss
conor-darcy: Yes, there is margin & padding that added unnecessary space when there are no buttons, see the 'Only Content' example.

packages/genesys-spark-components/src/components/stable/gux-modal/example.html
conor-darcy: Here's the V5 ticket:
https://inindca.atlassian.net/browse/COMUI-3138

packages/genesys-spark-components/src/components/stable/gux-modal/gux-modal.scss
gavin-everett-genesys: Ahh okay I was confused because you nesting the class inside the `footer` and are applying it like `footer class="gux-no-buttons"></footer>` but in the tsx file I cannot find where you are rendering it the function just looks like 

```
 private renderFooter(): JSX.Element {
    if (hasSlot(this.root, 'footer')) {
      return (
        <footer>
          <slot name="footer" />
        </footer>
      ) as JSX.Element;
    } else {
      return this.renderButtonFooter();
    }
  }
```

conor-darcy: Sorry @gavin-everett-genesys, yes you are correct, good spot. 
I have 2 `.gux-no-buttons`, don't need the one on the footer (copy & paste probably).

katie-bobbe-genesys (approved):

PR #610, opened 2024-09-20
[WIP]feat(copy-to-clipboard): detect overflow in tooltip position

321gillian:
✅ Closes: COMUI-3077

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-tooltip/gux-tooltip.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-tooltip/gux-tooltip.tsx
@@ -13,6 +13,7 @@ import {
 import {
   autoUpdate,
   computePosition,
+  detectOverflow,
   flip,
   hide,
   offset,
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-tooltip/gux-tooltip.tsx
```diff
@@ -117,7 +118,31 @@ export class GuxTooltip {
...
+
+        if (state.placement.includes('top') && overflowData.top > 0) {
+          return {
+            y: state.y + overflowData.top
+          };
+        }
+
+        return {};
+      }
+    };
```
321gillian, line 143: `gux-copy-to-clipboard` uses `gux-tooltip` and not the new one. Would it be appropriate to update to `gux-tooltip-beta` with this change (if it's required)? 

Also,  I tested and `gux-tooltip-beta` would require updates similar to the ones I've addd in `gux-tooltip` here for this scenario to be supported.

packages/genesys-spark-components/src/components/stable/gux-tooltip/gux-tooltip.tsx
```diff
@@ -117,7 +118,31 @@ export class GuxTooltip {
   }
 
   private updatePosition(): void {
-    const middleware = [offset(16), flip(), shift(), hide()];
+    const overflowDetection = {
+      name: 'overflowDetection',
+      async fn(state) {
+        const overflowData = await detectOverflow(state, {
```
321gillian, line 124: [Floating UI - Detect Overflow](https://floating-ui.com/docs/detectOverflow)

PR #616, opened 2024-09-24
feat(datepicker): allow a width to be applied to the datepicker input

conor-darcy:
This will allow you to set a width on the datepicker with CSS and will stretch the input to that size.


gavin-everett-genesys:
If you apply a width of say `800px` to a datepicker with the mode of `range` the popover will look uncentered. I am not sure if this is ok or ? A datepicker without the mode of `range` looks fine to me.

conor-darcy:
> If you apply a width of say `800px` to a datepicker with the mode of `range` the popover will look uncentered. I am not sure if this is ok or ? A datepicker without the mode of `range` looks fine to me.

Yeah, it does look a little weird but the popover is left aligned so its consistent. probably not really related to this however I've asked UX what their thoughts are & I'll make a new ticket to address it if they want any changes. Maybe they want it aligned to the input & not the component as a whole.

gavin-everett-genesys (approved):

PR #618, opened 2024-09-24
feat(dismiss-button): adding small variant

jordanstith15:


321gillian:
LGTM apart from the token stuff

packages/genesys-spark-components/src/components/stable/gux-dismiss-button/gux-dismiss-button.scss
```diff
@@ -1,7 +1,28 @@
+:host {
+  // all these to be removed when token issues are fixed
```
jason-evans-genesys, line 2: Should we create a JIRA ticket to track this and add a TODO?

packages/genesys-spark-components/src/components/stable/gux-dismiss-button/gux-dismiss-button.tsx
```diff
@@ -30,7 +36,10 @@ export class GuxDismissButton {
     return (
       <gux-button-slot
         accent="ghost"
-        class={this.position == 'inherit' ? 'gux-inherit' : undefined}
+        class={{
+          'gux-inherit': this.position == 'inherit',
```
jason-evans-genesys, line 40: should we have a `===` instead?

packages/genesys-spark-components/src/components/stable/gux-dismiss-button/gux-dismiss-button.scss
jordanstith15: Waiting for Daragh on that. He said he'd have one likely by the end of the week if he can't get his fixes to work.

gavin-everett-genesys (approved):

packages/genesys-spark-components/src/components/stable/gux-dismiss-button/gux-dismiss-button.tsx
```diff
@@ -21,6 +24,9 @@ export class GuxDismissButton {
   @Prop()
   position: GuxDismissButtonPosition = 'absolute';
 
+  @Prop()
+  size: GuxDismissButtonSize = 'medium';
```
jason-evans-genesys, line 28: Just curious if we will only have ever 2 values for this size: `medium` or `small`, or we will have `large` as well in the future?

daragh-king-genesys (approved):

packages/genesys-spark-components/src/components/stable/gux-dismiss-button/gux-dismiss-button.tsx
jordanstith15: I'd prefer to change as little code as possible so I'll keep it that way, but I do think in this case it's unlikely to be much of a difference, especially with the variable having specific strings for types.

PR #627, opened 2024-09-26
fix(popover): Using native popover-api

jason-evans-genesys:
✅ Closes: COMUI-1637

Do we want to also use the popover-api for the popover-list component?

**Update:** I am closing this PR since I opened this new PR to handle creating a new popover-beta component: https://github.com/MyPureCloud/genesys-spark/pull/642

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.scss
@@ -16,6 +16,7 @@
   min-width: 280px;
   padding: var(--gse-ui-popover-gap);
   background-color: var(--gse-ui-popover-backgroundColor);
+  border: none;
   border-radius: var(--gse-ui-popover-borderRadius);
   box-shadow: var(
       --gse-ui-popover-boxShadow-x,
--- a/packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.scss
@@ -1,10 +1,8 @@
+/* This is necessary to override the position of the popover on the page. By default, an HTML popover will
+   render in the middle of the screen, which we do not want */
 :popover-open {
   position: absolute;
   inset: unset;
-  top: 5px;
-  width: 200px;
-  height: 100px;
-  margin: 0;
 }
 
 .gux-popover-wrapper {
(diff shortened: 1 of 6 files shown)


packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
```diff
@@ -213,7 +221,17 @@ export class GuxPopover {
   }
 
   componentDidLoad(): void {
+    if (this.for) {
```
jason-evans-genesys, line 224: Should we make the `for` prop required since the popover requires target button?

packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
```diff
@@ -45,7 +46,7 @@ export class GuxPopover {
    * Indicates the id of the element the popover should anchor to
    */
   @Prop()
-  for: string;
+  for!: string;
```
jason-evans-genesys, line 49: I added this `!` to make the `for` prop required since the popover needs a target element to be clicked to open the popover.

packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.scss
```diff
@@ -6,7 +13,9 @@
   display: inline-block;
   min-width: 280px;
   padding: var(--gse-ui-popover-gap);
+  overflow: hidden;
   background-color: var(--gse-ui-popover-backgroundColor);
+  border: none;
```
jason-evans-genesys, line 18: the native popover was adding a black border

packages/genesys-spark-components/src/components/stable/gux-popover/gux-popover.tsx
```diff
@@ -113,12 +123,32 @@ export class GuxPopover {
     ) {
       this.dismiss();
     }
+
+    if (clickedForElement) {
+      this.popupElement?.showPopover();
```
jason-evans-genesys, line 128: I was getting an error in the popover test file that `this.popupElement?.showPopover` doesn't not exist and I tried to think of a way to mock up this method with jest but was having some trouble. Do you all think using a `?.` here would be ok? Only worry I have is that this would swallow any errors of this method not invoking when it always should in the front end. I think ideally we'd mock up this method because that seems cleaner to me, but wanted to see what you all thought.

daragh-king-genesys:
@jason-evans-genesys can this branch be deleted?

jason-evans-genesys:
> @jason-evans-genesys can this branch be deleted?

@daragh-king-genesys Yes, I will delete it.

PR #632, opened 2024-09-30
feat(table): update select all on addition or removal of rows

conor-darcy:
Select all checkbox will update if a row is deleted from the dom.
I've also added a new method to manually set the state of gux-all-row-select.


packages/genesys-spark-components/src/components/stable/gux-table/gux-all-row-select/gux-all-row-select.tsx
```diff
@@ -49,6 +50,26 @@ export class GuxAllRowSelect {
     this.inputElement.indeterminate = indeterminate;
   }
 
+  @Method()
+  async setCheckedState(checkedState: GuxAllRowSelectState): Promise<void> {
```
daragh-king-genesys, line 54: If this function is useful we should be able to use it in the `updateSelectAllBoxState` method of `gux-table` and it should simplify the code/make it easier to read.

packages/genesys-spark-components/src/components/stable/gux-table/gux-table.tsx
```diff
@@ -42,9 +42,12 @@ export class GuxTable {
   root: HTMLElement;
 
   private resizeObserver: ResizeObserver;
-  private slotObserver: MutationObserver = new MutationObserver(() =>
-    forceUpdate(this)
-  );
+  private slotObserver: MutationObserver = new MutationObserver(() => {
+    if (this.selectAllCheckbox) {
```
daragh-king-genesys, line 46: I am not sure this if guard is needed. The `updateSelectAllBoxState` method that is eventually calls has a guard against the select all checkbox not existing already.

gavin-everett-genesys:
What should the behaviour be if this is set dynamically ? At the moment if I set the checked state dynamically using the function it does work for example I pass `setCheckedState('checked')` it sets the `gux-all-row-select` to `checked` but all of the `gux-row-select` do not change to `checked` which I think they should  or no ?

conor-darcy:
@daragh-king-genesys Keeping the setting of checked & indeterminate states separate to keep it inline with the native checkbox. If we try & update the table rows themeselves based on a possible 3 states we won't know what to do with 'indeterminate' without a lot of work.

packages/genesys-spark-components/src/components/stable/gux-table/example.html
```diff
@@ -438,11 +438,26 @@ <h2>Data table with sortable columns</h2>
...
+  </li>
+</ul>
+<gux-button id="selectAll">Select all</gux-button>
+<gux-button id="deselecAll">Deselect all</gux-button>
 <gux-table object-table selectable-rows>
   <table slot="data">
     <thead>
       <tr data-row-id="head">
-        <th><gux-all-row-select></gux-all-row-select></th>
+        <th><gux-all-row-select id="allRowSelect"></gux-all-row-select></th>
```
gavin-everett-genesys, line 460: I think it might be easier to just query select the `gux-all-row-select` element instead of adding an `id`.

packages/genesys-spark-components/src/components/stable/gux-table/example.html
```diff
@@ -970,7 +985,21 @@ <h2>Data table with fixed columns</h2>
...
 
-    })
+    });
+
+  const selectAll = document.getElementById('selectAll');
+  const deselectAll = document.getElementById('deselecAll');
+  const allRowSelect = document.getElementById('allRowSelect');
+
+  selectAll.addEventListener('click', () => {
+    allRowSelect.setIndeterminate(false);
```
gavin-everett-genesys, line 995: Would it be better to house this within the `setChecked` method in the shadowDOM like,


```
async setChecked(checked: boolean): Promise<void> {
    this.selected = checked;
   this.setIndeterminate(false);
    this.internalallrowselectchange.emit(this.selected);
  }
```
or
```
async setChecked(checked: boolean): Promise<void> {
    this.selected = checked;
   this.inputElement.indeterminate = false
    this.internalallrowselectchange.emit(this.selected);
  }
```
as I am not sure is it valid to call a method with a Watch decorator internally.

packages/genesys-spark-components/src/components/stable/gux-table/example.html
conor-darcy: That would select all the `gux-all-row-select`'s on the page, I just want to select that particular one as it relates to one particular table.

packages/genesys-spark-components/src/components/stable/gux-table/example.html
conor-darcy: Yeah I did initially look at that but decided to keep the setting of checked & indeterminate states separate to keep it inline with the native checkbox as there were a few issues.

Apparently 'indeterminate' is a presentational state, just visual,  while the value of a checkbox is either checked or unchecked. I thought it best to stick with this to avoid any unforeseen issues down the road by diverging from the native checkbox.

conor-darcy:
> What should the behaviour be if this is set dynamically ? At the moment if I set the checked state dynamically using the function it does work for example I pass `setCheckedState('checked')` it sets the `gux-all-row-select` to `checked` but all of the `gux-row-select` do not change to `checked` which I think they should or no ?

I've an example with select/deselect all buttons using the methods on the `gux-all-row-select`.
I don't have an example setting the indeterminate alone as I would not know what rows should be selected or not.

gavin-everett-genesys (approved):

PR #646, opened 2024-10-07
feat(avatar-groups): Create avatar groups 

321gillian:
Very much a WIP just creating this for visibility.


jason-evans-genesys:
I pulled the branch down and spun it up and I see a flickering like in the video below:

https://github.com/user-attachments/assets/575abd9d-c06f-4dd1-9351-71523e5cc41a

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-groups.tsx
```diff
@@ -0,0 +1,229 @@
...
+  private overflowCount: number = 0;
+
+  get countAvatars() {
+    return this.avatarList.length ?? 0;
+  }
+
+  private checkGroupContainerWidthForLayout() {
+    readTask(() => {
+      const container = this.root.shadowRoot.querySelector('.gux-avatar-group');
+      const containerWidth = container.clientWidth;
```
katie-bobbe-genesys, line 37: You might be able to do this in a programmatic way as follows. Here is what I'm thinking
1. remove the hard coded `display:none` breakpoints in the scss file
2.      
Here where you check the containerWidth, should be able to compare it with the scrollWidth to programmatically show and hide avatars
```
const containerWidth = container.clientWidth;
const scrollWidth = container.scrollWidth;
// if scrollWidth > containerWidth remove the last child avatar and recalculate until scrollWidth < containerWidth
      ```
Using the above logic you could add a css class to the avatars you want to hide, a class that has the property `display: none` (instead of defining the breakpoints in the scss file) and add to the tally of `overflowAvatars`

**Comment by katie-bobbe-genesys** [comment]
Overall the API and component itself is looking really good so far 👍

File: packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-groups.tsx (hunk 1 above)
Comment (gavin-everett-genesys) [reply packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-groups.tsx:37]: I agree with @katie-bobbe-genesys  on this as I think the hiding and showing of avatars plus updating of the overflow counter could be done in one file. You might also run into issues aswell with the container queries for example if there can be an unlimited number of `gux-avatars` within a group then the container query will have to be more dynamic to handle that scenario but if theres is a fixed amount it would be fine. Katies idea seems good to me by conditionally adding a class based on the breakpoint widths which you have in the `getMaxAvatarsToDisplay` function already.

File: packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.scss [hunk 2]
```diff
@@ -0,0 +1,83 @@
...
+
+.gux-overflow-list-item {
+  display: flex;
+  gap: 8px;
+  align-items: center;
+  width: max-content;
+}
+
+@container avatar-group (width <= 59px) {
+  ::slotted(gux-avatar-focusable-beta:nth-child(n + 1):nth-child(-n + 7)) {
```
jason-evans-genesys, line 40: Just for my own understanding what does this line of CSS do?

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-groups.tsx
jason-evans-genesys: I also agree with @katie-bobbe-genesys here. Great thinking.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-groups.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  @State()
+  private overflowCount: number = 0;
+
+  get countAvatars() {
+    return this.avatarList.length ?? 0;
+  }
+
+  private checkGroupContainerWidthForLayout() {
+    readTask(() => {
```
jason-evans-genesys, line 35: Could we have one commented line at the top of this function to describe at a high level what this function is doing just for immediate clarity on the functionality here?

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-groups.tsx
```diff
@@ -0,0 +1,229 @@
...
+    this.avatarList = Array.from(
+      this.root.children
+    ) as HTMLGuxAvatarFocusableBetaElement[];
+
+    this.avatarList.forEach((avatar, index) => {
+      const avatarComp = avatar.querySelector(
+        'gux-avatar-beta'
+      ) as HTMLGuxAvatarBetaElement;
+
+      if (avatarComp) {
```
jason-evans-genesys, line 121: What is `avatarComp` exactly? Is that an individual avatar item in the list?

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-groups.tsx
```diff
@@ -0,0 +1,229 @@
...
+        avatarComp.accent = this.getAccent(index) as GuxAvatarAccent;
+      }
+    });
+  }
+
+  private getAccent(index): string {
+    return ((index % 12) + 1).toString();
```
jason-evans-genesys, line 128: why 12 here? Just for my own understanding.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-groups.tsx
```diff
@@ -0,0 +1,229 @@
...
+      this.overflowAvatars.map(focusableAvatar =>
+        this.addListItemForAvatar(focusableAvatar)
+      );
+    }
+  }
+
+  private addListItemForAvatar(
+    focusableAvatar: HTMLGuxAvatarFocusableBetaElement
+  ): void {
+    const listItem = document.createElement('gux-list-item');
```
jason-evans-genesys, line 156: Instead of manually creating elements could we have a function that renders a list of items based on the value of a javascript variable value?

jason-evans-genesys: Also could we document more of the business logic at the top of this function for my clarity of the functionality?

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-groups.tsx
```diff
@@ -0,0 +1,283 @@
...
+  Prop,
+  writeTask
+} from '@stencil/core';
+import { afterNextRenderTimeout } from '@utils/dom/after-next-render';
+import { trackComponent } from '@utils/tracking/usage';
+import { GuxAvatarAccent } from './../gux-avatar/gux-avatar.types';
+import { randomHTMLId } from '@utils/dom/random-html-id';
+
+/**
+ * @slot - Some gux-avatars
```
gavin-everett-genesys, line 17: should this be `@slot - slot for gux-avatar-focusable-beta elements`

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-groups.tsx
```diff
@@ -0,0 +1,283 @@
...
+export class GuxAvatarGroup {
+  @Element()
+  root: HTMLElement;
+
+  private resizeObserver?: ResizeObserver;
+  private listItemClickListeners: Map<HTMLElement, (event: Event) => void> =
+    new Map();
+
+  @Prop()
+  avatarLimit: number = 7;
```
gavin-everett-genesys, line 33: should this be a type `export type GuxAvatarLimit = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7;`

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-groups.tsx
```diff
@@ -0,0 +1,283 @@
...
+  @State()
+  private maxAvatarsToDisplay: number = 0;
+
+  get countAvatars() {
+    return this.avatarList.length ?? 0;
+  }
+
+  async componentWillLoad(): Promise<void> {
+    this.avatarList = Array.from(
+      this.root.children
```
gavin-everett-genesys, line 50: Should there be some sort of validation to assure that these are `gux-avatar-focusable-beta` elements or log a warning if they are not ?

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -0,0 +1,226 @@
...
+import { afterNextRenderTimeout } from '@utils/dom/after-next-render';
+import { trackComponent } from '@utils/tracking/usage';
+import { GuxAvatarAccent } from '../gux-avatar/gux-avatar.types';
+import { randomHTMLId } from '@utils/dom/random-html-id';
+
+/**
+ * @slot - Some gux-avatars
+ */
+
+const MAX_AVATARS = 7;
```
321gillian, line 19: I think initially it would be good to have just a grouped component without it being responsive. 

I had been looking at making it responsive initially but resizing itself on resize was creating a loop that throttling, debouncing or using requestAnimationFrame couldn't fix. At this point, it was kind of getting to be a rabbit hole so I'll pause that effort until later and get the basic overflow working for the minute. 

I looked at other options using container queries or changing the type of slotted items, at this point, I think this is the simplest. Will continue to look at improving the add/remove from popup method. There's no simple way to use slotted content and reconfigure it from what I've found so far.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -0,0 +1,203 @@
...
+})
+export class GuxAvatarGroup {
+  @Element()
+  root: HTMLElement;
+
+  private listItemClickListeners: Map<HTMLElement, (event: Event) => void> =
+    new Map();
+
+  @Prop()
+  avatarLimit: number = 7;
```
gavin-everett-genesys, line 41: Should this be an enum/type with values up to 7 ?

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  private renderListItem(avatar: ProcessedAvatar): JSX.Element {
+    const handleClick = (event: MouseEvent) =>
+      this.handleListItemClick(event, avatar);
+    this.listItemClickListeners.set(avatar.focusableAvatar, handleClick);
+
+    return (
+      <gux-list-item key={avatar.name} onClick={handleClick}>
+        <div class="gux-overflow-list-item">
```
gavin-everett-genesys, line 148: I think this will need to be a `span` element as the `gux-list-item` wraps the slotted content in a `button` tag and slotting a `div` wont be valid HTML.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -0,0 +1,203 @@
...
+  button?: HTMLButtonElement;
+  link?: HTMLAnchorElement;
+  img?: HTMLImageElement;
+  name: string;
+  accent: GuxAvatarAccent;
+  isOverflow: boolean;
+}
+
+/**
+ * @slot - Some gux-avatars
```
gavin-everett-genesys, line 25: Does this slot still exist or no longer in use ? I was just wondering as I did not know it was possible to have a component that has no slot since it technically slots gux-avatar-focusable-beta.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+  private renderListItem(avatar: ProcessedAvatar): JSX.Element {
+    const handleClick = (event: MouseEvent) =>
+      this.handleListItemClick(event, avatar);
+    this.listItemClickListeners.set(avatar.focusableAvatar, handleClick);
+
+    return (
+      <gux-list-item key={avatar.name} onClick={handleClick}>
+        <div class="gux-overflow-list-item">
+          {this.renderAvatar(avatar, 'xsmall')}
```
gavin-everett-genesys, line 149: I think the avatars within the `gux-list` should not be focusable as there will be conflicts between the `gux-list-item` and it focusing on the `gux-avatar` element. I think just having the `gux-list-item` focus will be fine

gavin-everett-genesys: Also to add to the above the `gux-avatar` probably wont need the tooltip if in the `gux-list` but probably a UX question.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -0,0 +1,203 @@
...
+      const button = focusableAvatar.querySelector('button');
+      const link = focusableAvatar.querySelector('a');
+      const img = avatar.querySelector('img');
+      return {
+        focusableAvatar,
+        avatar,
+        button,
+        link,
+        img,
+        name: avatar.name,
```
gavin-everett-genesys, line 87: You will probably want optional chaining operators on these if the values are null eg `name : avatar?.name  || ''` or where u are doing the query selection.

packages/genesys-spark-components/src/components/stable/gux-list/gux-list.tsx
```diff
@@ -23,7 +23,7 @@ import { trackComponent } from '@utils/tracking/usage';
  * @slot - collection of gux-list-item, gux-list-divider elements
  */
 
-const validFocusableItems = ['gux-list-item'];
+const validFocusableItems = ['gux-list-item', 'gux-avatar-list-item-beta'];
```
gavin-everett-genesys, line 26: I was wondering what other peoples thoughts are on this as I need to do something similar in the `text-editor` component. I need to be able to customise the styles of the `gux-list-item` but I cannot do that due to the shadowDOM. If I were to create my own custom component I would add it here like `gux-rich-style-list-item` but then wouldn't that component be available for devs to use in there own `gux-lists` which I am not sure we if we want ? 🤔  Like with this change devs could use the `gux-list` and slot `gux-avatar-list-item-beta`. Maybe something for collab time.

321gillian: Would it be worth creating a separate gux-avatar-list?

gavin-everett-genesys: Yeah I was going to go down that route aswell but then you will have muiltiple `gux-*-list` components (even though they would be internal) with the same functionality. Maybe thats fine 🤔  or it might be possible to centralise the `gux-list` a bit. In my current branch I turned the `gux-list-item.scss` styles into a mixin and I reference it in my `gux-rich-style-list-item` style file as it would be duplicated scss essentially.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -0,0 +1,223 @@
...
+    if (invalidElements) {
+      logWarn(
+        this.root,
+        'gux-avatar-group-beta: Invalid child element detected. All child elements must be either buttons, anchor tags or gux-avatar-beta components'
+      );
+    }
+  }
+
+  private processAvatars(): void {
+    const avatarElements = Array.from(this.root.children) as HTMLElement[];
```
321gillian, line 104: I've created an alternate version of this which uses an avatarData prop set with JS instead of trying to work with the children of the avatar-group. `[feature/COMUI-3049_data-prop](https://github.com/MyPureCloud/genesys-spark/tree/feature/COMUI-3049_data-prop)`.

PR #647, opened 2024-10-08
fix(flyout-menu): added flip to sub-menu

daragh-king-genesys:
added flip to sub-menu

✅ Closes: COMUI-3099

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2731,18 +2731,18 @@
       }
     },
     "node_modules/@floating-ui/dom": {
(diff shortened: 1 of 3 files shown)


jordanstith15 (approved):

gavin-everett-genesys:
This works if you have a sub-menu as a child of the menu. But in the example docs it has a submenu nested within a submenu. The first submenu flips correctly but the second just overlays the parent menu.

It technically is correct since its not going outside the viewport but looks incorrect.

PR #649, opened 2024-10-08
fix(table-toolbar): fix console errors when slots are absent

gavin-everett-genesys:
**Note** 

The `setActionIconOnlyProp()` uses the spread operator to copy the array but the spread operator cannot accept `null` or `undefined` values which will return an error if the slot does not exist in the DOM. To alleviate this issue I just return an empty array. 

[✅ Closes: COMUI-3148](https://inindca.atlassian.net/browse/COMUI-3148)


packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.tsx
```diff
@@ -139,7 +140,7 @@ export class GuxTableToolbar {
   }
 
   get allFilterContextual(): HTMLGuxTableToolbarCustomActionElement[] | null {
```
jordanstith15, line 142: should null be removed from the type here?

packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.tsx
```diff
@@ -139,7 +140,7 @@ export class GuxTableToolbar {
...
-    return this.filterActions?.concat(this.contextualActions);
+    return this.filterActions?.concat(this.contextualActions) ?? [];
```
jordanstith15, line 143: would this possibly concat a null value making the array at least `[null]`? Is that caught by `??`?

gavin-everett-genesys: Updated to a more straightforward approach 👍

packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.tsx
```diff
@@ -138,8 +139,11 @@ export class GuxTableToolbar {
...
   }
 
-  get allFilterContextual(): HTMLGuxTableToolbarCustomActionElement[] | null {
-    return this.filterActions?.concat(this.contextualActions);
+  get allFilterContextual(): HTMLGuxTableToolbarCustomActionElement[] {
+    return [
+      ...(this.filterActions || []),
+      ...(this.contextualActions || [])
+    ].filter(action => action != null);
   }
```
jordanstith15, line 147: So I should have described this more, but would making the same changes as `get permanentActions` in `get contextualActions` and `get filterActions` be better? That way it removes the null typing entirely and no changes are needed in `get allFilterContextual`.

Also could turn them into a tertiary depending on how you feel.
i.e.

```
get permanentActions(): HTMLGuxTableToolbarCustomActionElement[] {
    return this.permanentSlot?.hasChildNodes ?
        Array.from(
             this.permanentSlot?.querySelectorAll(
                 'gux-table-toolbar-action, gux-table-toolbar-custom-action'
             )
       )
       : [];
}
```

gavin-everett-genesys: Yeah this makes sense to me 👍

jordanstith15 (approved):

PR #656, opened 2024-10-10
fix(gux-visualization): fix memory leak in gux-visualization

JackieChiles:
Fixes a memory leak in gux-visualization by calling the vega-embed finalize function to remove event listeners and other items.


JackieChiles:
@caitlanspronk here's the PR for the chart memory leak SERVOPS.

JackieChiles:
We'll need to follow up with this same change in the v4 chart components package, but starting with a patch to v3.

daragh-king-genesys (approved):

PR #660, opened 2024-10-16
fix(time-picker): Added validation input check (WIP)

jason-evans-genesys:
✅ Closes: COMUI-825

WIP

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.scss
@@ -1,5 +1,8 @@
 @use '~genesys-spark/dist/scss/mixins.scss';
 @use '~genesys-spark/dist/scss/focus.scss';
+@use '../gux-form-field/functional-components/gux-form-field-error/gux-form-field-error.scss';
+
+@include gux-form-field-error.Style;
 
 .gux-time-picker {
   position: relative;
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.scss
```diff
@@ -1,5 +1,8 @@
 @use '~genesys-spark/dist/scss/mixins.scss';
 @use '~genesys-spark/dist/scss/focus.scss';
+@use '../gux-form-field/functional-components/gux-form-field-error/gux-form-field-error.scss';
```
jason-evans-genesys, line 3: My intuition is that the time-picker isn't a form-field component so technically maybe we should move the error css in this file to a separate one that can be reused throughout the app, but let me know what you all think.

packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.tsx
```diff
@@ -78,6 +83,26 @@ export class GuxTimePicker {
   @Listen('blur')
   onBlur() {
     if (this.valueLastChange !== this.value) {
+      // Format input time to match format found in the popup time list (e.g. "01:30" -> "1:30", "00:30" -> "12:30", etc)
+      const split = this.value.split(':');
```
jason-evans-genesys, line 87: This logic here needs some reworking. I'm wondering if we need to handle different time formats from the user or will the user have to input with a specific time format all the time? For instance could the user input "01:30" and also "13:30" and also "1:30"?

packages/genesys-spark-components/src/components/stable/gux-time-picker/gux-time-picker.tsx
```diff
@@ -78,6 +83,26 @@ export class GuxTimePicker {
   @Listen('blur')
   onBlur() {
     if (this.valueLastChange !== this.value) {
+      // Format input time to match format found in the popup time list (e.g. "01:30" -> "1:30", "00:30" -> "12:30", etc)
```
jason-evans-genesys, line 86: Do you all think it makes sense to trigger input validation on blur or instead on input value change? I was worried triggering validation on input value change would be more intrusive than on blur.

PR #678, opened 2024-10-29
fix(tokens): fixed toggle border radii tokens

321gillian:


daragh-king-genesys (approved):
We should mention to UX to use the prefix `feature/` so that we get time machine branches to test.

PR #682, opened 2024-11-01
chore(other): fixed jest tests

daragh-king-genesys:
fixed jest tests


packages/genesys-spark-chart-components/package.json
```diff
@@ -33,8 +33,8 @@
     "stylelint": "stylelint --fix \"**/*.{css,html,scss}\"",
     "test": "TZ=UTC stencil test --spec --e2e",
     "test.ci": "TZ=UTC npm run test.ci.spec",
-    "test.ci.e2e": "TZ=UTC stencil test --e2e --max-workers=2",
-    "test.ci.spec": "TZ=UTC stencil test --spec --max-workers=2",
+    "test.ci.e2e": "TZ=UTC stencil test --e2e --maxWorkers=1",
+    "test.ci.spec": "TZ=UTC stencil test --spec --maxWorkers=1",
```
daragh-king-genesys, line 37: @321gillian @gavin-everett-genesys this PR is the same as the one you approved but I had to change `--runInBand` to `--maxWorkers=1`. In Jest they are the same but I think Stencil is not passing through the run in band flag correctly so I am using the max workers one instead.

gavin-everett-genesys (approved):

PR #685, opened 2024-11-04
docs(other): added report of unused ui tokens

daragh-king-genesys:
added report of unused ui tokens

✅ Closes: COMUI-2752


packages/genesys-spark-tokens/reports/unused-defined-ui-tokens.json
```diff
@@ -0,0 +1,555 @@
+[
+  "--gse-ui-accordion-contentItem-backgroundColor",
```
daragh-king-genesys, line 2: We can use this report to

- make sure our components are not missing tokens they should be using
- make suse UX has not left in tokens they are not using

katie-bobbe-genesys (approved):

PR #686, opened 2024-11-05
fix(tabs): check if tab-list is set on initialisation

gavin-everett-genesys:
**Note**

From what I could tell from the error and after inspecting the teams repository is that its an initialisation issue related to the `gux-tab-list`.  Sometimes the error will appear and other times it wont. As described in the ticket the tabs work as expected but its just the error that appears intermittently. This fix just adds a chaining operator to the`tabList` and `panel` to ensure no console error is emitted.

[✅ Closes: COMUI-3183](https://inindca.atlassian.net/browse/COMUI-3183)

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-tabs/gux-tabs.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-tabs/gux-tabs.tsx
@@ -103,8 +103,8 @@ export class GuxTabs {
       this.activeTab = panels[0].tabId;
     }
 
-    void tabList.guxSetActive(this.activeTab);
-    panels.forEach(
+    void tabList?.guxSetActive(this.activeTab);
+    panels?.forEach(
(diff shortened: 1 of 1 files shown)


packages/genesys-spark-components/src/components/stable/gux-tabs/gux-tabs.tsx
```diff
@@ -103,8 +103,8 @@ export class GuxTabs {
       this.activeTab = panels[0].tabId;
     }
 
-    void tabList.guxSetActive(this.activeTab);
-    panels.forEach(
+    void tabList?.guxSetActive(this.activeTab);
```
jason-evans-genesys, line 106: I notice `tabList` gets set in `onSlotchange`. Do we want to log a warning there if `tabListSlot.assignedElements()[0]` is not defined?

daragh-king-genesys (approved):

PR #687, opened 2024-11-05
fix(listbox): Added disabled prop

jason-evans-genesys:
✅ Closes: COMUI-3160


gavin-everett-genesys:
You may also need to check out how this will work with the dropdown and dropdown-multi for example should the dropdown be able to expand with a `disabled` `gux-listbox` etc. I think we may want to remove the example page also for the listbox and its not a design system component as discussed in previous calls.

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-listbox-multi.tsx
```diff
@@ -243,6 +246,13 @@ export class GuxListboxMulti {
     this.listboxOptions = (
       Array.from(this.root.children) as HTMLGuxOptionMultiElement[]
     ).filter(element => element.tagName === 'GUX-OPTION-MULTI');
+
+    if (this.disabled) {
+      this.listboxOptions.forEach(option => {
+        option.disabled = true;
+      });
```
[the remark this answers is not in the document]
jason-evans-genesys: Fixed

jason-evans-genesys:
> You may also need to check out how this will work with the dropdown and dropdown-multi for example should the dropdown be able to expand with a `disabled` `gux-listbox` etc. I think we may want to remove the example page also for the listbox and its not a design system component as discussed in previous calls.

@gavin-everett-genesys This has been addressed now. Right now the dropdowns do not expand if disabled. Thoughts on this?

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option/gux-option.scss
```diff
@@ -1,3 +1,13 @@
...
+
+:host {
+  &.gux-parent-disabled {
+    border-color: color-mix(
+      in srgb,
+      var(--gse-ui-button-disabled-opacity) 50%,
+      rgb(0 0 0 / 0%)
+    ); // TODO: which token should I use for opacity here? Ask UX team this.
+  }
+}
```
gavin-everett-genesys, line 13: Is this class being used ?

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option/gux-option.tsx
```diff
@@ -71,13 +73,22 @@ export class GuxOption {
...
   }
 
+  private hasDisabledParent(): boolean {
+    const parentListbox = getClosestElement(
+      'gux-listbox',
+      this.root
+    ) as HTMLGuxListboxElement;
+
+    return parentListbox?.disabled || false;
+  }
```
gavin-everett-genesys, line 83: Could this be  `return parentListbox?.disabled` as the `disabled` prop would be initially `false` anyways I think.

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
```diff
@@ -135,7 +146,7 @@ export class GuxOptionMulti {
         role="option"
         class={{
           'gux-active': this.active,
-          'gux-disabled': this.disabled,
+          'gux-disabled': this.disabled || this.hasDisabledParent(),
```
gavin-everett-genesys, line 149: This will probably also be needed in the `gux-option-icon` component.

gavin-everett-genesys:
Looks good to me 👍  I just had a few questions,

1. It does not seem to work in the dropdown context for example if I were to apply the `disabled` property to the listbox within a dropdown nothing seems to happen but if I use the listbox in a standalone context it works.
2. When the listbox is `disabled` it is still scrollable but I guess this is fine so users can view the contents of the list ? (maybe UX question).
3. Just wondering what would be the correct way for the dropdown to act for example if the listbox is disabled does that mean the entire dropdown should be disabled also ?

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-option-multi/gux-option-multi.tsx
jason-evans-genesys: Added

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option/gux-option.tsx
jason-evans-genesys: Fixed

jason-evans-genesys: Fixed

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option/gux-option.scss
jason-evans-genesys: Removed

jason-evans-genesys:
> Looks good to me 👍 I just had a few questions,
> 
>     1. It does not seem to work in the dropdown context for example if I were to apply the `disabled` property to the listbox within a dropdown nothing seems to happen but if I use the listbox in a standalone context it works.
> 
>     2. When the listbox is `disabled` it is still scrollable but I guess this is fine so users can view the contents of the list ? (maybe UX question).
> 
>     3. Just wondering what would be the correct way for the dropdown to act for example if the listbox is disabled does that mean the entire dropdown should be disabled also ?

@gavin-everett-genesys After our collab talk the team decided that if the listbox is wrapped in a parent component (e.g. dropdown) then the disabled prop will go on the parent component. In this case of adding the disabled prop to true on the dropdown then the listbox becomes disabled and the dropdown collapses.

I'll ask UX about item 2 that you had mentioned.

packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/example.html
```diff
@@ -366,10 +366,23 @@ <h2>Unknown Value Set</h2>
   >
 
   <h2>Disabled</h2>
-  <gux-dropdown-multi value="a" disabled>
+  <gux-button
+    id="toggle-disabled-state"
+    onclick="(function() {
+      const dropdown = document.getElementById('toggle-disabled-state-dropdown');
+      const disabled = dropdown.getAttribute('disabled') === 'true';
+      dropdown.setAttribute('disabled', !disabled);
```
daragh-king-genesys, line 374: This can be simplified to 
```
document.getElementById('toggle-disabled-state-dropdown').toggleAttribute('disabled');
```
https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-listbox-multi.tsx
```diff
@@ -70,6 +70,9 @@ export class GuxListboxMulti {
   @Prop()
   emptyMessage: string;
 
+  @Prop()
+  disabled: boolean = false;
```
daragh-king-genesys, line 74: add a comment that this is used by child elements

daragh-king-genesys (commented):
Both comments apply in two places.

packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/example.html
jason-evans-genesys: Fixed

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-listbox-multi.tsx
jason-evans-genesys: Fixed

daragh-king-genesys (approved):

PR #690, opened 2024-11-06
feat(segmented-control): added disabled state

daragh-king-genesys:
added disabled state

✅ Closes: COMUI-3009

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-segmented-control/example.html
+++ b/packages/genesys-spark-components/src/components/beta/gux-segmented-control/example.html
@@ -92,6 +92,33 @@
   </gux-segmented-control-beta>
 </div>
 
+<h2>Text</h2>
+
+<div>
+  <gux-segmented-control-beta
+    value="day"
+    disabled
+    onchange="notify(event)"
+    oninput="notify(event)"
+  >
+    <gux-segmented-control-item value="month">
+      <div slot="text">Month</div>
+    </gux-segmented-control-item>
+    <gux-segmented-control-item value="week">
+      <div slot="text">Week</div>
+    </gux-segmented-control-item>
+    <gux-segmented-control-item value="day">
+      <div slot="text">Day</div>
+    </gux-segmented-control-item>
+    <gux-segmented-control-item value="hour" disabled>
+      <div slot="text">Hour</div>
+    </gux-segmented-control-item>
+    <gux-segmented-control-item value="minute">
+      <div slot="text">Minute</div>
+    </gux-segmented-control-item>
+  </gux-segmented-control-beta>
+</div>
+
 <h2>Get value</h2>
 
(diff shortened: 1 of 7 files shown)


katie-bobbe-genesys:
Overall this looks good to me, but I would recommend changing the cursor style so that it is a cursor "arrow" instead of a "pointer" when hovering over the disabled segmented control items

packages/genesys-spark-components/src/components/beta/gux-segmented-control/example.html
```diff
@@ -92,6 +92,33 @@ <h2>Icon and Text</h2>
   </gux-segmented-control-beta>
 </div>
 
+<h2>Text</h2>
```
gavin-everett-genesys, line 95: Should this be `Disabled`.

packages/genesys-spark-components/src/components/beta/gux-segmented-control/gux-segmented-control-item/gux-segmented-control-item.scss
```diff
@@ -30,6 +30,14 @@ button {
     border-radius: var(--gse-ui-segmentedControl-borderRadius);
   }
 
+  &.gux-parent-disabled {
```
gavin-everett-genesys, line 32: Optional to add `user-select:none` to prevent selection of text.

packages/genesys-spark-components/src/components/beta/gux-segmented-control/gux-segmented-control-item/gux-segmented-control-item.tsx
```diff
@@ -113,7 +114,7 @@ export class GuxSegmentedControlItem {
             'gux-selected': this.selected
           }}
           type="button"
-          disabled={this.disabled}
+          disabled={this.disabled || this.hasDisabledParent()}
```
gavin-everett-genesys, line 117: I was just wondering if you just `disable` a segmented-control-item it will just disable the button and not the container so the border color looks like its not disabled. But from the designs it looks like they just disable the button element.

daragh-king-genesys: got the ok for the change from UX

gavin-everett-genesys: I should have caught this earlier but when `disabled` should the tooltip appear it currently does. Rest of changes look good to me 👍

daragh-king-genesys: I think it appearing is correct too.

gavin-everett-genesys (approved):

PR #701, opened 2024-11-12
chore(popover-list-beta): Added popover-list-beta component

jason-evans-genesys:
✅ Closes: COMUI-3184

There is a separate ticket to get the arrow to show: COMUI-3210

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-popover-list-beta/example.html
+++ b/packages/genesys-spark-components/src/components/beta/gux-popover-list-beta/example.html
@@ -1,0 +1,49 @@
+<h1>gux-popover-list</h1>
+<div class="scroll-container">
+  <div class="scroll-interior">
+    <gux-button id="popover-target">Example Element</gux-button>
+    <gux-popover-list-beta
+      id="popover-example"
+      position="top"
+      for="popover-target"
+      is-open
+    >
+      <gux-list>
+        <gux-list-item onclick="notify(event, 'item-1')">Item 1</gux-list-item>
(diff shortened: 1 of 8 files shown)


gavin-everett-genesys:
Changes look good to me 👍  , I spotted one edge case  see image below, 
 
Screenshot 2024-11-19 at 10 47 21

The `target` element still seems to display when the `for` element is gone from the screen or div. But it will dissapear after a few more pixels.  I was thinking its something related to the offset. If its not an easy change I am happy to approve anyways :)

jason-evans-genesys:
> Changes look good to me 👍 , I spotted one edge case see image below,
> 
> Screenshot 2024-11-19 at 10 47 21
> 
> The `target` element still seems to display when the `for` element is gone from the screen or div. But it will dissapear after a few more pixels. I was thinking its something related to the offset. If its not an easy change I am happy to approve anyways :)

@gavin-everett-genesys it looks like this issue is not present in firefox but it is in chrome. Good catch, I'll look for a fix.

jason-evans-genesys:
> > Changes look good to me 👍 , I spotted one edge case see image below,
> > Screenshot 2024-11-19 at 10 47 21
> > The `target` element still seems to display when the `for` element is gone from the screen or div. But it will dissapear after a few more pixels. I was thinking its something related to the offset. If its not an easy change I am happy to approve anyways :)
> 
> @gavin-everett-genesys it looks like this issue is not present in firefox but it is in chrome. Good catch, I'll look for a fix.

@gavin-everett-genesys Doesn't look like this is an easy change after all. There's another ticket to add the arrow so maybe this offset issue can be tackle in that ticket?

gavin-everett-genesys (approved):

PR #702, opened 2024-11-13
feat(avatar-group): Add avatar grouping

321gillian:
This PR just handles the grouping of visible avatars, handling the hidden avatars will following the next PR. 

✅ Closes: COMUI-3049


packages/genesys-spark-components/src/components/beta/gux-avatar-group/example.html
```diff
@@ -0,0 +1,71 @@
...
+    ><img slot="image" src="https://i.pravatar.cc/300" alt="Alan Butler"
+  /></gux-avatar-beta>
+  <gux-avatar-beta name="Conor Darcy"></gux-avatar-beta>
+  <gux-avatar-beta name="Elliot Fitzgerald"></gux-avatar-beta>
+  <gux-avatar-beta name="Greg Hayes">
+    <img slot="image" src="https://i.pravatar.cc/300" alt="Greg Hayes"
+  /></gux-avatar-beta>
+  <gux-avatar-beta name="Iseult Jones"></gux-avatar-beta>
+  <gux-avatar-beta name="John King"></gux-avatar-beta>
+</gux-avatar-group-beta>
```
321gillian, line 15: There could be a use case for focus at a group level similar to this [SAP Avatar Groups](https://experience.sap.com/fiori-design-web/avatar-group-web-component/#anatomy1) but I don't think it needs to go in with this.

gavin-everett-genesys:
time machine branch probably needs to be rebuilt on jenkins as its returning 403

321gillian:
> https://apps.inindca.com/common-ui-docs/genesys-webcomponents/feature/COMUI-3049_v3

thanks for heads-up, it should be working now

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -0,0 +1,118 @@
...
+      this.interactiveElements.slice(1).forEach(element => {
+        element.tabIndex = -1;
+      });
+    }
+
+    this.avatars = Array.from(
+      this.root.querySelectorAll('gux-avatar-beta')
+    ) as HTMLGuxAvatarBetaElement[];
+
+    if (!this.avatars.length) {
```
gavin-everett-genesys, line 64: optional `if (this.avatars.length === 0)`

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -0,0 +1,118 @@
...
+      logWarn(this.root, 'No avatars found in slot');
+    }
+
+    // This is required for the aria-label
+    this.count = this.avatars.length;
+
+    this.avatars.forEach((avatar: HTMLGuxAvatarBetaElement, index: number) => {
+      // Only small avatars can be included in avatar-group
+      avatar.size = 'small';
```
gavin-everett-genesys, line 73: I am not sure if this is something to worry about, but this can be bypassed if devs set the size after the component loads.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  private getAccent(index): string {
+    const totalAvatarAccentVariations = 12;
+    return `${(index % totalAvatarAccentVariations) + 1}` as GuxAvatarAccent;
+  }
+
+  render(): JSX.Element {
+    return (
+      <div
+        class="gux-avatar-group"
```
gavin-everett-genesys, line 111: Should this have a `role` of type `list` also ?

packages/genesys-spark-components/src/components/beta/gux-avatar-group/i18n/en.json
```diff
@@ -0,0 +1,3 @@
+{
+  "avatarGroup": "avatar group containing {count} items. Use arrows to navigate."
```
gavin-everett-genesys, line 2: Minor to specify just `left` and `right` arrow keys.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
321gillian: I guess they can bypass it but it'll look wrong so it might be ok I think

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
321gillian: I think in this case role `group` with custom directions might be preferable to role `list`. 

There are use cases for an avatar group item that is either just an avatar, an avatar wrapped in a button or an avatar wrapped with a link. 


If I was to use a list item like gux-list-item which needs an interactive element within the `Host`
`
      <Host role="listitem">
        <button type="button" tabIndex={-1} disabled={this.disabled}>
          <slot></slot>
        </button>
      </Host>
`


I can't just slot a button or link in there and have it work the same way. It's important that the original structure like this:
`    <a href="www.google.com" target="_blank">
      <gux-avatar-beta name="Alan Butler">
        Alan Butler
      </gux-avatar-beta>
    </a>
    `
   is maintained otherwise the link will be read out as a button by a screenreader
   
 Kind of faking it by creating an interactive element inside the list item component to match is not a good way of doing it either. 
 
 So as the least version of evil, I think that using the group and declaring the navigation explicitly might be the most straightforward. Any other thoughts on how to work around this?

conor-darcy: Group seems fine to me. It doesn't really feel like a list...probably why its called 'avatar-group' & not 'avatar-list'.

gavin-everett-genesys: Looking at it again , I would go with `group` 👍

PR #709, opened 2024-11-18
feat(form-footer): Remove modal-desktop variant and tokens

321gillian:
COMUI-3223


packages/genesys-spark-components/src/components/beta/gux-form/gux-form-footer/gux-form-footer.scss
```diff
@@ -36,12 +36,3 @@
...
 }
-
-:host(.gux-form-footer-modal-desktop) ::slotted(footer) {
-  gap: var(--gse-ui-formFooter-sideSheet-desktop-gap);
-  justify-content: flex-end;
-  padding-block: var(--gse-ui-formFooter-modal-desktop-topPadding)
-    var(--gse-ui-formFooter-modal-desktop-bottomPadding);
-  padding-inline: var(--gse-ui-formFooter-modal-desktop-horizontalPadding)
-    var(--gse-ui-formFooter-modal-desktop-horizontalPadding);
-}
```
gavin-everett-genesys, line 47: I guess this is not a breaking change is it ? Like if a team is using this variant their layout will break. Do you know what is reason for the removal was ?

321gillian: I've sent the question to design but I was thinking since the form is in beta anyway it would be ok to remove it.

daragh-king-genesys: Ya it is ok to remove because it is in beta. What I usually do is search bitbucket to see if I can find any usage of it. I would guess no one is using this but if they are just reach out to them and let them know UX have removed it from the DS so we are removing it from the beta component

daragh-king-genesys (approved):

packages/genesys-spark-components/src/components/beta/gux-form/gux-form-footer/gux-form-footer.scss
321gillian: How would you go about searching all of bitbucket?

PR #715, opened 2024-11-19
fix(table-toolbar): hide tooltip when menu is expanded

gavin-everett-genesys:
I think we had this issue with another component aswell but I cannot recall if anyone knows maybe it should be looked at as part of the tooltip implementation maybe if possible ?

https://inindca.atlassian.net/browse/COMUI-3212


packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar-menu-button/gux-table-toolbar-menu-button.scss
```diff
@@ -11,16 +11,15 @@
...
+  border: none;
 
-    gux-icon {
-      width: var(--gse-ui-icon-small-size);
-      height: var(--gse-ui-icon-small-size);
-    }
+  gux-icon {
+    align-self: center;
+    width: var(--gse-ui-icon-small-size);
+    height: var(--gse-ui-icon-small-size);
```
321gillian, line 22:
```suggestion
    block-size: var(--gse-ui-icon-small-size);
    inline-size: var(--gse-ui-icon-small-size);
```

321gillian:
> I think we had this issue with another component aswell but I cannot recall if anyone knows maybe it should be looked at as part of the tooltip implementation maybe if possible ?
> 
> https://inindca.atlassian.net/browse/COMUI-3212

I know we were talking about this in relation to the avatar group. It seems like maybe the tooltip should have some sort of way to detect if more tooltips are showing. Not sure but a ticket for looking into it might be good

321gillian (approved):

packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar-menu-button/gux-table-toolbar-menu-button.scss
daragh-king-genesys: linter should take care of this if you rebase on main

PR #725, opened 2024-11-22
fix(table-toolbar): always show label when in the condensed display

daragh-king-genesys:
always show label when in the condensed display

✅ Closes: COMUI-3213

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.tsx
@@ -259,7 +259,11 @@ export class GuxTableToolbar {
 
   render(): JSX.Element {
     return (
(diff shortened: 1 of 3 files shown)


packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.tsx
```diff
@@ -259,7 +259,11 @@ export class GuxTableToolbar {
 
   render(): JSX.Element {
     return (
-      <Host role="toolbar" aria-orientation="horizontal">
+      <Host
+        role="toolbar"
+        aria-orientation="horizontal"
+        gs-layout={this.displayedLayout}
```
gavin-everett-genesys, line 265: Out of the two solutions I would +1 for this as its a more encapsulated. Would there be any downside to exposing this `gs-layout` attribute though ?

jordanstith15 (approved):

PR #729, opened 2024-11-25
fix(pagination-cursor): Fixed alignment of icon

jason-evans-genesys:
✅ Closes: COMUI-3236


jordanstith15 (changes requested):
Code looks fine, but build is failing in the snapshot tests.

packages/genesys-spark-components/src/components/stable/gux-pagination-cursor/gux-pagination-cursor.scss
```diff
@@ -45,6 +45,10 @@
     }
   }
 
+  .gux-simple-button-content {
+    display: flex;
+  }
+
```
daragh-king-genesys, line 51: I don't see a big advantage with adding a wrapper element with a class instead of just adding the needed css into the `gux-icon` block

```
  gux-icon {
    display: flex;
    inline-size: var(--gse-ui-button-icon-size);
    block-size: var(--gse-ui-button-icon-size);
  }
```

jason-evans-genesys: fixed

jordanstith15 (approved):

PR #733, opened 2024-11-26
feat(tokens): Dark Mode Pt2

VirtualClover:
Adjusted dark mode tokens based on review feedback. 
Haven't generated new CSS vars, since we'll do that after the review is finalized.

> [!NOTE]  
> We moved box shadows from `semantic/global` to `semantic/mode/light` and `semantic/mode/dark` to allow us to have different shadows for each mode, this won't affect CSS vars since they will still be generated with the same name


packages/genesys-spark-tokens/data/semantic/mode/dark.json
```diff
@@ -490,6 +490,20 @@
           "type": "color",
           "description": "The default background color for inputs."
         }
+      },
+      "search": {
+        "matching_text": {
+          "first_level": {
+            "value": "{theme.background.custom_accents.bold.120}",
```
gavin-everett-genesys, line 497: should these tokens have a `d` prefix at the end ?

packages/genesys-spark-tokens/data/semantic/theme/flare.json
```diff
@@ -59,15 +59,15 @@
             "type": "color"
           },
           "40d": {
-            "value": "{color.azure_blue.1000}",
+            "value": "{color.haze.850}",
```
gavin-everett-genesys, line 62: was this color meant to change ?

conor-darcy:
@VirtualClover I've re-generated the tokens, can you check if everything is as expected?

packages/genesys-spark-tokens/data/semantic/mode/dark.json
VirtualClover: Not in this case, since the search match color is the same in light and dark

packages/genesys-spark-tokens/data/semantic/theme/flare.json
VirtualClover: Yes, we change the color of the secondary button from blue to gray-ish in dark mode

VirtualClover:
> @VirtualClover I've re-generated the tokens, can you check if everything is as expected?

Yes! LGTM :)

conor-darcy (approved):

PR #734, opened 2024-11-27
[WIP] fix(table): z-index issues with gux-truncate

jordanstith15:
Closes [COMUI-3094](https://inindca.atlassian.net/browse/COMUI-3094)

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi-tag/readme.md
+++ b/packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi-tag/readme.md
@@ -7,10 +7,10 @@
 
 ## Properties
 
-| Property         | Attribute         | Description       | Type      | Default |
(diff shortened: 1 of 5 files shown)


jordanstith15:
The fix seems to be to remove a z-index that was added to fix another issue before. ([COMUI-2951](https://inindca.atlassian.net/browse/COMUI-2951))
Removing it does not seem to reintroduce the older issue in my local in either FF or Chrome, but will test in a PR demo first before making it ready to review

jordanstith15:
> The fix seems to be to remove a z-index that was added to fix another issue before. ([COMUI-2951](https://inindca.atlassian.net/browse/COMUI-2951)) Removing it does not seem to reintroduce the older issue in my local in either FF or Chrome, but will test in a PR demo first before making it ready to review

Tested in demo in FF, Chrome, and Safari. Everything looks good and the old issue hasn't resurfaced.

daragh-king-genesys (approved):

PR #737, opened 2024-11-29
fix(button-multi): fixed accessibility warning

daragh-king-genesys:
fixed accessibility warning

✅ Closes: COMUI-3241


321gillian (approved):
There's another very similar issue here - https://inindca.atlassian.net/browse/COMUI-3197

PR #744, opened 2024-12-02
fix(popover-list-beta): Focus first list item on popover show

jason-evans-genesys:
✅ Closes: COMUI-1576

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-popover-list-beta/gux-popover-list.tsx
+++ b/packages/genesys-spark-components/src/components/beta/gux-popover-list-beta/gux-popover-list.tsx
@@ -18,6 +18,7 @@ import {
   shift,
   hide
 } from '@floating-ui/dom';
+import { afterNextRenderTimeout } from '@utils/dom/after-next-render';
 
 import { OnClickOutside } from '@utils/decorator/on-click-outside';
 import { trackComponent } from '@utils/tracking/usage';
@@ -76,6 +77,8 @@ export class GuxPopoverList {
   @Event()
   guxdismiss: EventEmitter<void>;
 
+  private listElement: HTMLGuxListElement;
+
   @Listen('keydown')
   onKeyDown(event: KeyboardEvent): void {
     switch (event.key) {
@@ -98,12 +101,6 @@ export class GuxPopoverList {
       !clickedForElement
     ) {
       this.dismiss();
-    }
-
-    if (clickedForElement) {
-      this.popupElement.togglePopover();
-      this.isOpen = !this.isOpen;
-      this.runUpdatePosition();
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/genesys-spark-components/src/components/beta/gux-popover-list-beta/gux-popover-list.tsx
```diff
@@ -192,14 +199,25 @@ export class GuxPopoverList {
...
+      void this.listElement.guxFocusFirstItem();
+    });
+  }
+
   connectedCallback(): void {
     trackComponent(this.root, { variant: this.position });
   }
 
+  componentWillLoad(): void {
+    this.listElement = this.root.querySelector('gux-list');
```
gavin-everett-genesys, line 213: This can be called in `connectedCallback` I think.

gavin-everett-genesys:
I think for focus rings they should only be shown when navigating via keyboard and not through mouse click. So if i click the target element via mouse no focus ring should appear.

packages/genesys-spark-components/src/components/beta/gux-popover-list-beta/gux-popover-list.tsx
jason-evans-genesys: fixed

jason-evans-genesys:
> I think for focus rings they should only be shown when navigating via keyboard and not through mouse click. So if i click the target element via mouse no focus ring should appear.

@gavin-everett-genesys good call. Ok fixes pushed.

gavin-everett-genesys:
> > I think for focus rings they should only be shown when navigating via keyboard and not through mouse click. So if i click the target element via mouse no focus ring should appear.
> 
> @gavin-everett-genesys good call. Ok fixes pushed.

Will approve I was just wondering did you push these changes ?

jason-evans-genesys:
@daragh-king-genesys this PR is ready for review again after implementing the target element click changes we discussed in a recent standup.

jason-evans-genesys:
> > > I think for focus rings they should only be shown when navigating via keyboard and not through mouse click. So if i click the target element via mouse no focus ring should appear.
> > 
> > 
> > @gavin-everett-genesys good call. Ok fixes pushed.
> 
> Will approve I was just wondering did you push these changes ?

@gavin-everett-genesys I made these changes now.

321gillian:
> > > > I think for focus rings they should only be shown when navigating via keyboard and not through mouse click. So if i click the target element via mouse no focus ring should appear.
> > > 
> > > 
> > > @gavin-everett-genesys good call. Ok fixes pushed.
> > 
> > 
> > Will approve I was just wondering did you push these changes ?
> 
> @gavin-everett-genesys I made these changes now.

Can you check this again? The first item seems to still be getting focus on click for the timemachine branch

katie-bobbe-genesys:
1. I would recommend checking the behavior in Chrome as well as Firefox. The first item is getting focus on click for the time machine in Chrome (but not Firefox). We will not want the focus ring when using the mouse

2. when the popover list has an initial open position, the focus ring is showing on the first element, which we don't want

jason-evans-genesys:
@katie-bobbe-genesys @gavin-everett-genesys @321gillian Pushed some fixes to address the issues you all found.

katie-bobbe-genesys (approved):

PR #746, opened 2024-12-03
feat: override CDN URL via assetsUrl option to registerSparkComponents

jasonmobley:
Add new `assetsUrl` field to registerOptions to allow overriding the CDN URL when calling registerSparkComponents().

This makes it easier to test a local build of Spark components in an app without making major hacks to how Spark components are registered.

https://inindca.atlassian.net/browse/COMUI-3244

### DIFF
--- a/packages/genesys-spark-components/stencil.config.ts
+++ b/packages/genesys-spark-components/stencil.config.ts
@@ -1,4 +1,5 @@
 import { Config } from '@stencil/core';
+import { Credentials } from '@stencil/core/internal';
 import { sass } from '@stencil/sass';
 import copy from 'rollup-plugin-copy';
 import generateMetadata from './scripts/generate-component-data';
@@ -8,6 +9,16 @@ const testConsoleReporter =
   process.env.DEFAULT_TEST_REPORTER === 'true'
     ? 'default'
     : ['jest-silent-reporter', { useDots: true }];
+
+// Optionally host the dev server with https if a cert and key are provided via env variables, e.g.
+// for running local dev build within a https app using assetsUrl option of registerSparkComponents.
+let https: Credentials | undefined = undefined;
+if (process.env.SPARK_HTTPS_CERT && process.env.SPARK_HTTPS_KEY) {
+  https = {
+    key: process.env.SPARK_HTTPS_KEY,
+    cert: process.env.SPARK_HTTPS_CERT
+  };
+}
 
 export const config: Config = {
   namespace: 'genesys-webcomponents',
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/stencil.config.ts
```diff
@@ -9,6 +10,16 @@ const testConsoleReporter =
...
 
+// Optionally host the dev server with https if a cert and key are provided via env variables, e.g.
+// for running local dev build within a https app using assetsUrl option of registerSparkComponents.
+let https: Credentials | undefined = undefined;
+if (process.env.SPARK_HTTPS_CERT && process.env.SPARK_HTTPS_KEY) {
+  https = {
+    key: process.env.SPARK_HTTPS_KEY,
+    cert: process.env.SPARK_HTTPS_CERT
+  };
+}
```
jasonmobley, line 21: This part is not strictly necessary, but it goes hand-in-hand. If my localhost app uses https, then HSTS will prevent the browser from loading assets from a non-https localhost dev server. My preference would be to add a dependency on @genesys/localcert but that package is in our private repo so the next best thing is giving myself a way to run with localcert sans dependency.

Now I can do this to get a localhost dev server:

```shell
> SPARK_HTTPS_CERT=$(npx -y @genesys/localcert produce certificate -c) SPARK_HTTPS_KEY=$(npx -y @genesys/localcert produce certificate -k) npm run dev.gsc
```

packages/genesys-spark-components/stencil.config.ts
```diff
@@ -9,6 +10,16 @@ const testConsoleReporter =
...
+
```
jasonmobley, line 22: This part is not strictly necessary, but it goes hand-in-hand. If my localhost app uses https, then HSTS will prevent the browser from loading assets from a non-https localhost dev server. My preference would be to add a dependency on @genesys/localcert but that package is in our private repo so the next best thing is giving myself a way to run with localcert sans dependency.

Now I can do this to get a localhost dev server:

```shell
> SPARK_HTTPS_CERT=$(npx -y @genesys/localcert produce certificate -c) SPARK_HTTPS_KEY=$(npx -y @genesys/localcert produce certificate -k) npm run dev.gsc
```

daragh-king-genesys:
@jasonmobley can you duplicate this into the charts package too and then we can merge.

jasonmobley:
> @jasonmobley can you duplicate this into the charts package too and then we can merge.

Done in 126aa55

daragh-king-genesys (approved):

PR #752, opened 2024-12-06
feat(link): adding beta component for hyperlinks

jordanstith15:


packages/genesys-spark-components/src/components/beta/gux-link/example.html
```diff
@@ -0,0 +1,35 @@
+<h2>Standalone Link</h2>
+<gux-link-beta standalone>
+  <a href="#" target="_top">Link</a>
+</gux-link-beta>
```
gavin-everett-genesys, line 4: From the ticket it mentions that a standalone link always has an icon at the end or has this changed ? I did read that the breadcrumbs do use the the `standalone` without the icon at the end but just wanted to confirm.

packages/genesys-spark-components/src/components/beta/gux-link/gux-link.scss
```diff
@@ -0,0 +1,112 @@
...
+
+  ::slotted(a:active) {
+    text-decoration: none !important;
+  }
+}
+
+:host([size='small'][standalone]) {
+  ::slotted(a) {
+    font-family: var(--gse-ui-links-standalone-small-text-fontFamily),
+      var(--gse-semantic-theme-fontFamily-body), sans-serif;
```
gavin-everett-genesys, line 92: does sans-serif need a token.

packages/genesys-spark-components/src/components/beta/gux-link/gux-link.tsx
```diff
@@ -0,0 +1,39 @@
...
+})
+export class GuxLink {
+  @Element()
+  private root: HTMLElement;
+
+  @Prop()
+  size: 'medium' | 'small' = 'medium';
+
+  @Prop()
+  standalone: boolean = false;
```
gavin-everett-genesys, line 18: this seems fine to be my only concern is if the add other types of links in the future potentially like external links or something. maybe a type would be better.

packages/genesys-spark-components/src/components/stable/gux-toast/example.html
```diff
@@ -19,7 +21,9 @@ <h2>Gux Success Toast</h2>
 
 <gux-toast toast-type="success" aria-live="polite">
   <div slot="message">Success toast with Link without Title</div>
-  <a slot="link" href="#">Link</a>
+  <a slot="link" href="#"
+    >Link<gux-icon icon-name="external-link" decorative></gux-icon
+  ></a>
 </gux-toast>
 
 <h2>Gux Warning Toast</h2>
```
gavin-everett-genesys, line 29: There is one toast in here that has a Link without and icon

packages/genesys-spark-components/src/components/beta/gux-link/example.html
jordanstith15: It does not necessarily, but highly recommended iirc. I clarified it with UX. They were fine with the solution of slotting an icon. It was also the best option I could think of with edge cases like breadcrumbs.

I guess it would make sense to remove the non-icon example though

packages/genesys-spark-components/src/components/beta/gux-link/gux-link.scss
jordanstith15: iirc sans-serif is the fallback if the two other tokens fail to resolve for some reason. The second token is already a fallback. I took it from the global classes we already have and just changed the first token

packages/genesys-spark-components/src/components/beta/gux-link/gux-link.tsx
jordanstith15: I think standalone covers external links (by adding the external link icon to it). I thought to do it that way too, but couldn't think of any other types that could possibly exist style-wise. Everything else is controlled by the user through the slotted <a>.

packages/genesys-spark-components/src/components/stable/gux-toast/example.html
jordanstith15: oops. that's a miss on my part

packages/genesys-spark-components/src/components/beta/gux-link/gux-link.scss
```diff
@@ -0,0 +1,112 @@
...
+  ::slotted(a:active) {
+    color: var(--gse-ui-links-active-foregroundColor) !important;
+  }
+
+  ::slotted(a:visited) {
+    color: var(--gse-ui-links-visited-foregroundColor) !important;
+  }
+
+  ::slotted(a:focus-visible) {
+    @include focus.gux-focus-ring;
```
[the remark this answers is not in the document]
jordanstith15: So I do have a border-radius set above. The token just isn't updated to the newest value yet. It will be fixed after this merges per UX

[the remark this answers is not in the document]
jordanstith15: aaannnddd the token was pushed to my branch anyways.. should be good now lol

packages/genesys-spark-components/src/components/stable/gux-toast/example.html
gavin-everett-genesys: I think this just needs to be updated I did a re-review all looks good will approve after this 👍

jordanstith15: Updated

gavin-everett-genesys (approved):

PR #806, opened 2024-12-23
feat(avatar-groups): Add overflow

321gillian:
I think there's enough in this as it is so I've added another follow-on task for the next part: https://inindca.atlassian.net/browse/COMUI-3367


packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group-item/gux-avatar-group-item.tsx
```diff
@@ -104,32 +105,33 @@ export class GuxAvatarGroupItem {
...
-          <span class="gux-avatar-initials" aria-hidden="true">
-            {generateInitials(this.name)}
-          </span>
-        </slot>
-        <gux-tooltip-beta
-          aria-hidden="true"
-          visual-only
-          placement="top"
-          ref={el => (this.tooltip = el)}
+      <Host role="menuitem">
```
321gillian, line 108: This role needs to be in the light dom, otherwise the menu won't announce the number of items in the menu

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.service.ts
```diff
@@ -51,30 +54,37 @@ export function setFocusTarget(element: Element): void {
...
+    void setItemTabIndex(lastFocusableElement, 0);
   }
 }
 
 function focusPreviousSiblingLoop(currentElement: Element): void {
-  const previousFocusableElement =
-    currentElement.previousElementSibling as HTMLGuxAvatarGroupItemBetaElement;
+  const groupItems = getGroupItems(currentElement);
+  const currentElementIndex = groupItems.findIndex(
+    (item: Element) => item === currentElement
```
jason-evans-genesys, line 81: I think doing a comparison on an object property (e.g. ID or something) vs the whole object might be safer

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.service.ts
```diff
@@ -87,58 +97,66 @@ function focusPreviousSiblingLoop(currentElement: Element): void {
...
+  return slottedItems as GuxAvatarGroupChild[];
 }
 
 function setItemTabIndex(element: Element, newIndex: number) {
   const button = getGroupItemButton(element);
   if (button) {
     button.tabIndex = newIndex;
   } else {
-    console.log('No button found in the gux-avatar-group-item element');
+    console.log('No button found in the element');
```
jason-evans-genesys, line 157: instead of `log` does `warning` make more sense?

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -15,43 +27,75 @@ import {
   shadow: { delegatesFocus: true }
 })
 export class GuxAvatarGroup {
+  private processedGroupItems: ProcessedItem[] = [];
+
   @Element()
   root: HTMLElement;
 
+  @Prop()
+  quantity: GuxAvatarGroupQuantity = 7;
```
jason-evans-genesys, line 36: Just thinking that if something changes in `GuxAvatarGroupQuantity` in `gux-avatar-group.types.ts` then we'd potentially have to change this value here, like if the max quantity is 8 or something. Is there a way to get this default number from referencing `gux-avatar-group.types.ts` somehow?

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-overflow/gux-avatar-overflow-item/gux-avatar-overflow-item.scss
```diff
@@ -0,0 +1,139 @@
...
+      color: var(--gse-ui-avatar-media-initialsForeground-default);
+      background-color: var(--gse-ui-avatar-media-initialsBackground-accent8);
+    }
+
+    &.gux-accent-9 {
+      color: var(--gse-ui-avatar-media-initialsForeground-default);
+      background-color: var(--gse-ui-avatar-media-initialsBackground-accent9);
+    }
+
+    &.gux-accent-10 {
```
jason-evans-genesys, line 119: is it possible to do a for loop here to dynamically render all 12 with only a few lines of code? Not a big deal if we keep it this way though.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-overflow/gux-avatar-overflow.service.ts
```diff
@@ -0,0 +1,108 @@
...
+      focusNextSiblingLoop(currentElement);
+      break;
+
+    case 'ArrowRight':
+      event.preventDefault();
+      event.stopPropagation();
+
+      break;
+
+    case 'ArrowLeft':
```
jason-evans-genesys, line 26: I think we could combine some of these like:

`case 'ArrowLeft':
case 'ArrowRight':
event.preventDefault();
      event.stopPropagation();
      break;
`

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-overflow/gux-avatar-overflow-item/gux-avatar-overflow-item.tsx
```diff
@@ -0,0 +1,111 @@
...
+  name: string;
+
+  /**
+   * Manually sets avatar accent
+   */
+  @Prop()
+  accent: GuxAvatarAccent = 'auto';
+
+  async componentWillLoad(): Promise<void> {
+    trackComponent(this.root);
```
gavin-everett-genesys, line 42: can track `variant` here.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-overflow/gux-avatar-overflow-item/gux-avatar-overflow-item.tsx
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  private getAccent(): string {
+    if (this.accent !== 'auto') {
+      return `gux-accent-${this.accent}`;
+    }
+    const hashedName = this.name
+      ?.split('')
+      .reduce((acc, char) => acc + char.charCodeAt(0), 0);
+    const hashedNameAccent = (hashedName % 12).toString();
```
gavin-everett-genesys, line 70: Was wondering if its possible to get the number of accents dynamically instead of hardcoding  `12`  here just incase the number of accent do change.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -64,14 +108,72 @@ export class GuxAvatarGroup {
...
-    resetFocusableSibling(clickedElement);
-    setFocusTarget(clickedElement);
+
+    if (
+      clickedElement.tagName === 'GUX-AVATAR-GROUP-ITEM-BETA' &&
+      !clickedElement.classList.contains('gux-hidden')
+    ) {
+      resetFocusableSibling(clickedElement);
+      setFocusTarget(clickedElement);
+    }
```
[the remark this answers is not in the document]
321gillian: Basically, there's a convention in menus and avatar groups that I've seen where the last item visited by keyboard or clicked is set at the current focus target. So if you tab back up from below then you return to the last item clicked / the one you tabbed away from.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-overflow/gux-avatar-overflow-item/gux-avatar-overflow-item.scss
321gillian: The color of the fonts for the accents vary so it wouldn't be a straightforward loop. I think it's simpler to do it like this.

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
321gillian: I had a go at doing this using an enum in the types. Let me know what you think

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-overflow/gux-avatar-overflow-item/gux-avatar-overflow-item.tsx
321gillian: I did a bit of refactoring here to make this function a service in gux-avatar. I added a constant there for the auto accents and then used that to build the type. Not sure if this is the right approach so any other feedback is welcome

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-overflow/gux-avatar-overflow-item/gux-avatar-overflow-item.scss
jason-evans-genesys: Sounds good

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
jason-evans-genesys: Looks good

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -15,43 +26,77 @@ import {
...
+    });
+  }
+
+  private hideOverflowGroupItems(): void {
+    const overflowItems = this.processedGroupItems.filter(
+      item => item.isOverflow
+    );
+
+    overflowItems.forEach(item => {
+      item.groupItem.setAttribute('hidden', '');
```
321gillian, line 93: @daragh-king-genesys Replace the class with 'hidden' attribute. [MDN Hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden)

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group-item/gux-avatar-group-item.tsx
```diff
@@ -50,15 +51,6 @@ export class GuxAvatarGroupItem {
...
 
-  /*
-   * Focus button element
-   */
-  @Method()
-  // eslint-disable-next-line @typescript-eslint/require-await
-  async guxFocus(): Promise<void> {
-    this.buttonElement.focus();
-  }
-
```
321gillian, line 61: Removed `guxFocus` and used focus delegation instead

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-overflow/gux-avatar-overflow.scss
```diff
@@ -0,0 +1,78 @@
...
+.gux-menu-wrapper {
+  position: fixed;
+  inset-block-start: 0;
+  inset-inline-start: 0;
+  z-index: var(--gse-semantic-zIndex-popup);
+  flex-direction: column;
+  max-block-size: var(--gse-ui-menu-maxHeight);
+  padding: var(--gse-ui-menu-padding);
+  margin: 0;
+  overflow-y: scroll;
```
daragh-king-genesys, line 64: I think this is giving the menu padding to take a srollbar even when it is not needed. Maybe `auto` would be better

daragh-king-genesys (commented):
LGTM

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -1,57 +1,102 @@
...
+  shadow: true
 })
 export class GuxAvatarGroup {
+  private processedGroupItems: ProcessedGroupItem[] = [];
+
   @Element()
   root: HTMLElement;
 
+  @Prop()
+  quantity: GuxAvatarGroupQuantity = Math.max(
```
jordanstith15, line 35: I think this is missing the type declaration? I see we set the type of the default value, but not the variable itself iirc. Meaning that the number isn't limited, which leads to my next questions. Are we limiting the quantity to a max of 7 to start and that's why we have the number array stuff?

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
```diff
@@ -1,57 +1,102 @@
...
+    ...GROUP_QUANTITIES
+  ) as GuxAvatarGroupQuantity;
```
jordanstith15, line 37: So we are purposely limiting the max quantity to the max of the number array? Is that why it's done this way?

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-overflow/gux-avatar-overflow.scss
321gillian: Done

packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group.tsx
321gillian: I checked this with Monse and we don't need to limit it to a max so I've updated it to remove the restriction.

jordanstith15 (approved):
LGTM

PR #811, opened 2025-01-06
chore(a11y): adding check based on non-text content

conor-darcy:
Adding new check as I've noticed with icons that some may need tooltips attached.

### DIFF
--- a/packages/genesys-spark-components/documentation/A11Y_TESTING.md
+++ b/packages/genesys-spark-components/documentation/A11Y_TESTING.md
@@ -70,7 +70,27 @@ To check the status of the manual accessibility testing in the project, run `npm
 
 The following criteria should be met to pass the manual accessibility checks
 
-1. [2.1.1 Keyboard](https://www.w3.org/WAI/WCAG22/Understanding/keyboard):
+1. [1.1.1 Non-text Content](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content)
+
+- **WCAG criterion summary:**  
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/documentation/A11Y_TESTING.md
```diff
@@ -70,103 +70,123 @@ To check the status of the manual accessibility testing in the project, run `npm
 
 The following criteria should be met to pass the manual accessibility checks
 
-1. [2.1.1 Keyboard](https://www.w3.org/WAI/WCAG22/Understanding/keyboard):
+1. [1.1.1 Non-text Content](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content)
```
katie-bobbe-genesys, line 73: I am fine with adding this but I don't think we need to redo our manual accessibility checklist checks. I think the reason I didn't add this initially is because it's something that can be commonly picked up by the automated testing and overlaps with  [2.5.3 Label in Name](https://www.w3.org/WAI/WCAG22/Understanding/label-in-name)

   - **WCAG summary:**
     "For user interface components with labels that include text or images of text, the name contains the text that is presented visually"

   - **How to check:**
     Everything that is presented visually should also be presented to the screen reader. Icons and images should have screen reader accessible text. Make sure that what is read out by the screen reader matches what is shown visually.

conor-darcy: @katie-bobbe-genesys 100%, thanks.

katie-bobbe-genesys (approved):

PR #813, opened 2025-01-07

daragh-king-genesys:
update dependencies to wanted


daragh-king-genesys:
This was the first dependency update in a while (2+ years) that I was able to use `npm update --save` instead of updating dependencies one at a time to not break our lock file.

daragh-king-genesys:
> @daragh-king-genesys I did a clean install locally, ran build/tests etc and read through the changelogs for the dependencies to see if there was anything unexpected there. Is that the right way to review this kind of change or would you have any other suggestions?

The only other thing I do is make sure the time machine branch was created successfully and click around a bit on it to make sure there is nothing obviously broken

PR #815, opened 2025-01-08
chore(table-toolbar): manual check-a11y

gavin-everett-genesys:
manual check-a11y

[✅ Closes: COMUI-3340](https://inindca.atlassian.net/browse/COMUI-3340)


packages/genesys-spark-components/scripts/check-a11y.mjs
```diff
@@ -9,7 +9,7 @@ const __dirname = dirname(__filename);
 // The description of each success criterion is copied directly from the WCAG Understanding Success Criteria pages
 const a11yRules = [
   {
-    id: '1.1.1',
+    id: '111',
```
gavin-everett-genesys, line 12: Script was failing so removed the dots.

PR #822, opened 2025-01-08
feat(dropdown): option icon position added

daragh-king-genesys:
option icon position added

✅ Closes: COMUI-3355

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.scss
@@ -201,7 +201,12 @@
 
 .gux-selected-icon {
   display: flex;
+  flex-direction: row;
   align-items: center;
+
+  &.gux-icon-position-end {
+    flex-direction: row-reverse;
+  }
 
   gux-icon {
     inline-size: var(--gse-ui-formControl-input-inputIcon-size);
--- a/packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option-icon/gux-option-icon.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option-icon/gux-option-icon.tsx
@@ -105,7 +105,11 @@ export class GuxOptionIcon {
     return parentListbox?.disabled;
   }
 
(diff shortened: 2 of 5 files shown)


packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.scss
```diff
@@ -201,8 +201,13 @@
 
 .gux-selected-icon {
   display: flex;
+  flex-direction: row;
   align-items: center;
 
+  &.gux-icon-position-end {
+    flex-direction: row-reverse;
+  }
```
gavin-everett-genesys, line 209: I was just reading the accessibility for `row-reverse` (https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction#accessibility) and it says it has some drawbacks for users with low vision as they wont have access to the correct  reading order as the DOM order does not change. I was wondering are we okay with this ? Would it work in the shadowDOM if you conditionally render the icon to be after the `gux-truncate` if the `iconPosition=end` ?

packages/genesys-spark-components/src/components/stable/gux-listbox/options/gux-option-icon/gux-option-icon.tsx
```diff
@@ -41,6 +41,9 @@ export class GuxOptionIcon {
   @Prop()
   iconColor: string;
 
+  @Prop()
+  iconPosition: 'start' | 'end' = 'start';
```
gavin-everett-genesys, line 45: Minor to add an example in the docs

daragh-king-genesys: Done

packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.scss
daragh-king-genesys: I addressed this in the `gux-option-icon` component but here as the icon can only be decorative I don't think there is an accessibility concern.

gavin-everett-genesys (approved):

PR #847, opened 2025-01-17
chore(form-field-range): refactoring some css

jordanstith15:
Cleaning up some css that was changed based on the new tokens from [COMUI-3356](https://inindca.atlassian.net/browse/COMUI-3356)

[COMUI-3356]: https://inindca.atlassian.net/browse/COMUI-3356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-range/gux-form-field-range.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-range/gux-form-field-range.scss
@@ -84,8 +84,7 @@ $thumb-diameter: calc(var(--gse-ui-rangeSlider-handle-height) / 2);
     gap: var(--gse-ui-rangeSlider-gap);
     place-content: stretch flex-start;
     align-items: center;
-    // We will not use a token for this height style below
-    block-size: 32px;
+    block-size: var(--gse-ui-rangeSlider-set-height);
     font-size: var(--gse-ui-rangeSlider-label-text-fontSize);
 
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-range/gux-form-field-range.scss
```diff
@@ -35,7 +35,7 @@
 $range-active: var(--gse-ui-rangeSlider-bar-selected-backgroundColor);
 $track-width: 100%;
 $track-height: var(--gse-ui-rangeSlider-bar-height);
-$thumb-diameter: calc(var(--gse-ui-rangeSlider-handle-width) / 2);
+$thumb-diameter: calc(var(--gse-ui-rangeSlider-handle-height) / 2);
```
gavin-everett-genesys, line 38: Should this stay using the `width` token ?

jordanstith15: so I changed it because everything it's used in has to do with block margins and sizing, which is height in LTR from my understanding. It makes more sense to use height in this case imo.

Maybe we should talk to UX about the token naming too if we are switching everything to block and inline sizing?

gavin-everett-genesys (approved):

PR #857, opened 2025-01-22
feat(form-field-checkbox): adding aria attributes to help with error accessibility

conor-darcy:
Adds 2 aria attributes to the form-field-checkbox & form-field-radio.

See: https://www.smashingmagazine.com/2023/02/guide-accessible-form-validation/
[aria-invalid](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-invalid)
[aria-errormessage](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-errormessage)

aria-errormessage has partial support - https://a11ysupport.io/tech/aria/aria-errormessage_attribute

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/tests/gux-form-field-checkbox.spec.ts
+++ b/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/tests/gux-form-field-checkbox.spec.ts
@@ -61,9 +61,38 @@ describe('gux-form-field-checkbox', () => {
     `
     ].forEach((html, index) => {
       it(`should render component as expected (${index + 1})`, async () => {
+        const page = await newSpecPage({
+          components,
+          html,
+          language
+        });
+
+        // wait for change inside component
+        await page.waitForChanges();
+        expect(page.root).toMatchSnapshot();
+      });
+    });
+
+    describe('when the component has an error slot', () => {
+      it('should apply aria invalid and errormessage', async () => {
+        const html = `
+          <gux-form-field-checkbox>
+            <input slot="input" type="checkbox" name="food-1[]" value="pizza"/>
+            <label slot="label">Pizza</label>
+            <span slot="error">This is an error message</span>
+          </gux-form-field-checkbox>
+        `;
         const page = await newSpecPage({ components, html, language });
 
-        expect(page.root).toMatchSnapshot();
+        await page.waitForChanges();
+
+        expect(
+          page.root.querySelector('input').getAttribute('aria-invalid')
+        ).toBe('true');
+
+        expect(
+          page.root.querySelector('input').getAttribute('aria-errormessage')
+        ).toBeDefined();
       });
     });
   });
--- a/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/tests/gux-form-field-checkbox.spec.ts
+++ b/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/tests/gux-form-field-checkbox.spec.ts
@@ -67,8 +67,6 @@ describe('gux-form-field-checkbox', () => {
           language
         });
 
-        // wait for change inside component
-        await page.waitForChanges();
         expect(page.root).toMatchSnapshot();
       });
     });
@@ -83,8 +81,6 @@ describe('gux-form-field-checkbox', () => {
           </gux-form-field-checkbox>
         `;
         const page = await newSpecPage({ components, html, language });
-
-        await page.waitForChanges();
 
         expect(
           page.root.querySelector('input').getAttribute('aria-invalid')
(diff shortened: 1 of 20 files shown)


packages/genesys-spark-components/src/components/stable/gux-form-field/gux-form-field.service.ts
```diff
@@ -83,6 +83,11 @@ export function validateFormIds(
     error.setAttribute('id', errorId);
     describedByIds.push(errorId);
 
+    if (input instanceof HTMLInputElement) {
```
jason-evans-genesys, line 86: This is probably outside the scope of this PR but I noticed this `validateFormIds` function this code is in is getting kind of long. Do you think it's worth us having a separate ticket to refactor this function?

packages/genesys-spark-components/src/components/stable/gux-form-field/gux-form-field.service.ts
```diff
@@ -83,6 +83,11 @@ export function validateFormIds(
...
+      input.setAttribute('aria-invalid', 'true');
+      input.setAttribute('aria-errormessage', errorId);
```
jason-evans-genesys, line 88: Saw this in the documentation: 
```An ARIA attribute dedicated to errors already exists — aria-errormessage — but it’s not yet supported by most screen readers. So, for now, you are better off avoiding it and sticking with aria-describedby```

File: packages/genesys-spark-components/src/components/stable/gux-form-field/gux-form-field.service.ts [hunk 3]
```diff
@@ -83,6 +83,11 @@ export function validateFormIds(
     error.setAttribute('id', errorId);
     describedByIds.push(errorId);
```
daragh-king-genesys, line 84: The error text is now connected to the input twice. `aria-describedby` and `aria-errormessage`. Should we remove the error id from the  `aria-describedby`?

packages/genesys-spark-components/src/components/stable/gux-form-field/gux-form-field.service.ts
daragh-king-genesys, line 86: Why this restriction?

daragh-king-genesys (commented):
I am surprised the change in `gux-form-field.service.ts` did not effect more snapshots.

packages/genesys-spark-components/src/components/stable/gux-form-field/gux-form-field.service.ts
conor-darcy: Yes jason, here a ticket to refactor it
https://inindca.atlassian.net/browse/COMUI-3429

packages/genesys-spark-components/src/components/stable/gux-form-field/gux-form-field.service.ts
conor-darcy: Thanks Jason, I didn't see that, I'll remove it & maybe we can look at it again in the future.

packages/genesys-spark-components/src/components/stable/gux-form-field/gux-form-field.service.ts
conor-darcy: Removing based on Jason's comment above, the aria-describedby hack is GEFN.

packages/genesys-spark-components/src/components/stable/gux-form-field/gux-form-field.service.ts
conor-darcy: Some slotted 'inputs' are actually web components, `aria-invalid` can only be applied to an` <input>`.
Eg. without the condition in `gux-form-field-phone`, `aria-invalid` is added to the slotted `gux-phone-input-beta`.

conor-darcy:
> I am surprised the change in `gux-form-field.service.ts` did not effect more snapshots.

@daragh-king-genesys Yes, some snapshots missing from error states. I've added them.

daragh-king-genesys:
@conor-darcy still some failing snapshots

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/tests/gux-form-field-checkbox.spec.ts
```diff
@@ -61,10 +61,39 @@ describe('gux-form-field-checkbox', () => {
...
       it(`should render component as expected (${index + 1})`, async () => {
-        const page = await newSpecPage({ components, html, language });
+        const page = await newSpecPage({
+          components,
+          html,
+          language
+        });
 
+        // wait for change inside component
+        await page.waitForChanges();
```
daragh-king-genesys, line 71: What changes? We don't usually have to do this.

daragh-king-genesys (commented):
LGTM will approve once tests are passing

daragh-king-genesys (approved):

PR #861, opened 2025-01-28
chore(form-field-text-like): Manual accessibility updates (WIP)

jason-evans-genesys:
✅ Closes: COMUI-3408

*Update: spoke with @katie-bobbe-genesys and the current solution in this PR is not what we want so will brainstorm some other ideas.


jason-evans-genesys:
@katie-bobbe-genesys I set the input's `aria-describedby` to be the id of the loading element, however, this doesn't seem to trigger anything new from the screenreader when using voiceover on mac. Are we sure `aria-describedby` is needed here?

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-text-like/gux-form-field-text-like.tsx
```diff
@@ -176,6 +179,10 @@ export class GuxFormFieldTextLike {
       setSlotAriaDescribedby(this.root, this.input, 'suffix');
     }
 
+    if (this.loading) {
+      setSlotAriaDescribedby(this.root, this.input, 'loading');
```
jason-evans-genesys, line 183: @katie-bobbe-genesys thoughts on this implementation? I'm slotting in a span with the text that I then link up to the input tag with aria-describedby. I was thinking the developer could slot in the exact loading text they wanted. Thoughts?

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-text-like/example.html
```diff
@@ -210,6 +210,7 @@ <h2>Examples</h2>
     <gux-form-field-text-like loading>
       <input slot="input" type="text" name="e-1" />
       <label slot="label">Default</label>
+      <span slot="loading">Content is loading</span>
```
jason-evans-genesys, line 213: @katie-bobbe-genesys my idea was to slot in the loading text and then read it out to the screenreader.

PR #865, opened 2025-01-29
chore(toast): address a11y issue and update example usage guidance

katie-bobbe-genesys:


daragh-king-genesys (commented):
Could we add a check/warning in the console if the component can't find a parent with the correct `aria-live` attribute?

daragh-king-genesys (approved):

PR #868, opened 2025-01-30
fix(dropdown-multi): fixing typeahead text and adding selected options text in input

jordanstith15:
Closes [COMUI-3312](https://inindca.atlassian.net/browse/COMUI-3312), [COMUI-3325](https://inindca.atlassian.net/browse/COMUI-3325), and [COMUI-3326](https://inindca.atlassian.net/browse/COMUI-3326)

[COMUI-3312]: https://inindca.atlassian.net/browse/COMUI-3312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[COMUI-3325]: https://inindca.atlassian.net/browse/COMUI-3325?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[COMUI-3326]: https://inindca.atlassian.net/browse/COMUI-3326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-listbox-multi.tsx
```diff
@@ -257,8 +257,10 @@ export class GuxListboxMulti {
         listboxOption.value
       );
       if (this.filterType !== 'custom') {
-        listboxOption.filtered = !listboxOption.textContent
-          .trim()
+        listboxOption.filtered = !listboxOption.shadowRoot
+          .querySelector('slot')
+          ?.assignedNodes()[0]
+          ?.textContent.trim()
```
jordanstith15, line 263: This allows the user to slot things in any order. For some reason, that querySelector plus assignNodes only grabs the default slot and not the unnamed ones

packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi.tsx
```diff
@@ -471,12 +471,33 @@ export class GuxDropdownMulti {
...
 
+  private getSelectedOptionText(): string | false {
+    const selectedListboxOptionElement = this.getOptionElementByValue(
+      this.value
+    );
+
+    if (selectedListboxOptionElement?.length) {
+      return selectedListboxOptionElement
+        .map(option => {
+          return option.shadowRoot
```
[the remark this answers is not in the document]
jordanstith15: Took a look and it doesn't seem to simplify the code unfortunately, but making a service function for the component is probably a good idea

packages/genesys-spark-components/src/components/stable/gux-listbox-multi/gux-listbox-multi.tsx
```diff
@@ -256,9 +257,9 @@ export class GuxListboxMulti {
       listboxOption.selected = this.getSelectedValues().includes(
         listboxOption.value
       );
-      if (this.filterType !== 'custom') {
-        listboxOption.filtered = !listboxOption.textContent
-          .trim()
+      if (this.filterType !== 'custom' && this.filterType !== 'none') {
```
jordanstith15, line 260: Looks like my change exposed this missing condition and that's why the snapshots and gux-phone-input were acting weird

gavin-everett-genesys:
LGTM 👍  Some manual testing stuff I found,

- The text when truncated does not show a tooltip.
- When the dropdown is expanded and you have one or more items selected should the items selected be the the display text instead of "Select a country" ? I think this only occurs in the first dropdown with the filter applied. 
- If you go to the `Unknown Value Set` example and dynamically add the Jelly option it will display blank in the dropdowns display text.

jordanstith15:
> https://apps.inindca.com/common-ui-docs/genesys-webcomponents/feature/COMUI-3312

1. I don't think it needs a tooltip. It's only a placeholder that is still read out by screen-readers, and clicking the input will show you the selected items. UX just wanted it as an extra thing to show that there are selected items, according to the ticket.

2. I personally think it should show Select a Country when open like it does, but I think it's supposed to update as you select according to UX. I'll look into it.

3. Will look into this as well. Looks like it isn't re-rendering when the new option is added until the dropdown is opened and then closed. I guess the value isn't fully updating until then? Not sure. Just speculation on where to look

packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi.tsx
```diff
@@ -347,16 +348,14 @@ export class GuxDropdownMulti {
...
-      return listboxOptionElements.filter(element =>
-        values.includes(element.value)
-      );
-    }
-    return;
+    const values = value ? value.split(',') : [];
+
+    return listboxOptionElements.filter(element =>
+      values.includes(element.value)
+    );
```
jordanstith15, line 358: Didn't like that this was returning undefined despite return type being array, which caused every call of this method to need to do a null-check and overly complicating things.

packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi.tsx
```diff
@@ -229,7 +230,7 @@ export class GuxDropdownMulti {
   onClearselected(event: CustomEvent): void {
     event.stopPropagation();
 
-    this.updateValue('');
+    this.updateValue(undefined);
```
jordanstith15, line 233: Change was needed based on the change described in the comment on line 358, otherwise the number selected tag was still showing "1" when nothing was selected.
This was the only place value was being set to empty string instead of undefined when empty and it was causing inconsistencies in code and behavior

gavin-everett-genesys:
Jellyfish option LGTM 👍 

I think you mentioned this in one of your comments and maybe you did not push a fix on the latest branch it shows the `1` still if you select and deselect.

https://github.com/user-attachments/assets/aa95b555-e4f4-47f7-ab0b-608fbd1687c7

jordanstith15:
> Jellyfish option LGTM 👍
> 
> I think you mentioned this in one of your comments and maybe you did not push a fix on the latest branch it shows the `1` still if you select and deselect.


I fixed it for when you hit the X button. This is new and am looking into it

gavin-everett-genesys (approved):

PR #872, opened 2025-02-04
feat(rich-text-editor) - Responsive design menu

gavin-everett-genesys:
https://inindca.atlassian.net/browse/COMUI-2963


packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-action/gux-rich-text-editor-action-rich-style/gux-rich-text-editor-action-rich-style.tsx
```diff
@@ -35,6 +35,9 @@ export class GuxRichTextEditorActionRichStyle {
   @Prop({ mutable: true })
   value: string;
 
+  @Prop()
+  isMenu: boolean = false;
```
gavin-everett-genesys, line 39: The reason for this property is because I wanted to use this component for the responsive list as it has the internals I would need. I could create a new component and just render this inside it but didn't think it was needed but open to opinions. I know this `prop` is exposed in the API but I think it should not have any downsides.

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
```diff
@@ -16,12 +31,202 @@ import { hasSlot } from '@utils/dom/has-slot';
...
+  typographicalEmphasisActions: string[] = [];
+
+  @State()
+  textStylingActions: string[] = [];
+
+  @State()
+  listsAndIndentationActions: string[] = [];
+
+  @State()
+  insertingActions: string[] = [];
```
gavin-everett-genesys, line 52: I do combine these into one down below in the render area so maybe not all of these are needed. Open to opinions.

gavin-everett-genesys:
Getting build error with observer looking into it.

gavin-everett-genesys:
This PR is now ready for review. Discussed with UX how the link action would work in the condensed menu we agreed that it would be better to just have the link action always visible in the toolbar if the user includes it.

321gillian:
Screenshot 2025-02-12 at 11 47 28
Is there a way to add some kind of colour indicator alongside the highlight colours? The paragraph style list still kind of points to the expanded version but the highlight one seems quite different. Might be for another ticket.

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
```diff
@@ -16,12 +31,210 @@ import { hasSlot } from '@utils/dom/has-slot';
...
+    const actionGroups = Array.from(
+      this.root.querySelectorAll('gux-rich-text-editor-action-group')
+    );
+
+    // Hide last visible action group until there's no overflow.
+    while (this.isOverFlowing()) {
+      let hidden = false;
+      for (let i = actionGroups.length - 1; i >= 0; i--) {
+        if (!actionGroups[i].classList.contains('gux-hidden')) {
+          actionGroups[i].classList.add('gux-hidden');
```
321gillian, line 103: Would this attribute be useful here? https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
```diff
@@ -16,12 +31,210 @@ import { hasSlot } from '@utils/dom/has-slot';
...
+  private i18n: GetI18nValue;
+
   @Element()
   root: HTMLElement;
 
   @Prop()
   disabled: boolean = false;
 
+  @State()
+  typographicalEmphasisActions: string[] = [];
```
321gillian, line 43: Are there any limits on what type of string could be in here or is it open?

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
```diff
@@ -16,12 +31,210 @@ import { hasSlot } from '@utils/dom/has-slot';
...
+                <gux-truncate max-lines={1}>
+                  {this.i18n('noHighlight')}
+                </gux-truncate>
+              </gux-rich-style-list-item>
+            </gux-rich-text-editor-sub-list>
+          )}
+        </gux-rich-text-editor-action-rich-style>
+      ) as JSX.Element;
+    }
+  }
```
321gillian, line 236: If possible, it might be worth splitting this function up into smaller parts

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-action/gux-rich-text-editor-action-rich-style/gux-rich-text-editor-action-rich-style.tsx
321gillian: The meaning established for `gux-rich-text-editor-action-rich-style` in the previous version linked `style` to the 'paragraph style' meaning.

I think if you want this to be handle both scenarios, it needs to be a kind of flexible generic menu. The style bit needs to be removed. Otherwise I think it's probably better to just duplicate and make a distinct overflow menu component.

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
321gillian: Not sure if this is better or worse. If you decide to type the actions (I don't know much about what allowed in an rte), you could make a type for all actions and then use that. I'm not against the distinct state items for each though.

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-list/gux-rich-text-editor-sub-list/gux-rich-text-editor-sub-list.tsx
```diff
@@ -0,0 +1,230 @@
...
+  @Prop()
+  label: string;
+
+  @State()
+  isShown: boolean = false;
+
+  @Method()
+  async guxFocus(): Promise<void> {
+    this.buttonElement.focus();
+  }
```
321gillian, line 47: You might not need this if you're using `delegatesFocus`. I think you can just call `focus` on the sub-list and this will happen.

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-list/gux-rich-style-list-item/gux-rich-style-list-item.tsx
```diff
@@ -12,15 +20,24 @@ import { getClosestElement } from '@utils/dom/get-closest-element';
...
-  @Prop()
+  // Reflect is used here so we can access the value in the light DOM for executing actions.
+  @Prop({ reflect: true })
   value: string;
 
+  @Method()
+  // eslint-disable-next-line @typescript-eslint/require-await
+  async guxFocus(): Promise<void> {
+    this.richStyleElement.focus();
+  }
```
321gillian, line 39: Again might not be needed. This was the case when I was looking at the avatar group items. I could remove the `guxFocus` stuff and just use focus delegation and call `focus` directly.

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-list/gux-rich-text-editor-sub-list/gux-rich-text-editor-sub-list.scss
```diff
@@ -0,0 +1,41 @@
...
+button {
+  &.gux-sub-list-button {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+  }
+
+  gux-icon {
+    align-self: center;
+    color: #1b2c48; // Token Required;
```
321gillian, line 18: Is there a ticket for this?

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-action/gux-rich-text-editor-action-rich-style/gux-rich-text-editor-action-rich-style.tsx
gavin-everett-genesys: Yeah after looking at this again, just decided to create its own internal component called `gux-rich-text-editor-menu` for the responsive list as it probably looks a bit confusing with the extra `isMenu` prop also.

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
gavin-everett-genesys: Did some research on using the `hidden` attribute vs `display:none`. AFAIK in this context  it would be more reasonable to use `display:none` over the `hidden` attribute. The `hidden` attribute is about semantics whether something is currently part of the page content. Example usages of the `hidden` attribute would be an error message or a thank you message that do not exist until an action is complete whereas `display:none`  is about the presentation of the page content. By using `display:none` here it would indicate the groups still exist as part of the page content but we have hidden it for connivence.



https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute

https://stackoverflow.com/questions/6708247/what-is-the-difference-between-the-hidden-attribute-html5-and-the-displaynone

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
gavin-everett-genesys: Tidied this up as best I can let me know if its more readable now.

321gillian (approved):

PR #874, opened 2025-02-05
feat(avatar-groups): Add to group element

321gillian:
COMUI-3051


packages/genesys-spark-components/src/components/beta/gux-avatar-group/gux-avatar-group-add-item/gux-avatar-group-add-item.tsx
```diff
@@ -0,0 +1,67 @@
...
+  render(): JSX.Element {
+    return (
+      <Host role="menuitem">
+        <button
+          type="button"
+          aria-label={this.i18n('addToGroup')}
+          tabIndex={-1}
+          class="gux-avatar"
+        >
+          <span aria-hidden="true">+</span>
```
gavin-everett-genesys, line 54: Would `fa/plus-regular` icon work here ?

321gillian: Good question, I've sent this query to UX since it was '+' text in the Figma

321gillian: Confirmed with UX that it's the '+' text for now. They may swap it out for the icon later on but this is what they want to start with.

PR #884, opened 2025-02-11
fix(tooltip-base-beta): improve accessibility

daragh-king-genesys:
✅ Closes: COMUI-3420


packages/genesys-spark-components/src/components/beta/gux-tooltip-base/gux-tooltip-base.tsx
```diff
@@ -230,6 +246,7 @@ export class GuxTooltipBase {
           data-placement={this.placement}
         >
           <slot name="content" />
+          <div id="arrow"></div>
```
daragh-king-genesys, line 249: If you use something like
```
<div class="arrow" ref={el => (this. arrowElement = el)}></div>
```
you won't need to manually find the element in the setForElement function

PR #892, opened 2025-02-12
docs(loading-message): added example that highlights aria-live functi…

daragh-king-genesys:
…onality

added example that highlights aria-live functionality

✅ Closes: COMUI-812


jason-evans-genesys:
@daragh-king-genesys is the expected behavior to have an announcement while using voiceover when the loading message is injected into the DOM?

gavin-everett-genesys (approved):

PR #899, opened 2025-02-14
fix(option): trim whitespace for subtext selection

321gillian:
✅ Closes: COMUI-3491

Here's an example for testing:
```
<gux-option value="us-east-1" id="gux-option-4qupoj7mux" role="option" class="gux-show-subtext gux-selected" aria-selected="true" aria-disabled="false" hydrated=""><!--?lit$820817559$-->Americas (US East) (<!--?lit$820817559$-->us-east-1)

        <span slot="subtext">
https://login.mypurecloud.com</span>
      </gux-option>
```

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
@@ -438,13 +438,11 @@ export class GuxDropdown {
   }
 
   private renderOption(option: HTMLGuxOptionElement): JSX.Element {
-    let optionText = option.textContent;
+    let optionText = option.textContent.trim();
     if (hasSlot(option, 'subtext')) {
+      // TODO: use getOptionDefaultSlot(option)?.textContent.trim() once Stencil fix for assignedNodes test issue is in (v4.27.2)
       const subtext = option.querySelector('[slot=subtext]');
-      optionText = optionText.substring(
-        0,
-        optionText.length - subtext.textContent.length
-      );
+      optionText = optionText.replace(subtext.textContent, '');
     }
     return (
       <gux-truncate ref={el => (this.truncateElement = el)} dir="auto">


jordanstith15:
Using your example for testing, I'm having trouble reproducing the original issue. Both your demo link and the latest version look the same to me..

jordanstith15 (changes requested):
I found an issue. It's the same one that I ran into with my dropdown-multi PR. I must have missed it here.

If you have the subtext slot before the text, then the selected text and the tooltip completely break.
IIRC I completely rewrote the logic for this in the dropdown-multi so it doesn't even need to check whether there is subtext

321gillian:
> I found an issue. It's the same one that I ran into with my dropdown-multi PR. I must have missed it here.
> 
> If you have the subtext slot before the text, then the selected text and the tooltip completely break. IIRC I completely rewrote the logic for this in the dropdown-multi so it doesn't even need to check whether there is subtext
> 
> Edit: You should be able to just do `getOptionDefaultSlot(option)?.textContent.trim()` getOptionDefaultSlot is in the gux-listbox.service.ts file

Hi @jordanstith15 

I ran into some issues with using getOptionDefaultSlot. It was causing the spec tests for dropdown  and pagination to break with TypeError: _a.assignedNodes is not a function.  I tried adding some conditions to check that the slot was present etc but it seems like maybe a timing issue coming from the dropdown. In this case, I thought maybe it might get around this by just taking the text from the light dom. 

Let me know if you have any other thoughts on it. 
Thanks,
Gillian

jordanstith15:
> > I found an issue. It's the same one that I ran into with my dropdown-multi PR. I must have missed it here.
> > If you have the subtext slot before the text, then the selected text and the tooltip completely break. IIRC I completely rewrote the logic for this in the dropdown-multi so it doesn't even need to check whether there is subtext
> > Edit: You should be able to just do `getOptionDefaultSlot(option)?.textContent.trim()` getOptionDefaultSlot is in the gux-listbox.service.ts file
> 
> Hi @jordanstith15
> 
> I ran into some issues with using getOptionDefaultSlot. It was causing the spec tests for dropdown and pagination to break with TypeError: _a.assignedNodes is not a function. I tried adding some conditions to check that the slot was present etc but it seems like maybe a timing issue coming from the dropdown. In this case, I thought maybe it might get around this by just taking the text from the light dom.
> 
> Let me know if you have any other thoughts on it. Thanks, Gillian

I found this issue with stencil that is likely the cause
https://github.com/stenciljs/core/issues/2830

And there was a merged fix only a month and a half ago which we might not be updated to
https://github.com/stenciljs/core/pull/6108

I'll approve for now, but we might want to change it once we have the fix from stencil.

jordanstith15 (approved):

PR #920, opened 2025-02-19
fix(list): address accessibility console error

katie-bobbe-genesys:
Fix a console error that appears after opening a component that contains a `gux-list` then hovering over the first element:

```
gux-context-menu.html:1 Blocked aria-hidden on an element because its descendant retained focus. 
The focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.
Element with focus: span
Ancestor with aria-hidden:  <span tabindex=​"-1" aria-hidden=​"true">​</span>​
```

I tried adding a test to catch this error but I could not get the accessibility test to fail.


jordanstith15:
LGTM. Just needs the snapshots updated

packages/genesys-spark-components/src/components/stable/gux-list/gux-list.tsx
```diff
@@ -84,14 +84,9 @@ export class GuxList {
...
   }
 
-  private renderFocusTarget(): JSX.Element {
-    return (<span tabindex="-1" aria-hidden="true"></span>) as JSX.Element;
-  }
-
   render(): JSX.Element {
     return (
-      <Host role="list">
-        {this.renderFocusTarget()}
```
katie-bobbe-genesys, line 94: There is an issue with having this focus target element. Both of the following cause accessibility issues
1. Having a focusable element with a role of `aria-hidden` is a violation
2. If you remove the `aria-hidden`, having an element with an invalid role as a descendent of the `list` is a violation.
I have removed this focus target span and instead put a tab index of `-1` on the Host.

packages/genesys-spark-components/src/components/stable/gux-list/gux-list-item/gux-list-item.tsx
```diff
@@ -18,27 +16,6 @@ export class GuxListItem {
   @Prop()
   disabled: boolean = false;
 
-  @Listen('mouseup')
```
katie-bobbe-genesys, line 21: @daragh-king-genesys I am tagging you here because you seem to be the one who added this code. Can you remember if this code has any utility that I may be overlooking? I am removing it because it is causing unwanted focus behavior after my changes in the `gux-list` file.

jordanstith15 (approved):

packages/genesys-spark-components/src/components/stable/gux-list/gux-list-item/gux-list-item.tsx
daragh-king-genesys: Sorry I missed this comment at the time Katie. I assume this is why you have not merged.
I have no idea and happy to remove it if you think it is causing issues.

While I was looking into it I noticed the same code is in a few other list item like components:
- gux-month-list-item
- gux-rich-highlight-list-item
- gux-rich-style-list-item

We will need an issue to look into removing the code from there too as I assume they are based on this original component.

daragh-king-genesys (approved):

PR #934, opened 2025-02-24
fix(tab): focus ring cut off on bottom

jordanstith15:
Closes [COMUI-3411](https://inindca.atlassian.net/browse/COMUI-3411)

[COMUI-3411]: https://inindca.atlassian.net/browse/COMUI-3411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-tabs/gux-tab/gux-tab.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-tabs/gux-tab/gux-tab.scss
@@ -94,10 +94,10 @@ gux-tab {
     }
 
     &:focus-visible {
-      outline: none;
-      border: var(--gse-ui-tabs-focusRing-border-width)
+      outline: var(--gse-ui-tabs-focusRing-border-width)
         var(--gse-ui-tabs-focusRing-border-style)
         var(--gse-ui-tabs-focusRing-border-color);
+      outline-offset: calc(var(--gse-ui-tabs-focusRing-border-width) * -1);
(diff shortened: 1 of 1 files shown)


katie-bobbe-genesys:
This is pretty subtle, but I noticed when the selected tab gets tab focus, the underline that indicates the selected state gets hidden by this focus ring. Is there a way to preserve the dark underline so that we can see the selected state when focused? Or maybe we need some kind of other stronger indicator for the selected tab state.

Before: Focus ring is cut off but we can still see selected state
image

After: Cannot see selected state
image

jordanstith15:
> This is pretty subtle, but I noticed when the selected tab gets tab focus, the underline that indicates the selected state gets hidden by this focus ring. Is there a way to preserve the dark underline so that we can see the selected state when focused? Or maybe we need some kind of other stronger indicator for the selected tab state.
> 
> Before: Focus ring is cut off but we can still see selected state image
> 
> After: Cannot see selected state image

So I personally think it's fine because you would see it's selected before you focus and if you move to another tab at all, but one thing I could do that looks alright is change the outline color when on the selected one. I tried increasing the negative offset of the outline, but because the border-radius is also set on focus, it makes the active border-bottom look weird.

The other option that seems to work too is to increase the width of the border bottom by a pixel.

daragh-king-genesys (approved):

katie-bobbe-genesys (approved):

PR #948, opened 2025-02-27
fix(form-field-file-beta): moved error slot location

daragh-king-genesys:
moved error slot location


gavin-everett-genesys:
LGTM - I was just wondering when in single file mode and dragging multiple files. It just adds one file I assume just the first of the multiple selected files. I guess thats fine is it ? Or should no files be added at all.

PR #955, opened 2025-02-28
chore(side-panel-beta): add manual a11y check

katie-bobbe-genesys:


packages/genesys-spark-components/src/components/beta/gux-side-panel/gux-side-panel.tsx
```diff
@@ -54,27 +55,29 @@ export class GuxSidePanel {
...
-          <gux-dismiss-button
-            position="inherit"
-            onClick={this.onDismissHandler.bind(this)}
-          />
-          <slot name="heading" />
-        </header>
-        {this.renderDescription()}
-        <div class="gux-side-panel-content">
-          <slot name="content" />
+      <Host role="complementary">
```
katie-bobbe-genesys, line 58: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/complementary_role

daragh-king-genesys (approved):

packages/genesys-spark-components/src/components/beta/gux-side-panel/gux-side-panel.tsx
jason-evans-genesys: Does it make sense using an aside element instead of adding this role on the Host?

PR #970, opened 2025-03-06
fix(dropdown-multi): fixing issue when user sets value to empty string

jordanstith15:
Closes [COMUI-3516](https://inindca.atlassian.net/browse/COMUI-3516)

[COMUI-3516]: https://inindca.atlassian.net/browse/COMUI-3516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi.tsx
@@ -175,7 +175,7 @@ export class GuxDropdownMulti {
    */
   @Method()
   getSelectedValues(): Promise<string[]> {
-    return Promise.resolve(this.value?.split(',') || []);
+    return Promise.resolve(this.value?.split(',').filter(i => i !== '') || []);
   }
 
   @Listen('keydown')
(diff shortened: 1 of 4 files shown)


packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi.tsx
```diff
@@ -175,7 +175,7 @@ export class GuxDropdownMulti {
    */
   @Method()
   getSelectedValues(): Promise<string[]> {
-    return Promise.resolve(this.value?.split(',') || []);
+    return Promise.resolve(this.value?.split(',').filter(i => i !== '') || []);
   }
```
daragh-king-genesys, line 180: I think you could move this logic of converting the value string into an array into one function and then call it in the three locations.

```
private convertValueToArray(value: string): string[] {
  if (value) {
    return value.split(',');
  }

  return [];
}

jordanstith15: Now is where I find out that something like `value?.<function>` is not quite the same as an if. Empty string returns false in an if statement and tertiary, but returns true for that 🤦‍♂️

daragh-king-genesys (approved):

PR #979, opened 2025-03-10
chore(form-field-text-like): Added role=status for loading indicator

jason-evans-genesys:
✅ Closes: COMUI-3408

I added role="status" around the loading indicator for now so voiceover will pick up when the loading indicator is visible or not. 

There is a new ticket to ask UX why we need a loading indicator for all inputs and then also maybe making an API change to slot in the loading indicator element: https://inindca.atlassian.net/browse/COMUI-3539


packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-text-like/example.html
```diff
@@ -232,7 +240,13 @@ <h2>Examples</h2>
   </fieldset>
 </form>
 
-<style>
+<style
+  onload="(function () {
+  window.setInterval(() => {
+    document.getElementById('dynamic-loading').toggleAttribute('loading');
```
jason-evans-genesys, line 246: Do you all think it's not optimal to use setInterval here or does it not matter since it's an example page and I'm trying to demonstrate the loading indicator dynamically showing/hiding so we can see if voiceover picks up on the change?

katie-bobbe-genesys: Were you able to get voiceover to announce the change when it is dynamically showing/hiding? I wasn't able to get my voiceover to read it out

jason-evans-genesys: @katie-bobbe-genesys I notice voiceover announcing the change if I'm focused into this input:
Screenshot 2025-03-24 at 4 39 22 PM

gavin-everett-genesys: @jason-evans-genesys It does announce `Loading` for me in chrome and firefox when I am on the text-field itself when using voiceover for me. So I think this is fine ? 

See video below : 

https://github.com/user-attachments/assets/9379f933-9f87-4991-862a-c32def179279

Happy to approve there is just a failing build.

jason-evans-genesys: @gavin-everett-genesys build error fixed!

gavin-everett-genesys (approved):

PR #982, opened 2025-03-12
fix(copy-to-clipboard): display none on tooltip when hidden

321gillian:
COMUI-3518

There are behaviour questions that need to be worked out in relation to the tooltip also but this is something we should fix. 

Before:
tooltip can be reactivated by hovering over the area where it was previously shown
https://github.com/user-attachments/assets/c6856ccd-4a2d-4e00-8263-cd4a7673a7c1

After:
tooltip can be hovered over when displayed but when dismissed, it must be again triggered by the owning element 
https://github.com/user-attachments/assets/202b84f4-c79c-47a5-a9fc-9d52642152c1

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-copy-to-clipboard/gux-copy-to-clipboard.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-copy-to-clipboard/gux-copy-to-clipboard.scss
@@ -59,7 +59,7 @@ button {
   }
 }
 
-gux-tooltip {
+gux-tooltip.gux-show {
   display: inline-flex;
   align-items: center;
 


packages/genesys-spark-components/src/components/beta/gux-tooltip-base/gux-tooltip-base.tsx
```diff
@@ -99,16 +99,6 @@ export class GuxTooltipBase {
...
-  @Listen('pointerenter')
-  handlePointerenter() {
-    this.show();
-  }
-
-  @Listen('pointerleave')
-  handlePointerleave() {
-    this.hide();
-  }
-
```
321gillian, line 111: I removed this because I couldn't see what it was for from testing. Happy to replace if we identify that.

packages/genesys-spark-components/src/components/beta/gux-tooltip-base/gux-tooltip-base.tsx
```diff
@@ -181,9 +171,17 @@ export class GuxTooltipBase {
...
       }
+
+      this.resetTooltipPosition();
     }, 350);
   }
 
+  private resetTooltipPosition() {
+    // This prevents tooltip from reappearing on pointerenter it's previous position
+    this.root.style.left = '';
+    this.root.style.top = '';
```
daragh-king-genesys, line 182: Is this just moving the tooltip to somewhere else on the screen?

daragh-king-genesys (approved):

packages/genesys-spark-components/src/components/beta/gux-tooltip-base/gux-tooltip-base.tsx
321gillian: Good point, redid it using pointer-events. I'll test again in the morning

daragh-king-genesys:
@321gillian did you figure out exactly why this was happening? I don't understand how an element with `display:none` has pointer events. What am I missing?

321gillian:
> @321gillian did you figure out exactly why this was happening? I don't understand how an element with `display:none` has pointer events. What am I missing?

I had another look this morning based on your question and it looks like the issue was coming from the copy to clipboard side which was overriding `display:none;`

daragh-king-genesys (approved):

PR #986, opened 2025-03-13
fix(button): Icon examples clickable in NVDA

321gillian:
COMUI-3510

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-button/example.html
+++ b/packages/genesys-spark-components/src/components/stable/gux-button/example.html
@@ -1,5 +1,9 @@
 <h1>gux-button</h1>
-
+<p>
+  Due to a known accessibility issue, we now recommend using
+  <a href="gux-button-slot.html">gux-button-slot</a> in the case of
+  <b>icon only</b> buttons.
+</p>
 <h2>Enabled</h2>
 <h3>Text Only</h3>
 <div class="container">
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-button/example.html
```diff
@@ -17,46 +17,46 @@ <h3>Text Only</h3>
...
-    ><gux-icon
-      icon-name="fa/diamond-regular"
-      screenreader-text="This will be read by a screen reader"
-    ></gux-icon
+  <gux-button onclick="notify(event)">
+    <gux-screen-reader-beta>
+      This will be read by a screen reader
+    </gux-screen-reader-beta>
+    <gux-icon icon-name="fa/diamond-regular" decorative></gux-icon
   ></gux-button>
```
321gillian, line 25: I'm not sure whether it makes more sense in this scenario to use provide a way to set the button aria-label rather than introducing another slotted component. We have `guxTitle` prop but I know we've looked at phasing that out, also don't know if tooltips will always be needed. 

It could be something like `<gux-button gux-title="This  will be read by a screenreader" tooltip-disabled="false"><gux-icon icon-name="fa/diamond-regular" decorative><gux-icon></gux-button>`

Suggestions welcome

katie-bobbe-genesys:
I would be curious to hear more about this issue at collaboration time tomorrow. I need to set up my machine with NVDA to test out this bug. 

If this is an issue that may affect more than just the `gux-button`, I'm wondering if would we want to add the ` <gux-screen-reader-beta>` sub component within the `gux-icon` itself.

321gillian:
https://github.com/user-attachments/assets/8b446bb3-c15b-4711-aa2e-386b928b3e7f

Adding a video example here to show what happens before we discuss

321gillian:
Based on some further comments the new version still has some issue in Browse mode on NVDA and so an alternative approach may be required. In this case, I think something which uses the aria label might work but I'll update when ready

packages/genesys-spark-components/src/components/stable/gux-button/example.html
```diff
@@ -1,6 +1,94 @@
...
+  Note: Not possible to set aria label on the button within gux-button without
+  using gux-title currently
+</p>
+<p><b>NVDA: </b><i>clickable button Gux-button with icon using gux-title</i></p>
+<p><b>Clickable?</b> Yes</p>
+<gux-button
+  onclick="notify(event)"
+  gux-title="Gux-button with icon using gux-title"
+  ><gux-icon icon-name="fa/diamond-regular" decorative></gux-icon
+></gux-button>
```
321gillian, line 51: This approach provides the clearest text to screenreader and is clickable but gux-title isn't recommended.

katie-bobbe-genesys:
I was doing some research into some of the button tickets that came up, and was wondering if maybe we should just promote the usage of `gux-button-slot` instead of `gux-button`. The icon only button examples work as expected for `gux-button-slot`. Also, the `aria-label` behavior works as expected when using `gux-button-slot`, which addresses [this ticket ](https://inindca.atlassian.net/browse/COMUI-3730)

321gillian:
> I was doing some research into some of the button tickets that came up, and was wondering if maybe we should just promote the usage of `gux-button-slot` instead of `gux-button`. The icon only button examples work as expected for `gux-button-slot`. Also, the `aria-label` behavior works as expected when using `gux-button-slot`, which addresses [this ticket ](https://inindca.atlassian.net/browse/COMUI-3730)

That could work. In that case, should all of the icon button examples be removed from `gux-button` and replaced with a reference to `gux-button-slot`

katie-bobbe-genesys (approved):

PR #992, opened 2025-03-20
fix(time-zone-beta): fix translations

321gillian:
COMUI-3550


packages/genesys-spark-components/src/genesys-spark-utils/intl.ts
```diff
@@ -95,7 +95,7 @@ function browserHasRegionData(localeString: string): boolean {
 // Currently, login page and web-directory store the English user selection as `en-us`.
 // We will remove this override once those apps migrate from using en-us to en as part of a future epic.
 function domLocaleOverride(localeString: string): string {
-  if (localeString.toLowerCase() === 'en-us') {
+  if (localeString?.toLowerCase() === 'en-us') {
```
jason-evans-genesys, line 98: Curious, if localeString is null then is there an issue higher up in the chain? Wondering the scenario when localeString is empty and passed into this function.

321gillian: This was just a thing that happened when I ran `npm run build`. I assume it's linting related.

daragh-king-genesys (approved):

PR #993, opened 2025-03-20
fix(popover-beta): add method to close popover

gavin-everett-genesys:
Open to opinions on this change and also if approved do we think we should apply this to the stable component also or just have it as a beta feature ?

✅ Closes: COMUI-3552

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-popover-beta/gux-popover.tsx
+++ b/packages/genesys-spark-components/src/components/beta/gux-popover-beta/gux-popover.tsx
@@ -8,7 +8,8 @@ import {
   Listen,
   Prop,
   Watch,
-  State
+  State,
+  Method
 } from '@stencil/core';
 import {
   autoUpdate,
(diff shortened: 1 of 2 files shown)


daragh-king-genesys:
I think we might need to change the internal `dismiss` function as it calls `togglePopover`. This might cause an issue as we can no longer guarantee that dismiss will not be called when the popover is closed. I think there is a `hidePopover`method that we can use but we should test to make sure it works as expected.

daragh-king-genesys:
If just adding it to the beta in enough for the dev that opened the ticket we can go with that for now.

gavin-everett-genesys:
> I think we might need to change the internal `dismiss` function as it calls `togglePopover`. This might cause an issue as we can no longer guarantee that dismiss will not be called when the popover is closed. I think there is a `hidePopover`method that we can use but we should test to make sure it works as expected.

Yeah its just for the `beta` component. I looked into the `hidePopover()` method and it works fine and makes more sense in this case since the method is just for dismissing the popover 👍 . `togglePopover()` probably could cause an issue if the popover was already hidden and it was called it would show it instead of hiding it and the states would probably go out of sync.

daragh-king-genesys (approved):

daragh-king-genesys (approved):
We may need to add this to `gux-popover-list-beta`

gavin-everett-genesys:
> We may need to add this to `gux-popover-list-beta`

Will create a ticket for this 👍

PR #995, opened 2025-03-21
test(copy-to-clipboard): add exclusion

321gillian:


packages/genesys-spark-components/src/components/stable/gux-copy-to-clipboard/tests/gux-copy-to-clipboard.e2e.ts
```diff
@@ -1,5 +1,14 @@
 import { newSparkE2EPage, a11yCheck } from '../../../../test/e2eTestUtils';
 
+const axeExclusions = [
+  {
+    issueId: 'button-name',
+    target: 'gux-copy-to-clipboard,button',
+    exclusionReason:
+      'Description is provided by tooltip which is not detected by Axe as it has display: none set. Manually confirmed to work on VoiceOver & NVDA.'
```
daragh-king-genesys, line 8: Would it be better if we gave the button screen reader only text and hide the tooltip from the sceenreader?

321gillian: @daragh-king-genesys 

I was looking into this. I think in order to do this I'd need to swap out `gux-tooltip` for `gux-tooltip-beta`, unless ?I add the visual only prop to `gux-tooltip`. 

If I do swap them out, I can't see a clear way to style content within `gux-tooltip-beta` with `display: inline-flex` as it doesn't have the `gux-show` class on the host (it's on `gux-tooltip-base-beta`). I could try and bring the base tooltip into this component?

If I have a hideable styleable tooltip within `gux-copy-to-clipboard` and use the `gux-screenreader-beta` to add the button, then I was thinking of maybe using `gux-announce` to indicate the copy activity. 

The announcement on initial focus then would be something like 'jane.doe@example.com Copy to Clipboard' and on click 'jane.doe@example.com Copy to Clipboard, Copied to Clipboard'. That seems worse that what we have at the minute. 

Any ideas?

daragh-king-genesys:
https://github.com/MyPureCloud/genesys-spark/pull/1010

PR #996, opened 2025-03-24
chore(other): added hi lang to tests and examples

daragh-king-genesys:
added hi lang to tests and examples


packages/genesys-spark-chart-components/src/components/beta/gux-visualization/tests/d3-rs-intl.spec.ts
```diff
@@ -988,6 +988,97 @@ describe('@redsift/d3-rs-intl', () => {
...
+              'Aug',
+              'Sep',
+              'Oct',
+              'Nov',
+              'Dec'
+            ],
+            time: '%-I:%M:%S %p'
+          },
+          months: [
+            'January',
```
jason-evans-genesys, line 1050: Would it make sense to use month array variables instead of copy and pasting these months/days or do you think it's fine as is because it's a test file?

321gillian (approved):

PR #997, opened 2025-03-24
feat(modal): added inaccessibleFocusTrap option

daragh-king-genesys:
added inaccessibleFocusTrap option

✅ Closes: COMUI-3524


docs/migration-guides/v4/gux-modal-legacy.md
```diff
@@ -57,6 +57,7 @@ Steps:
 - There are two mechanisms for showing/hiding modals in v4, to support both of the common approaches we see.
   - You can render the modal, and use the `showModal` and `close` methods on the `gux-modal` element to show and hide it.
   - Alternatively, if you only want to render the modal when visible, add the `open` attribute to have it render in an initially open state
+- If you were relying on the focus trap that was present in v3 in an inaccessible way you can use the new `inaccessible-focus-trap` attribute to get that functionality in v4. This will be an accessibility violation and you should look to fix the UX of your app so that it is not required ASAP.
```
thomas-c-dillon, line 60: Didn't we all discuss and agree that we were not doing this? What did I miss?

MattCheely:
I don't think we should pursue this approach. The APIs already exist in our micro-frontend architecture for handling cross-frame modals. One thing I've seen repeatedly is that if we add an easy to enable feature to a shared library as a stopgap, people will assume it's a correct/blessed approach and adopt it over solutions that provide a better and more accessible user experience. Often, they won't even discover that those other options exist, because they can immediately put hands on something that looks like it works at a first pass. 

The existing APIs should be improved if they aren't meeting teams needs (although I suspect most of the issue is awareness). If we enable this workaround, I expect that we'll never get the feedback we need to make those improvements as teams will silently adopt this change until it becomes a critical issue for a customer, and then we'll have to unravel all of that adoption.

PR #998, opened 2025-03-24
feat(other): update global navigation tokens

daragh-king-genesys:
update global navigation tokens

✅ Closes: COMUI-3537


VirtualClover (changes requested):
Code-side looks good BUT there are some duplicated variables in Figma that need to be solved before merging this 
cc. @samlimby-genesys 
image

VirtualClover (approved):
Figma issues fixed :)
We can proceed with the merge

PR #1003, opened 2025-03-26
chore(form-field-dropdown): Fixed input focus after expanded

jason-evans-genesys:
✅ Closes: COMUI-3412

**Update:** I closed this PR and opened this one below to implement a different solution: https://github.com/MyPureCloud/genesys-spark/pull/1024

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-dropdown/a11yManualChecklist.md
+++ b/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-dropdown/a11yManualChecklist.md
@@ -1,13 +1,13 @@
 # gux-form-field-dropdown manual accessibility testing status
 
-**Last Updated:** 2025-01-13T17:15:53.348Z
+**Last Updated:** 2025-03-28T17:24:43.556Z
 | Pass | WCAG Success Criterion | Notes |
 | --- | --- | --- |
 | ✅ | [1.1.1 Non-text Content](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content.html) | - |
 | ✅ | [2.1.1 Keyboard](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html) | - |
 | ✅ | [2.1.2 No Keyboard Trap](https://www.w3.org/WAI/WCAG22/Understanding/no-keyboard-trap.html) | - |
 | ✅ | [2.4.3 Focus Order](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html) | - |
-| ❌ | [2.4.7 Focus Visible](https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html) | [COMUI-3412](https://inindca.atlassian.net/browse/COMUI-3412): Focus ring is not visible when opening the dropdown initially but it is activated when arrow kets are pressed. |
+| ✅ | [2.4.7 Focus Visible](https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html) | - |
 | ✅ | [2.4.11 Focus Not Obscured](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum) | - |
 | ✅ | [2.5.3 Label in Name](https://www.w3.org/WAI/WCAG22/Understanding/label-in-name.html#dfn-name) | - |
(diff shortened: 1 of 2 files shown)


daragh-king-genesys:
Is part of the PR missing? There are only changes to the example page currently in the PR.

jason-evans-genesys:
> Is part of the PR missing? There are only changes to the example page currently in the PR.

@daragh-king-genesys It seemed like this was an implementation issue of the component. There's specific code that checks if the component is filterable (i.e. `filter-type` prop is set to a specific value) in order to trigger focus properly.

katie-bobbe-genesys (changes requested):
The `filter-type` prop should not be a requirement for this component to be accessible. It's an optional prop. I think there needs to be a different solution for this.

jason-evans-genesys:
> The `filter-type` prop should not be a requirement for this component to be accessible. It's an optional prop. I think there needs to be a different solution for this.

@katie-bobbe-genesys great catch. I'll put in another solution fix.

jason-evans-genesys:
@daragh-king-genesys @katie-bobbe-genesys I looked into this issue more and I think what's going on is that when you don't have 'filter-type="starts-with"' set on the dropdown or dropdown-multi components then internally those components will not add a focus ring on the input, which I think is correct because it seems like we only want a focus ring on the input if the user can actually type in text to filter the dropdown options. I noticed that the dropdown component example page will not add a focus ring on the input within the `Value Set` section because there's nothing for the user to type in that input to filter the dropdown options. So I think all we have to do is run the ally manual checklist for this PR and note on the ticket that the component is working as intended and how there should not be a focus ring unless if it's filterable. Thoughts?

katie-bobbe-genesys:
So I think that there might need to be a change in the code here to resolve this issue. Take the following example code and copy this into the docs site:
```
  <gux-dropdown>
    <gux-listbox aria-label="Animals">
      <gux-option value="a" disabled>Ant</gux-option>
      <gux-option value="b">Bat</gux-option>
      <gux-option value="c">Cat</gux-option>
      <gux-option value="d">Dog</gux-option>
      <gux-option value="e" disabled>Eel</gux-option>
      <gux-option value="f">Frog</gux-option>
      <gux-option value="g">Goat</gux-option>
      <gux-option value="h" disabled>Horse</gux-option>
      <gux-option value="i" disabled>Ibis</gux-option>
    </gux-listbox>
  </gux-dropdown>
```

Using the keyboard, interact with the dropdown in the following ways:
1. Tab to the dropdown and open it using the `down arrow`. You will see that focus is moved into the dropdown list and the focus ring is visible around the option.
2. Tab to the dropdown and open it using `space` or `enter`.  Focus is not moved into the dropdown list and there is no focus ring visible around the option.

I believe a complete fix to this issue would be to make the `space` and `enter ` key function the same as pressing the `down arrow` when opening the dropdown. And when doing this, test that there are no regressions for filterable dropdowns focus ring functionality

jason-evans-genesys:
> So I think that there might need to be a change in the code here to resolve this issue. Take the following example code and copy this into the docs site:
> 
> ```
>   <gux-dropdown>
>     <gux-listbox aria-label="Animals">
>       <gux-option value="a" disabled>Ant</gux-option>
>       <gux-option value="b">Bat</gux-option>
>       <gux-option value="c">Cat</gux-option>
>       <gux-option value="d">Dog</gux-option>
>       <gux-option value="e" disabled>Eel</gux-option>
>       <gux-option value="f">Frog</gux-option>
>       <gux-option value="g">Goat</gux-option>
>       <gux-option value="h" disabled>Horse</gux-option>
>       <gux-option value="i" disabled>Ibis</gux-option>
>     </gux-listbox>
>   </gux-dropdown>
> ```
> 
> Using the keyboard, interact with the dropdown in the following ways:
> 
>     1. Tab to the dropdown and open it using the `down arrow`. You will see that focus is moved into the dropdown list and the focus ring is visible around the option.
> 
>     2. Tab to the dropdown and open it using `space` or `enter`.  Focus is not moved into the dropdown list and there is no focus ring visible around the option.
> 
> 
> I believe a complete fix to this issue would be to make the `space` and `enter ` key function the same as pressing the `down arrow` when opening the dropdown. And when doing this, test that there are no regressions for filterable dropdowns focus ring functionality

Thanks @katie-bobbe-genesys for the help! I'll try your suggestions.

PR #1006, opened 2025-03-27
fix(rich-text-editor-beta): container now inherits size from host

gavin-everett-genesys:
container now inherits size from host

[✅ Closes: COMUI-3568](https://inindca.atlassian.net/browse/COMUI-3568)

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor.scss
+++ b/packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor.scss
@@ -1,5 +1,7 @@
 :host {
   display: block;
+  inline-size: 100%;
+  block-size: 100%;
 }
 
 .gux-rich-text-editor-container {
@@ -7,8 +9,8 @@
   display: flex;
   flex-direction: column;
   gap: var(--gse-ui-rte-innerContainer-gap);
-  inline-size: 100%;
-  max-inline-size: 100%;
+  inline-size: inherit;
+  max-inline-size: inherit;
   padding: var(--gse-ui-rte-padding);
   overflow: auto;
   background-color: var(--gse-ui-formControl-input-backgroundColor);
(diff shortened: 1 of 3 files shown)


packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor.scss
```diff
@@ -1,14 +1,16 @@
...
 
 .gux-rich-text-editor-container {
   box-sizing: border-box;
   display: flex;
   flex-direction: column;
   gap: var(--gse-ui-rte-innerContainer-gap);
-  inline-size: 100%;
-  max-inline-size: 100%;
+  inline-size: inherit;
+  max-inline-size: inherit;
```
daragh-king-genesys, line 13: I don't think this is right. If the host is set to `50%` then the container will be `50%` of `50%`.

I thought this change was going to be 

```diff
inline-size: 100%;
max-inline-size: 100%;
+  block-size: 100%;
+  max-block-size: 100%;
```
and no changes on the host.

gavin-everett-genesys: Will look into it

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor.scss
```diff
@@ -2,15 +2,22 @@
   display: block;
 }
 
+:host([style*='resize']) {
+  overflow: auto;
+  resize: inherit;
+}
+
```
gavin-everett-genesys, line 9: @daragh-king-genesys What are your thoughts on this and do you see any downsides to this approach. After looking at various options I think this is okay. My initial thought was to have ,

`:host {
resize:inherit;
overflow:auto;
`
but I dont think this is ideal as it would `inherit` if it was a child of a parent container which I am not sure is desirable.

With the current change it works if they set it on the element itself like 

`gux-rich-text-editor-beta style="resize:both; inline-size:500px;` for example.

Let me know what you think.

daragh-king-genesys (approved):

PR #1012, opened 2025-04-01
fix(toast): add ability to hide the dismiss button

gavin-everett-genesys:
New property added `hideDismissButton`.

As far as I can see there should be no accessibility concerns here.

[✅ Closes: COMUI-3587](https://inindca.atlassian.net/browse/COMUI-3587)

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-toast/gux-toast.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-toast/gux-toast.tsx
@@ -32,6 +32,9 @@ export class GuxToast {
   @Prop()
   toastType: GuxToastTypes = 'success';
 
+  @Prop()
+  dismissible: boolean = true;
+
   @Event()
   guxdismiss: EventEmitter<void>;
 
@@ -46,9 +49,6 @@ export class GuxToast {
 
   @State()
   hasSecondaryButton: boolean = false;
-
-  @Prop()
-  hideDismissButton: boolean = false;
 
   componentWillLoad(): void {
     trackComponent(this.root, { variant: this.toastType });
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-toast/gux-toast.tsx
```diff
@@ -47,6 +47,9 @@ export class GuxToast {
   @State()
   hasSecondaryButton: boolean = false;
 
+  @Prop()
+  hideDismissButton: boolean = false;
+
```
daragh-king-genesys, line 52: Keep the @Props together in the file

daragh-king-genesys, line 52: What do you think of
```
 @Prop()
  dismissible: boolean = true;
```

gavin-everett-genesys: Yeah I dont mind, I tried to keep it consistent with the `popover` api since its similar. The `popover` has `displayDismissButton`.

daragh-king-genesys: Ya I just don't like them long names that are referencing elements instead of functionality.
Maybe changing `displayDismissButton` is something we should look int for v5

daragh-king-genesys:
I am happy with this now but would prefer a few more eyes on the api before we merge it. We can bring it up during collab time.

katie-bobbe-genesys:
What do you think about only allowing this prop on action toasts? Looking at the original ticket, it looks like this is needed so that users are made to interact with one of the action buttons.

gavin-everett-genesys:
> What do you think about only allowing this prop on action toasts? Looking at the original ticket, it looks like this is needed so that users are made to interact with one of the action buttons.

Yeah I agree with this, makes sense as the original ticket ask was specifically action toasts and I believe a standard toast should probably always have the dismiss button 👍  But then I am thinking would it not be consistent to do this 🤔

gavin-everett-genesys:
Closing for now as I have heard nothing from the team that requested this in the past month.

PR #1023, opened 2025-04-10
fix(tooltip): update examples of how tooltips work with buttons

gavin-everett-genesys:
update examples of how tooltips work with buttons

[✅ Closes: COMUI-3569](https://inindca.atlassian.net/browse/COMUI-3569)

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/example.html
+++ b/packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/example.html
@@ -41,21 +41,19 @@
     <div class="flex-example">
       <h4 class="flex-header">light</h4>
       <div class="flex-content">
-        <gux-button
-          >Hover<gux-tooltip-beta accent="light">
-            <div slot="content">Tooltip with light accent</div>
-          </gux-tooltip-beta></gux-button
-        >
+        <gux-button id="tooltip-light-accent">Hover</gux-button>
+        <gux-tooltip-beta for="tooltip-light-accent" accent="light">
+          <div slot="content">Tooltip with light accent</div>
+        </gux-tooltip-beta>
       </div>
     </div>
     <div class="flex-example">
       <h4 class="flex-header">dark</h4>
       <div class="flex-content">
-        <gux-button
-          >Hover<gux-tooltip-beta accent="dark">
-            <div slot="content">Tooltip with dark accent</div>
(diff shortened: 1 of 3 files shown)


packages/genesys-spark-components/src/components/stable/gux-button/gux-button.scss
```diff
@@ -68,7 +68,6 @@ button {
   }
 
   &[disabled] {
-    pointer-events: none;
```
gavin-everett-genesys, line 71: Removed in order for the tooltip to show when the button is `disabled`. Is there any context in which this is would be required. I dont think emits events as long as we have `pointer-events:none` on the host.

packages/genesys-spark-components/src/components/stable/gux-tooltip/example.html
```diff
@@ -39,18 +39,17 @@ <h3>accent</h3>
...
+          <div slot="content">Tooltip with light accent</div>
+        </gux-tooltip>
       </div>
     </div>
     <div class="flex-example">
       <h4 class="flex-header">dark</h4>
       <div class="flex-content">
-        <gux-button
-          >Hover<gux-tooltip accent="dark">
+        <gux-button disabled
```
jason-evans-genesys, line 51: Thoughts on having something specific for a disabled state in the example page instead of making the dark accent disabled? Maybe a separate disabled section or something?

packages/genesys-spark-components/src/components/beta/gux-tooltip-beta/example.html
```diff
@@ -189,6 +187,18 @@ <h3>showTooltip</h3>
     <p>Show the tooltip if it is currently hidden.</p>
   </section>
 
+  <h2>Using a gux-button with a gux-tooltip-beta</h2>
+  <gux-button id="tooltip-light">Hover</gux-button>
+  <gux-tooltip-beta for="tooltip-light" accent="light">
+    <div slot="content">Tooltip with light accent</div>
+  </gux-tooltip-beta>
+
+  <h3>Disabled gux-button with gux-tooltip</h3>
```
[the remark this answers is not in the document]
gavin-everett-genesys: Yeah we agreed this with Matt the previous week to not have the `tooltip` slotted but instead just to have it done via the `for` attribute. If you have have concerns we can go through them.

PR #1024, opened 2025-04-10
chore(dropdown): Focus listbox option on enter or space key press

jason-evans-genesys:
✅ Closes: COMUI-3412

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi.tsx
@@ -212,6 +212,15 @@ export class GuxDropdownMulti {
         if (this.canCreateNewOption() && this.isActiveElement()) {
           this.emitCreateOption();
         }
+        if (!this.expanded && !this.isFilterable()) {
+          setInitialActiveOption(this.listboxElement);
+        }
+        return;
+      case ' ':
+        if (!this.expanded && !this.isFilterable()) {
+          setInitialActiveOption(this.listboxElement);
+        }
+        return;
     }
   }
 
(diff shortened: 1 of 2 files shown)


katie-bobbe-genesys (changes requested):
There needs to be logic added so that 'setInitialActiveOption' is not triggered if focus is within the listbox
For `dropdown`:
Focus to initial active option works as expected, but when selecting an option using `space` or `enter` focus is briefly moved back to the initial active option before closing the dropdown, which is communicated to the screen reader and confusing
For `dropdown-multi`:
keyboard navigation does not work as expected. I would suggest testing this functionality in dropdown multi using the keyboard by trying to select multiple options. You'll see that focus keeps being redirected to the "initial active item"

jason-evans-genesys:
> There needs to be logic added so that 'setInitialActiveOption' is not triggered if focus is within the listbox For `dropdown`: Focus to initial active option works as expected, but when selecting an option using `space` or `enter` focus is briefly moved back to the initial active option before closing the dropdown, which is communicated to the screen reader and confusing For `dropdown-multi`: keyboard navigation does not work as expected. I would suggest testing this functionality in dropdown multi using the keyboard by trying to select multiple options. You'll see that focus keeps being redirected to the "initial active item"

Great finds! Pushed fixes for this.

packages/genesys-spark-components/src/components/stable/gux-dropdown-multi/gux-dropdown-multi.tsx
```diff
@@ -212,6 +212,15 @@ export class GuxDropdownMulti {
         if (this.canCreateNewOption() && this.isActiveElement()) {
           this.emitCreateOption();
         }
+        if (!this.expanded) {
```
katie-bobbe-genesys, line 215: In addition to `not expanded`, I think `not filterable` should also be added here for both dropdown multi and dropdown components
```
if (!this.expanded && !this.filterable)
```

jason-evans-genesys: Fixed.

katie-bobbe-genesys (approved):

PR #1031, opened 2025-04-11
fix(tooltip): update offset from 16px to 12px

katie-bobbe-genesys:
UX requested we update the offset value of tooltips globally from 16px to 12px https://inindca.atlassian.net/browse/COMUI-3634


daragh-king-genesys (approved):
Left a related comment on https://inindca.atlassian.net/browse/COMUI-2480

PR #1036, opened 2025-04-15
fix(toast): Break word on message

321gillian:
✅ Closes: COMUI-3635

To see the issue add this to docs:
```
<gux-toast toast-type="success">
  <div slot="title">Success Example</div>
  <div slot="message">accountswitcherdcatestuser2@account-switcherdca.test account successfully removed</div>
</gux-toast>
```

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-toast/gux-toast.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-toast/gux-toast.scss
@@ -71,6 +71,7 @@
         font-family: var(--gse-ui-toast-text-fontFamily);
         font-size: var(--gse-ui-toast-text-fontSize);
         line-height: var(--gse-ui-toast-text-lineHeight);
+        word-break: break-all;
       }
     }
 


packages/genesys-spark-components/src/components/stable/gux-toast/gux-toast.scss
```diff
@@ -71,6 +71,7 @@
         font-family: var(--gse-ui-toast-text-fontFamily);
         font-size: var(--gse-ui-toast-text-fontSize);
         line-height: var(--gse-ui-toast-text-lineHeight);
+        word-break: break-all;
```
jordanstith15, line 74: What's your opinion on using "break-word" vs "break-all"? "break-word" seems to keep the word together if it can fit on the next line, but breaks it if it doesn't.
https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

On another note, I love how the mdn docs have asian languages in their examples. They used to cause some annoying issues before lol

321gillian: I would have gone for `break-word` except that it seemed to be marked as deprecated on the docs? Do you think it's still ok?

jordanstith15: Oh I I didn't see that. They really put "experimental" and "deprecated" in a place that's easy to miss

jordanstith15 (approved):

jason-evans-genesys (approved):

jason-evans-genesys:
@321gillian I approved but it looks like there's a static code analysis failing. Maybe we can re-reun that check to see if it passes the next run?

packages/genesys-spark-components/src/components/stable/gux-toast/gux-toast.scss
321gillian: They really do!

PR #1043, opened 2025-04-16
perf(form-field): announcing error states [Feedback]

gavin-everett-genesys:
Looking for feedback in general on this change and if we should go ahead with doing this change. I did a bit of research around this also and will outline below my thoughts.

If someone could test this change in `NVDA` and `JAWS` that would be appreciated I cannot login to my AWS workspace so I am trying to get it resolved.

Links to research

- https://www.smashingmagazine.com/2023/02/guide-accessible-form-validation/#invalid-fields
- https://bitsofco.de/using-aria-live/
- https://web.dev/learn/forms/accessibility
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-invalid

My first question was should this be something we should be baking into the components or is this something we should just provide guidance on through our documentation. For this PR I implemented a baked in version but I think there may be cons to this approach which I will discuss.

**Types of Validation**
There are different types of validation that can be used and I am unsure this is consistent across the whole genesys cloud platform. The types of validation are,

- Instant Validation - Where the validation occurs on each keypress which would mean that the error would be announced each keypress which would be undesirable and I am sure not a lot of teams use this but just to take note of it.
- Afterwards Validation - Which typically occurs on focus out or on blur which as far as I know is the most used within genesys cloud and would probably be suitable for this use case.
- Submit Validation - Which occurs when you click the submit button. The problem I have with this is if you click submit and 5 fields go to error states what would be announced ( as far as I know its just the first one )  and how will the screen-reader know what to do ?

**Announcing Dynamic Content Changes**
From what I read in the articles and documentation above the correct way to announce dynamic changes in the case of an error dynamically appearing through validation is using `aria-live` which we suspected initially.

**Utilising aria-invalid**
I think we should be utilising `aria-invalid` to identify when invalid data has been passed to the input this is seen as a best practice when used in conjunction with `aria-live`. Voiceover will announce `invalid data` when focused on an input with and error state or invalid data.

**Problems identified**
So I am unsure if this is just a voiceover thing or not but for example if you tab to the first form-field in the example file then than to the next the error will display on the first form-field but the error is not announced by voiceover. But if you are to say tab to the first form-field then click away somewhere else the error will be announced. This is due to how the browser and voiceover works if you tab from one form control to another even if the error shows on the first input it wont be announced as the voiceover screenreader will be processing the new form control instead and will basically ignore the aria-live.

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-number/example.html
+++ b/packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-number/example.html
@@ -5,219 +5,55 @@
 <h2>Examples</h2>
 <form onchange="notify(event)" oninput="notify(event)">
   <fieldset>
-    <legend>Clearable</legend>
-
-    <gux-form-field-number>
-      <input slot="input" type="number" name="c-1" value="10" />
-      <label slot="label">Default</label>
-    </gux-form-field-number>
-
-    <gux-form-field-number clearable>
-      <input slot="input" type="number" name="c-2" value="10" />
-      <label slot="label">True</label>
-    </gux-form-field-number>
-
-    <gux-form-field-number clearable>
-      <input slot="input" type="number" name="c-3" value="10" disabled />
-      <label slot="label">True and disabled</label>
-    </gux-form-field-number>
-
-    <gux-form-field-number clearable="false">
-      <input slot="input" type="number" name="c-4" value="10" />
-      <label slot="label">False</label>
-    </gux-form-field-number>
-  </fieldset>
-
-  <fieldset>
-    <legend>Label Position</legend>
-
-    <gux-form-field-number>
-      <input slot="input" type="number" name="lp-1" value="10" />
-      <label slot="label">Default</label>
-    </gux-form-field-number>
-
-    <gux-form-field-number>
-      <input slot="input" type="number" name="lp-2" value="10" />
-      <label slot="label">Default long label</label>
-    </gux-form-field-number>
-
-    <gux-form-field-number label-position="above">
-      <input slot="input" type="number" name="lp-3" value="10" />
-      <label slot="label">Above</label>
-    </gux-form-field-number>
-
-    <gux-form-field-number label-position="beside">
-      <input slot="input" type="number" name="lp-4" value="10" />
-      <label slot="label">Beside</label>
-    </gux-form-field-number>
-
-    <gux-form-field-number label-position="screenreader">
-      <input slot="input" type="number" name="lp-5" value="10" />
-      <label slot="label">Screenreader</label>
-    </gux-form-field-number>
-  </fieldset>
-
-  <fieldset>
-    <legend>Input attributes</legend>
-
-    <gux-form-field-number>
-      <input slot="input" type="number" name="a-1" value="10" disabled />
-      <label slot="label">Disabled</label>
-    </gux-form-field-number>
-
-    <gux-form-field-number>
-      <input slot="input" type="number" name="a-2" value="10" required />
-      <label slot="label">Required</label>
-    </gux-form-field-number>
-
-    <gux-form-field-number>
-      <input slot="input" type="number" name="a-3" placeholder="Enter number" />
-      <label slot="label">Placeholder</label>
-    </gux-form-field-number>
-
-    <gux-form-field-number>
-      <input
-        slot="input"
-        type="number"
-        name="a-2"
-        value="10"
-        step="5"
-        min="-25"
-        max="25"
-      />
-      <label slot="label">Step / Min / Max</label>
-    </gux-form-field-number>
-  </fieldset>
-
-  <fieldset>
     <legend>Error</legend>
 
-    <gux-form-field-number>
+    <gux-form-field-number id="numberField">
       <input slot="input" type="number" name="e-1" value="10" />
       <label slot="label">Default</label>
-      <span slot="error">This is an error message</span>
     </gux-form-field-number>
-  </fieldset>
-
-  <fieldset>
-    <legend>Help</legend>
 
     <gux-form-field-number>
       <input slot="input" type="number" name="e-1" value="10" />
       <label slot="label">Default</label>
-      <span slot="help">This is a help message</span>
     </gux-form-field-number>
   </fieldset>
 
-  <fieldset>
-    <legend>Label Info</legend>
+  <style
+    onload="(function () {
 
-    <gux-form-field-number>
-      <input slot="input" type="number" name="e-1" value="10" />
-      <label slot="label">Default</label>
-      <gux-label-info-beta slot="label-info">
-        <span slot="content">This is some tooltip text</span>
-      </gux-label-info-beta>
-    </gux-form-field-number>
-  </fieldset>
 
-  <fieldset>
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-number/example.html
```diff
@@ -4,220 +4,56 @@ <h1>gux-form-field-number</h1>
 
 <h2>Examples</h2>
 <form onchange="notify(event)" oninput="notify(event)">
-  <fieldset>
-    <legend>Clearable</legend>
-
-    <gux-form-field-number>
-      <input slot="input" type="number" name="c-1" value="10" />
-      <label slot="label">Default</label>
```
gavin-everett-genesys, line 12: I will revert the changes in this file, I am just using it as an example so you dont have to make one yourself.

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-number/gux-form-field-number.tsx
```diff
@@ -215,9 +225,11 @@ export class GuxFormFieldNumber {
...
             )}
           </div>
-          <GuxFormFieldError show={this.hasError}>
-            <slot name="error" />
-          </GuxFormFieldError>
+          <div role="alert" aria-live="assertive">
+            <GuxFormFieldError show={this.hasError}>
+              <slot name="error" />
+            </GuxFormFieldError>
+          </div>
```
daragh-king-genesys, line 232: Could this be internal to `GuxFormFieldError`?

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-number/gux-form-field-number.tsx
```diff
@@ -158,6 +164,10 @@ export class GuxFormFieldNumber {
     this.hasError = hasSlot(this.root, 'error');
     this.hasHelp = hasSlot(this.root, 'help');
 
+    if (this.hasError) {
+      this.input.setAttribute('aria-invalid', 'true');
+    }
+
```
daragh-king-genesys, line 170: Would a @Watch on `hasError` work instead of repeating code like this in different functions?

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox-group/gux-form-field-checkbox-group.tsx
```diff
@@ -94,6 +94,13 @@ export class GuxFormFieldCheckboxGroupBeta {
     if (checkboxSlots) {
       checkboxSlots.forEach(item => {
         item.hasGroupError = hasGroupError;
+
+        const checkboxInput = item.querySelector('input');
+        if (hasGroupError) {
+          checkboxInput.setAttribute('aria-invalid', 'true');
+        } else {
+          checkboxInput.removeAttribute('aria-invalid');
+        }
```
daragh-king-genesys, line 103: Would it be better to handle this in the `gux-form-field-checkbox` component.
ie track the `hasGroupError` property in that component.

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox/gux-form-field-checkbox.tsx
```diff
@@ -48,6 +57,15 @@ export class GuxFormFieldCheckbox {
...
 
+  @Watch('hasError')
+  onHasError(hasError: boolean): void {
+    if (hasError) {
+      this.input.setAttribute('aria-invalid', 'true');
+    } else {
+      this.input.removeAttribute('aria-invalid');
+    }
+  }
+
```
daragh-king-genesys, line 68: Linked to previous comment: This is where we might have a watch 'hasGroupError' too.

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-radio-group/gux-form-field-radio-group.tsx
```diff
@@ -90,6 +90,12 @@ export class GuxFormFieldRadioGroupBeta {
     if (radioSlots) {
       radioSlots.forEach(item => {
         item.hasGroupError = hasGroupError;
+        const radioInput = item.querySelector('input');
+        if (hasGroupError) {
+          radioInput.setAttribute('aria-invalid', 'true');
+        } else {
+          radioInput.removeAttribute('aria-invalid');
+        }
```
daragh-king-genesys, line 98: Linked to previous comments: whatever you decide for checkbox group do here too.

packages/genesys-spark-components/src/components/beta/gux-phone-input/gux-phone-input.tsx
```diff
@@ -239,6 +248,10 @@ export class GuxPhoneInput {
   componentDidLoad(): void {
     this.setInput();
     this.setListBox();
+
+    if (this.hasError) {
+      this.inputElement.setAttribute('aria-invalid', 'true');
```
daragh-king-genesys, line 253: Are we sure we will not be causing issues here when we are changing attributes on the slotted inputs?

I am not sure what alternative we have to updating the attribute but maybe we will need to have some documentation letting devs know we will be changing this attribute etc....

321gillian:
https://github.com/user-attachments/assets/d8dcfe99-fc78-466f-92c8-00694816e6fe

NVDA

packages/genesys-spark-components/src/components/stable/gux-form-field/functional-components/gux-form-field-error/gux-form-field-error.tsx
```diff
@@ -10,6 +10,8 @@ export const GuxFormFieldError: FunctionalComponent<GuxFormFieldErrorProps> = (
 ): VNode => {
   return (
     <div
+      role="alert"
+      aria-live="assertive"
```
321gillian, line 14: Does this work without the `aria-live="assertive"`? Wondering because of the issue flagged in the alert section here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Guides/Live_regions#roles_with_implicit_live_region_attributes

gavin-everett-genesys: Yeah i removed the `aria-live="assertive"`. I read before if you use both it maximises compatibility but reading that it says it can also cause issues in voice over so I just removed it instead.

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-checkbox-group/gux-form-field-checkbox-group.tsx
gavin-everett-genesys: Now handling in the `checkbox` component instead.

gavin-everett-genesys:
Will work on the example page instead based on collab call.

PR #1048, opened 2025-04-16
fix(button): fixed legacy tertiary button color

daragh-king-genesys:
fixed legacy tertiary button color

✅ Closes: COMUI-3636


packages/genesys-spark-tokens/data/semantic/mode/light.json
```diff
@@ -1974,4 +1974,4 @@
       }
     }
   }
-}
```
jason-evans-genesys, line 1977: Are these line break changes intentional?

gavin-everett-genesys (approved):

PR #1054, opened 2025-04-22
chore(other): enable dark mode toggle switch on docs site

katie-bobbe-genesys:


gavin-everett-genesys (approved):

gavin-everett-genesys:
I was wondering should we add some accessible text to the button as it does not read out anything in voiceover ?

PR #1055, opened 2025-04-23
fix(dropdown-multi): add afterNextRender when slot content changes

gavin-everett-genesys:
**Steps to reproduce on the live docs**

1. Go to the `Unknown Value Set` dropdown where you add the Jellyfish option.
2. Set the `filter-type` to be `starts-with`
3. Click `Add Jellyfish Option`
4. You should see that the Jellyfish option does not appear in the `listbox-multi` list.

[✅ Closes: COMUI-3609](https://inindca.atlassian.net/browse/COMUI-3609)


gavin-everett-genesys:
Looking into issue with the build

PR #1064, opened 2025-04-29
fix(rich-text-editor-beta): remove gap when there is no toolbar

gavin-everett-genesys:
remove gap when there is no toolbar present

[✅ Closes: COMUI-3659](https://inindca.atlassian.net/browse/COMUI-3659)

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
+++ b/packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
@@ -16,6 +16,7 @@ import { getActionsFromGroup } from './gux-rich-text-editor.service';
 import { buildI18nForComponent, GetI18nValue } from 'i18n';
 import translationResources from './gux-rich-text-editor-action/i18n/en.json';
 import { afterNextRenderTimeout } from '@utils/dom/after-next-render';
+import { OnMutation } from '@utils/decorator/on-mutation';
 
 /**
  * @slot typographical-emphasis - Slot for typographical actions.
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
```diff
@@ -288,12 +298,25 @@ export class GuxRichTextEditor {
...
+  }
+
   render(): JSX.Element {
     return (
       <div
         class={{
           'gux-rich-text-editor-container': true,
-          'gux-disabled': this.disabled
+          'gux-disabled': this.disabled,
+          'gux-toolbar-hidden': !this.hasToolbar
```
gavin-everett-genesys, line 319: I could have just conditionally rendered the `gux-rich-text-editor-toolbar-container` aswell but this will affect all the resizing logic so instead of adding additional `if` guards around the code I think this is the least invasive way to just remove the gap. I have also tested dynamically addding elements to the toolbar and it works as expected.

packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
```diff
@@ -288,12 +298,25 @@ export class GuxRichTextEditor {
...
+  private hasToolbarChildren(): boolean {
+    const children = [
+      this.renderTypographicalEmphasis(),
+      this.renderTextStyling(),
+      this.renderListsIndentation(),
+      this.renderInserting(),
+      this.renderTextEditorMenu(),
+      this.renderGlobalAction()
+    ].filter(child => child !== null && child !== undefined);
+    return children.length > 0;
```
daragh-king-genesys, line 310: This is fine but you could also use [Array.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) here.

```
return [
      this.renderTypographicalEmphasis(),
      this.renderTextStyling(),
      this.renderListsIndentation(),
      this.renderInserting(),
      this.renderTextEditorMenu(),
      this.renderGlobalAction()
    ].some(child => child !== null && child !== undefined);
```

daragh-king-genesys (approved):

PR #1066, opened 2025-04-30
feat(tabs): apply the correct visual state on newly enabled or disabl…

conor-darcy:
…ed tabs

Correct visual style is applied if you toggle tabs disabled state on & off & select different tabs.


gavin-everett-genesys:
Can you explain more in detail of what this is fixing ? I just did a quick test of dynamically adding the `gux-disabled` attribute to a `gux-tab` but it still shows a blue underline when `disabled` which would indicate its `active` and `disabled` is this intended or did I miss a step ? I attached a video below,


https://github.com/user-attachments/assets/4a493af5-e79c-4205-a175-f0c084ef0f1e

conor-darcy:
@gavin-everett-genesys If a tab is enabled it doesn't get the blue bar when selected.
https://github.com/user-attachments/assets/e5aca403-7426-4ff7-aa36-b0cce978cc84

katie-bobbe-genesys:
The feature branch doesn't seem to be building, you might need to rebase on main?

conor-darcy:
@katie-bobbe-genesys Branch should build now.

PR #1069, opened 2025-05-02
feat(segmented-control-item): Add aria-current to selected item

321gillian:


packages/genesys-spark-components/src/components/beta/gux-segmented-control/gux-segmented-control-item/gux-segmented-control-item.tsx
```diff
@@ -114,6 +114,7 @@ export class GuxSegmentedControlItem {
             'gux-selected': this.selected
           }}
           type="button"
+          aria-current={this.selected}
```
katie-bobbe-genesys, line 117: I believe to work correctly `aria-current` needs to be set to the string `true`. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current#values

katie-bobbe-genesys (changes requested):

packages/genesys-spark-components/src/components/beta/gux-segmented-control/gux-segmented-control-item/gux-segmented-control-item.tsx
321gillian: Apologies, I had fixed this on my local but forgot to push it on Friday

katie-bobbe-genesys (approved):

PR #1073, opened 2025-05-06
fix(segmented-control-item): NVDA interactivity improvements

321gillian:
✅ Closes: COMUI-3562

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-segmented-control/example.html
+++ b/packages/genesys-spark-components/src/components/beta/gux-segmented-control/example.html
@@ -14,7 +14,7 @@
         icon-name="fa/align-left-regular"
         decorative
       ></gux-icon>
-      <div slot="text">Align left</div>
+      <span slot="text">Align left</span>
     </gux-segmented-control-item>
 
     <gux-segmented-control-item value="align-center" icon-only>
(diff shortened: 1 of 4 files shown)


packages/genesys-spark-components/src/components/beta/gux-segmented-control/example.html
```diff
@@ -14,7 +14,7 @@ <h2>Icon</h2>
         icon-name="fa/align-left-regular"
         decorative
       ></gux-icon>
-      <div slot="text">Align left</div>
+      <span slot="text">Align left</span>
```
daragh-king-genesys, line 17: Does this mean devs can't slot a `div`? I think that is an issue.

321gillian: I can't say for sure that they can't slot `divs` but it seems to create a difference in the way this component behaves with NVDA. 

`gux-segmented-control-item` in current version of docs in NVDA with Chrome: 
- Icon only and text versions of the item are announced as clickable by the screenreader
- Items can be navigated to using the button shortcut key
- Items **cannot** be clicked without manually changing from browse to focus mode

`gux-segmented-control-item` in updated version in NVDA with Chrome: 
- Icon only and text versions of the item are announced as clickable by the screenreader
- Items can be navigated to using the button shortcut key
- Items **can** be clicked without manually changing from browse to focus mode

PR #1082, opened 2025-05-09
fix(popover): Add tabindex -1 to gux-popover wrapper div

griffind:
Add tabindex -1 to gux-popover wrapper div so that it can be programmatically focused. This prevents dismissal on focus out and unexpected relatedTarget values.

✅ Closes: #3747


griffind:
Closes/fixes: https://inindca.atlassian.net/browse/COMUI-3747

PR #1089, opened 2025-05-15
chore(table-toolbar): Fixed layout scaled up

jason-evans-genesys:
✅ Closes: COMUI-3752

This PR fixes an issue where if you're in a condensed display layout and you increase the browser width, the permanent action buttons weren't moving from the menu actions dropdown to the root

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.tsx
@@ -87,7 +87,12 @@ export class GuxTableToolbar {
   }
 
   get permanentSlot(): Element | null {
-    return getSlot(this.root, 'permanent-actions');
+    // If permanent actions are not located at the root, then they are assumed to be located within the menu actions.
+    // Permanent actions will be located in the menu actions if the previous display layout was condensed.
+    return (
+      getSlot(this.root, 'permanent-actions') ||
+      getSlot(this.menuActionSlot as HTMLElement, 'permanent-actions')
+    );
   }
 
   get contextualSlot(): Element | null {


packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.tsx
```diff
@@ -87,7 +87,13 @@ export class GuxTableToolbar {
   }
 
   get permanentSlot(): Element | null {
-    return getSlot(this.root, 'permanent-actions');
+    const permanentActions = getSlot(this.root, 'permanent-actions');
+
+    // If permanent actions are not located at the root, then they are assumed to be located within the menu actions.
+    // Permanent actions will be located in the menu actions if the previous display layout was condensed.
+    return permanentActions
```
jason-evans-genesys, line 94: Honestly, I feel like a better solution would be more declarative instead of assuming a null state signifies the permanent actions reside within the menu action dropdown. Anyone have a more elegant solution for this? 

One idea is if I could know the previous display layout was condensed then I could declaratively determine that the permanent actions are currently located in the menu actions dropdown and could move it to the root. But I don't know if it's worth adding more display layout logic to track current and previous state. Thoughts?

packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.tsx
```diff
@@ -87,7 +87,13 @@ export class GuxTableToolbar {
...
+      ? permanentActions
+      : getSlot(this.menuActionSlot as HTMLElement, 'permanent-actions');
```
daragh-king-genesys, line 96: If I understand what is happening correctly this can be
``` js
get permanentSlot(): Element | null {
    // If permanent actions are not located at the root, then they are assumed to be located within the menu actions.
    // Permanent actions will be located in the menu actions if the previous display layout was condensed.
    return getSlot(this.root, 'permanent-actions') || getSlot(this.menuActionSlot as HTMLElement, 'permanent-actions');
}
```

jason-evans-genesys: Fixed

katie-bobbe-genesys (approved):

PR #1091, opened 2025-05-15
feat(dropdown) - set the initial tooltip placement to the right

gavin-everett-genesys:
https://inindca.atlassian.net/browse/COMUI-3750

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-tooltip/gux-tooltip.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-tooltip/gux-tooltip.tsx
@@ -131,7 +131,10 @@ export class GuxTooltip {
   private updatePosition(): void {
     const middleware = [
       offset(12),
-      flip(),
+      flip({
+        fallbackAxisSideDirection: 'start',
+        crossAxis: false
+      }),
       shift(),
       hide(),
       overflowDetection()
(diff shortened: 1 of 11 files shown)


packages/genesys-spark-components/src/components/stable/gux-tooltip/gux-tooltip.tsx
```diff
@@ -131,7 +131,10 @@ export class GuxTooltip {
   private updatePosition(): void {
     const middleware = [
       offset(12),
-      flip(),
+      flip({
+        fallbackAxisSideDirection: 'start',
+        crossAxis: false
```
gavin-everett-genesys, line 136: These settings were required as when the screen is small it was not detecting overflow. With these options added it will flip like so `[ 'right', 'left', 'top' , 'bottom']` . Notes are here  (https://github.com/floating-ui/floating-ui/pull/2082#issuecomment-1368384210)

Also on the docs website : https://floating-ui.com/docs/flip#fallbackaxissidedirection

gavin-everett-genesys: I dont know if we should apply this to the `tooltip-beta` as it does not really apply to it yet ?

daragh-king-genesys: I think we should keep the beta component in sync because something like this will be easily missed when we eventually transition `gux-truncate` to use `gux-tooltip-beta'.

daragh-king-genesys (commented):
LGTM just add it to the beta component too.

gavin-everett-genesys:
> My only concern here is that this might not work as expected for RTL languages?

I looked into this so it currently will still just flip to the right. I guess the intended would be to the left ? From the research I have done so far nothing has been implemented to support it regarding `placements` but there has been discussions https://github.com/floating-ui/floating-ui/issues/1530 . There is some workarounds but I am not sure if we want to go ahead using a workaround.  Ticket created for further discussion (https://inindca.atlassian.net/browse/COMUI-3764)

PR #1098, opened 2025-05-19
test(inline-alert): Playwright tests

321gillian:
✅ Closes: COMUI-3778


packages/genesys-spark-components/src/components/stable/gux-inline-alert/tests/gux-inline-alert.common.ts
```diff
@@ -0,0 +1,14 @@
+export const renderConfigs = [
+  {
+    html: `<gux-inline-alert accent="info"><div slot="content">Note: This is an information alert.</div></gux-inline-alert>`
```
daragh-king-genesys, line 3: I would add a description to each of these 
```
  description: 'render an "info" accent inline alert',
```

PR #1102, opened 2025-05-20
fix(calendar): added labels to month increment buttons

daragh-king-genesys:
added labels to month increment buttons


packages/genesys-spark-components/src/components/stable/gux-calendar/gux-calendar.tsx
```diff
@@ -547,6 +556,7 @@ export class GuxCalendar {
           <button
             type="button"
             class="gux-right"
+            aria-label={this.i18n('nextMonth')}
```
gavin-everett-genesys, line 559: In the `gux-calendar-beta` we have the `month` and `year` appended to this label eg `Next month : June 2025`. Should we do that here aswell ?

daragh-king-genesys: Maybe I do think that is better UX but I think what is currently there fixes the accessibility issue.
I could create a bug ticket to improve the text to match the beta component.

gavin-everett-genesys (approved):

PR #1111, opened 2025-05-22
test(flag-icon-beta): Add playwright tests

321gillian:
✅ Closes: COMUI-3779


packages/genesys-spark-components/src/test/playwrightTestUtils.ts
```diff
@@ -82,14 +89,15 @@ export async function setContent(page: E2EPage, html: string) {
 export async function checkRenders(
   renderConfigs: RenderConfig[],
   element?: string,
-  extraActions: ExtraActionsFn = () => Promise.resolve()
+  extraActions: ExtraActionsFn = () => Promise.resolve(),
+  performA11yCheck: boolean = true
```
daragh-king-genesys, line 93: Should we change this to be a single object instead of four (and probably more) arguments where only the first is required?
I find function calls like
```js
checkRenders([combinedFlagsConfig], 'gux-flag-icon', undefined, false);
```
hard to understand at a glance compared to 

```js
checkRenders({
  renderConfigs: [combinedFlagsConfig],
  element: 'gux-flag-icon',
  performA11yCheck: false
});
```

PR #1113, opened 2025-05-22
chore(icon): Updated brand icon svgs

jason-evans-genesys:
✅ Closes: COMUI-3792


gavin-everett-genesys:
Do these icons already exist within the brand folder on main ?

jordanstith15:
Am I the only one seeing the icons being replaced with large version of just a black outline of the removed icon, sometimes not even the full thing (google icon)?
Is that just a github thing?

daragh-king-genesys:
I don't think this PR is necessary as it is not adding or removing icons.
Ux (@VirtualClover) can look here https://apps.inindca.com/common-ui-docs/genesys-webcomponents/4.149.1/gux-icon-brand.html and make sure we are not missing any branch icons

PR #1114, opened 2025-05-22
Feature/com UI 1441 test

jason-evans-genesys:


packages/genesys-spark-components/src/components/stable/gux-list/gux-list.service.ts
```diff
@@ -1,7 +1,17 @@
 function getRootChildren(root: HTMLElement): Element[] {
   const slot = root.querySelector('slot');
 
-  return slot ? slot.assignedElements() : Array.from(root.children);
+  if (slot) {
+    // I think it'd be better to have more of a dynamic way to flatten the slot's assigned elements instead of this hacky way but first wanted
```
jason-evans-genesys, line 5: I didn't really change code here. This is pretty much how it used to be in this function, so you can ignore these code changes here.

packages/genesys-spark-components/src/components/stable/gux-table-toolbar/example.html
```diff
@@ -26,8 +26,18 @@ <h2>Table Toolbar Example</h2>
   </div>
 
   <div slot="menu-actions">
-    <gux-table-toolbar-action action="export"></gux-table-toolbar-action>
-    <gux-table-toolbar-action action="refresh"></gux-table-toolbar-action>
+    <!-- <gux-table-toolbar-action action="export"></gux-table-toolbar-action>
+    <gux-table-toolbar-action action="refresh"></gux-table-toolbar-action> -->
```
jason-evans-genesys, line 30: If we uncommented either of these code blocks and then added a nested slot then focus css won't trigger.

packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar.tsx
```diff
@@ -279,7 +281,17 @@ export class GuxTableToolbar {
           <div class="gux-permanent-menu-primary-wrapper">
             <slot name="permanent-actions"></slot>
             <gux-table-toolbar-menu-button show-menu={this.renderMenu()}>
-              <slot name="menu-actions"></slot>
+              {/* <slot name="menu-actions"></slot> */}
+
+              <gux-table-toolbar-custom-action>
```
jason-evans-genesys, line 286: I added this markup here to show how if we don't have two levels deep of slots then the Additional Actions dropdown buttons will focus properly. If we slot in the menu-actions then we'll have 2 slot levels and then the focus css won't trigger.

PR #1118, opened 2025-05-27

daragh-king-genesys:
update dependencies to wanted


packages/genesys-spark-components/package.json
```diff
@@ -66,7 +66,7 @@
     "@floating-ui/dom": "1.7.0",
     "@js-temporal/polyfill": "0.5.1",
     "@vvo/tzdb": "6.161.0",
-    "google-libphonenumber": "3.2.40",
+    "google-libphonenumber": "^3.2.41",
```
jordanstith15, line 69: I know it's only a patch update, but I'm a bit wary of any libphonenumber updates. I'd like to check the gux-phone-input in the demo link, but it seems to be taking a while to deploy.
Will check the npm libraries to see the difference too

PR #1129, opened 2025-05-30
fix(flyout-menu): add animation delay when showing

gavin-everett-genesys:
I am still awaiting feedback from UX in relation to the delay timing.

This solution waits `250ms` before showing the flyout menu this is the same wait time for when the flyout menu hides which resolves the issue of overlap when tabbing via keyboard. 

We have a same sort of implementation in the tooltip component where we wait `350ms` before showing the tooltip not that they are related but just to note.

[✅ Closes: COMUI-3571](https://inindca.atlassian.net/browse/COMUI-3571)

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-flyout-menu/gux-flyout-menu.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-flyout-menu/gux-flyout-menu.scss
@@ -22,9 +22,14 @@
   display: none;
   border: none;
   border-radius: ui.$gse-ui-menu-borderRadius;
+  opacity: 0;
 
   &.gux-shown {
     display: flex;
+    animation-name: fade-in;
+    animation-duration: 100ms;
+    animation-delay: 250ms;
+    animation-fill-mode: forwards;
   }
 
   .gux-arrow {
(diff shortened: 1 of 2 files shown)


gavin-everett-genesys:
Another example here (https://yale-a11y.gitlab.io/ui-component-library/accessible-flyout-menu#). If you hover a flyoutmenu and then move away there is a slight delay when closing. But if you hover over flyoutmenu and then immediately go to the next one there is no delay and it closes immediately.

packages/genesys-spark-components/src/components/stable/gux-flyout-menu/gux-flyout-menu.scss
```diff
@@ -22,9 +22,14 @@
   display: none;
   border: none;
   border-radius: ui.$gse-ui-menu-borderRadius;
+  opacity: 0;
 
   &.gux-shown {
     display: flex;
+    animation-name: fade-in;
+    animation-duration: 100ms;
+    animation-delay: 350ms;
```
gavin-everett-genesys, line 31: Confirmed by UX to standardize the delay times.

PR #1132, opened 2025-06-02
chore(tab-list): Added playwright tests

jason-evans-genesys:
✅ Closes: COMUI-3770


packages/genesys-spark-components/src/components/stable/gux-tabs/gux-tab-list/tests/gux-tab-list.common.ts
```diff
@@ -0,0 +1,6 @@
+export const renderConfigs = [
+  {
+    description: 'should render gux-tab-list',
+    html: `<gux-tab-list></gux-tab-list>`
```
jason-evans-genesys, line 4: Should I basically copy how gux-tab does playwright tests or does it make sense to have an empty gux-tab-list here like I've done?

PR #1154, opened 2025-06-06
test(table-toolbar): addition of playwright tests

gavin-everett-genesys:
addition of playwright tests

[✅ Closes: COMUI-3844](https://inindca.atlassian.net/browse/COMUI-3844)


packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar-menu-button/gux-table-toolbar-menu-button.tsx
```diff
@@ -135,10 +135,10 @@ export class GuxTableToolbarMenuButton {
                 <gux-icon
                   icon-name="fa/ellipsis-regular"
                   size="small"
-                  decorative
+                  screenreaderText={this.i18n('additionalActions')}
                 ></gux-icon>
-                {this.renderTooltip()}
               </button>
```
gavin-everett-genesys, line 140: This `button` element was violating https://dequeuniversity.com/rules/axe/4.3/input-button-name. So I move the tooltip outside the `button` element and applied `screen-reader` text to the icon.

daragh-king-genesys: I think this is repeating an issue in NVDA that @321gillian is looking to fix :) https://inindca.atlassian.net/jira/software/c/projects/COMUI/issues/?jql=project%20%3D%20%22COMUI%22%20AND%20textfields%20~%20%22A11Y%20%7C%20NVDA%20Screen%20Reader%20%22%20ORDER%20BY%20created%20DESC

daragh-king-genesys (commented):
Playwright/tests changes look ok just the accessibility change might be an issue.

packages/genesys-spark-components/src/components/stable/gux-table-toolbar/gux-table-toolbar-menu-button/gux-table-toolbar-menu-button.tsx
gavin-everett-genesys: I will discuss this with @321gillian on Monday. The other thought I had was adding just an `aria-label` to the `button` element.

321gillian: From looking at the time machine branch with NVDA, this seems to work as expected i.e. the button reads out the correct info and is clickable. Not sure why to be honest, as the below is not clickable by itself: 
```
<gux-button-slot icon-only>
  <button type="button" onclick="notify(event)">
    <gux-icon icon-name="fa/diamond-regular" screenreader-text="add John Smith to contact list"></gux-icon>
  </button>
</gux-button-slot>
```

If you want to make it more consistent with the recommended button slot approach then you can use the `gux-screen-reader-beta` tag rather than the `screenreader-text` attribute on the icon.

PR #1162, opened 2025-06-09
fix(table): Added horizontal scroll check for focus

jason-evans-genesys:
✅ Closes: COMUI-3357

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-table/gux-table.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-table/gux-table.tsx
@@ -51,10 +51,10 @@ export class GuxTable {
   private tableId: string = randomHTMLId('gux-table');
   private columnsWidths: object = {};
   /**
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-table/gux-table.tsx
```diff
@@ -436,7 +450,9 @@ export class GuxTable {
       >
         <div class="gux-table">
           <div
-            tabindex={this.isVerticalScroll ? '0' : '-1'}
+            tabindex={
+              this.isVerticalScroll || this.isHorizontalScroll ? '0' : '-1'
```
jason-evans-genesys, line 454: Do you all think it makes sense to just combine the vertical and horizontal check logic into one isScroll function? Not sure if I needed to separate the two out or not.

katie-bobbe-genesys (approved):

PR #1165, opened 2025-06-11
fix(dropdown): allow to filter dropdown on focus

gavin-everett-genesys:
Allow to filter dropdown on focus.

Open to opinions on this change, it did take a while to figure out a solution for it.

One noticeable change is if you click to open the filterable dropdown you cannot close it via click on the input you would have to click the arrow icon to close. This sort of makes sense to me as clicking on an input probably should not close the dropdown but I guess it would not be consistent with the existing behaviour we have. I looked online and some searable dropdowns do not close when clicking only when clicking on the arrow button.

[✅ Closes: COMUI-3860](https://inindca.atlassian.net/browse/COMUI-3860)

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
@@ -98,6 +98,7 @@ export class GuxDropdown {
   watchExpanded(expanded: boolean) {
     if (!expanded) {
       this.filter = '';
+      this.filterElement.value = '';
     }
   }
 
@@ -492,7 +493,7 @@ export class GuxDropdown {
   }
 
   private renderFilterInputField(): JSX.Element {
-    if (this.expanded && this.isFilterable()) {
+    if (this.isFilterable()) {
       return (
         <div class="gux-field gux-input-field" dir="auto">
           <div class="gux-field-content">
@@ -505,7 +506,6 @@ export class GuxDropdown {
               </div>
               <div class="input-and-dropdown-button">
                 <input
-                  onClick={this.fieldButtonClick.bind(this)}
                   class="gux-filter-input"
                   type="text"
                   aria-label={this.i18n('filterResults')}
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
```diff
@@ -531,10 +533,10 @@ export class GuxDropdown {
     return (
       <div
         class={{
-          'gux-target-container-expanded': this.expanded && this.isFilterable(),
-          'gux-target-container-collapsed': !(
-            this.expanded && this.isFilterable()
-          ),
+          'gux-target-container-filterable': this.isFilterable(),
+          'gux-target-container-filterable-active':
```
gavin-everett-genesys, line 537: Maybe its not idea to change these due to snapshot changes but it just made more sense to me. Let me know.

321gillian:
> > Had a look at this on the time machine and I noticed that it works fine when tabbing down the page but when coming back up via keyboard (using `shift + tab`) I get stuck opening and closing the dropdown.
> > Also when tabbing down the page, need to tab twice on the dropdown to get to the next item. Not sure if that's an issue or not https://github.com/user-attachments/assets/99c2a29a-562b-4ec8-a016-3a72bde1bf22
> 
> For the first point good spot Ive implemented a fix now 👍
> 
> For your second point, I dont think it will be possible to resolve this and if so it would probably be a bit hacky. This happens as when you initially focus on the dropdown that is `filterable` it will actually be focused on the `filterElement` ( ie the input element) so when you click tab the first time it will focus back on the actual dropdown element and the second tab press will transition to the next element in the dom. If a team mentions it I guess we could look at it.

Fix looks good. Might be good to run the second part by the UX team just to make sure that's what they expect.

321gillian:
Can't navigate through the options in the filterable list in the current time machine example (first one, other ones look ok to me)
Screenshot 2025-07-25 at 15 37 14

gavin-everett-genesys:
> Can't navigate through the options using up and down arrow keys in the filterable list in the current time machine example (first one, other ones look ok to me) Screenshot 2025-07-25 at 15 37 14

Think the focusing should be correct now let me know :)

321gillian (approved):

PR #1169, opened 2025-06-11
chore(other): switched to webappPipelineV2

daragh-king-genesys:
switched to webappPipelineV2

✅ Closes: COMUI-3744


packages/genesys-spark-components/scripts/stencil-wrapper.ts
```diff
@@ -11,35 +11,28 @@ import { defineCustomElements } from '../dist/loader';
...
+
+  '.mypurecloud.com',
+  '.mypurecloud.com.au',
+  '.mypurecloud.de',
+  '.mypurecloud.ie',
+  '.mypurecloud.jp',
+
+  '.pure.cloud',
+  '.maximus-pure.cloud'
+  // '.use2.us-gov-pure.cloud', Assets are not currently deployed to FedRAMP and should fallback to the default domain
```
jasonmobley, line 29: Rather than hard-coding all these could you add a dependency on the public service discovery package and then only hard-code the domains for lowers? I used the internal service discovery package (which includes lowers), but I could show you what I did in my package if it helps.

packages/genesys-spark/src/hosts.ts
jasonmobley, line 1: Man, it would be nice to capture this stuff in a utils package rather than repeating it everywhere.

packages/genesys-spark-components/scripts/stencil-wrapper.ts
daragh-king-genesys: I will look at this in a follow up PR

PR #1173, opened 2025-06-12
fix(tooltip): fixed memory leak

daragh-king-genesys:
fixed memory leak

✅ Closes: COMUI-3871


rinckemd:
Should cleanupUpdatePosition be called if showTooltip() is called twice with no change to the forId? I see a detached gux-tooltip hanging around for that case.

gavin-everett-genesys (approved):

katie-bobbe-genesys (approved):

PR #1174, opened 2025-06-13
fix(segmented-control-item): Improve NVDA accessibility

321gillian:
COMUI-3652

This is an imperfect solution to an issue that's likely caused by  this issue [Possible issue with NVDA support for Shadow DOM in Chromium](https://github.com/nvaccess/nvda/issues/17845). Open to other ideas but this is the only one I've found that works consistently.

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-segmented-control/gux-segmented-control-item/gux-segmented-control-item.tsx
+++ b/packages/genesys-spark-components/src/components/beta/gux-segmented-control/gux-segmented-control-item/gux-segmented-control-item.tsx
@@ -118,14 +118,14 @@ export class GuxSegmentedControlItem {
           disabled={this.disabled || this.hasDisabledParent()}
         >
           {this.renderIconSlot()}
-          <div
+          <span
             class={{
               'gux-text': true,
               'gux-icon-only': this.iconOnly
             }}
           >
             <slot name="text" />
-          </div>
+          </span>
         </button>
         {this.renderTooltip()}
       </div>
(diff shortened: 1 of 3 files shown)


packages/genesys-spark-components/src/components/beta/gux-segmented-control/gux-segmented-control-item/gux-segmented-control-item.tsx
```diff
@@ -116,6 +116,7 @@ export class GuxSegmentedControlItem {
           type="button"
           aria-current={this.selected ? 'true' : 'false'}
           disabled={this.disabled || this.hasDisabledParent()}
+          aria-label={getSlotTextContent(this.root, 'text')}
```
daragh-king-genesys, line 119: Why is the `text` slot not enough?

321gillian: According to the HTML docs, https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button the button should only contain `phrasing content`, like a `slot` or `span`. A `div` is considered `flow content` so might be taken as invalid in that context.

321gillian: Further to the above, the W3c validator will throw and error for `<button><div>Button text</div></button>`
 
https://validator.w3.org/nu/#textarea

Screenshot 2025-07-07 at 15 12 01

katie-bobbe-genesys (approved):
This looks good to me. I verified the changes with NVDA

PR #1186, opened 2025-06-18
feat(form-field) - addition of character count feature

gavin-everett-genesys:
Currently I have just implemented the feature to the `form-field-text-like` component and once this gets approved I will work on the rest of the `form-field` components.

Open to opinions on this API and any enhancements.

https://inindca.atlassian.net/browse/COMUI-3762


packages/genesys-spark-components/src/components/stable/gux-form-field/functional-components/gux-form-field-label/gux-form-field-label.scss
```diff
@@ -21,6 +21,9 @@
     }
 
     &.gux-above {
```
katie-bobbe-genesys, line 23: I think this may be causing a regression in the label for instances where there is more than one element in the label
image

gavin-everett-genesys: Good spot, addressed now 👍

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-text-like/gux-form-field-text-like.tsx
```diff
@@ -190,55 +233,71 @@ export class GuxFormFieldTextLike {
...
-      <GuxFormFieldContainer labelPosition={this.computedLabelPosition}>
-        <GuxFormFieldLabel
-          required={this.required}
-          position={this.computedLabelPosition}
-        >
-          <slot name="label" onSlotchange={() => this.setLabel()} />
-          <gux-form-field-label-indicator
-            variant={this.indicatorMark}
+      <Host>
+        {this.renderCharacterCount()}
```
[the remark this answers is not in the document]
gavin-everett-genesys: Its hard to configure it due to the `GuxFormFieldContainer` taking the `label-position` as a property. So for example when the label position is `beside` the flex box will be in a row flex direction which makes it hard to position the character counter unless I position it absolutely.

packages/genesys-spark-components/src/components/stable/gux-form-field/components/gux-form-field-text-like/gux-form-field-text-like.tsx
```diff
@@ -105,6 +115,11 @@ export class GuxFormFieldTextLike {
   @State()
   private hasHelp: boolean = false;
 
+  @State()
+  private wasExceeding: boolean = false;
+
+  @Event() characterLimitChange: EventEmitter<boolean>;
```
gavin-everett-genesys, line 121: From our call the last day I reduced the event emissions to just a single event which which will take the payload of whether the characters are exceeding or not which is a boolean value. Also maybe this should just be a generic error event potentially instead of being specific ?

I investigated other options but I do not think they are good options but let me know if you think they would be better. The first was to apply an attribute to the slotted input eg `data-character-limit-change=true|false` and the attribute would change based on the state internally. I just thought maybe by doing this it would be intrusive as we would be changing the consumers DOM. 

Another option maybe is to apply it to the host element.

packages/genesys-spark-components/src/components/stable/gux-form-field/gux-form-field.service.ts
```diff
@@ -219,3 +219,35 @@ export function getSlottedInput(
 function hasLabelSlot(root: HTMLElement): boolean {
   return Boolean(root.querySelector('label[slot="label"]'));
 }
+
+export function setCharacterCountAriaDescribedBy(): void {
+  if (this.root.hasAttribute('character-limit')) {
+    const characterlimit = this.root.shadowRoot?.querySelector(
+      '.gux-form-field-character-count'
+    );
+    console.info(characterlimit);
```
gavin-everett-genesys, line 228: will remove

PR #1193, opened 2025-06-19
fix(copy-to-clipboard): add keydown event listeners NVDA

gavin-everett-genesys:
add keydown event listeners. Space and Enter key not working in NVDA for chrome browser.

✅ Closes: COMUI-3863


gavin-everett-genesys:
Just testing out some stuff I will re add everyone when ready for review.

gavin-everett-genesys:
Using an `aria-label` on the button element resolves

PR #1194, opened 2025-06-19
chore(other): update deploy

daragh-king-genesys:
update deploy


devops/assets/pipeline.d/deploy.yml
```diff
@@ -59,7 +59,7 @@ deploy:
         family: webui
       name: prod
       schedule: off_hours
-      sequence: serial
+      sequence: parallel
```
jordanstith15, line 62: So iirc we are only promoting and not deploying to prod, but I am wondering if setting this to parallel will just auto-fail any jobs by skynet in prod. In the old deploy process, it definitely skipped/failed any auto-deploys with prod set as parallel.
Have you confirmed that this won't cause issues?

jordanstith15 (approved):

PR #1196, opened 2025-06-25
fix(copy-to-clipboard): button not clickable via keyboard in nvda

gavin-everett-genesys:
button not clickable via keyboard in nvda

✅ Closes: COMUI-3863


gavin-everett-genesys:
By adding an `aria-label` to the button element it resolves the issue in NVDA. I am assuming it has to due with the text being slotted.

conor-darcy (approved):

PR #1198, opened 2025-06-26
chore(form-field-phone): Fixed issue with double focus ring inputs

jason-evans-genesys:
Fixed issue where there was a double focus ring on the country code and input elements.

✅ Closes: COMUI-3415

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-phone-input/gux-phone-input.scss
+++ b/packages/genesys-spark-components/src/components/beta/gux-phone-input/gux-phone-input.scss
@@ -7,10 +7,10 @@
 }
 
 .gux-phone-text-input {
-  flex: 1 1 auto;
+  display: inline-flex;
+  flex: 1 1 0%;
   align-self: auto;
   order: 0;
-  inline-size: 100%;
   margin-inline-start: 4px;
   overflow: hidden;
   text-overflow: ellipsis;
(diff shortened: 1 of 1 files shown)


packages/genesys-spark-components/src/components/beta/gux-phone-input/gux-phone-input.scss
```diff
@@ -10,7 +10,10 @@
   flex: 1 1 auto;
   align-self: auto;
   order: 0;
-  inline-size: 100%;
+  inline-size: calc(
+    100% - 8px
```
jason-evans-genesys, line 14: Not sure if having 8px here is kind of arbitrary but I was trying to have a value that would prevent the input element from bleeding over to the right of its parent container.

packages/genesys-spark-components/src/components/beta/gux-phone-input/gux-phone-input.scss
```diff
@@ -38,7 +41,7 @@
   inline-size: 100%;
   block-size: ui.$gse-ui-phoneInput-dropdown-height;
   padding-block: 0;
-  padding-inline: 0 8px;
+  // padding-inline: 0 8px; /* Commented this out since having it causes extra space to the right of the child input element */
```
jason-evans-genesys, line 44: Not sure if this is problematic removing this line of code. Wanted to check first before I removed it. I don't know if it's needed and having it causes the child input element focus ring to not be flush to its parent container.

gavin-everett-genesys (approved):

PR #1199, opened 2025-06-26
chore(accordion): NVDA troubleshooting

katie-bobbe-genesys:
Adding a default slot seems to fix the NVDA issue with it not being interactive in Chrome using NVDA. Previously the 'focus' mode wasn't being triggered when the element was being focused

I tested it in NVDA on the VDI several times and confirmed that the focus mode is being triggered as expected on the feature branch. But I would love for someone else with VDI to confirm that this isn't just a fluke or something

Even the examples that only have named slots are working as expected, so this kind of fix will not be a breaking change and the fix should apply for teams that only have the named slots


packages/genesys-spark-components/src/components/beta/gux-date-time/tests/gux-date-time.e2e.playwright.ts-snapshots/gux-date-time-should-render-as-expected-for-full-format-1-darwin.png
katie-bobbe-genesys, line 1: Is there something missing in my setup that is making this happen? Or do we need to update the snapshots to use a certain time zone

katie-bobbe-genesys:
So good news is that this issue is just fixed after the Stencil update in v4.156. I confirmed that the accordion works in NVDA as expected in this version and does not work in the previous version. So this "fix" was just a fluke.

PR #1210, opened 2025-07-02
fix(month-picker-beta): add data-testid to year label

gavin-everett-genesys:
add data-testid to year label

[✅ Closes: COMUI-3910](https://inindca.atlassian.net/browse/COMUI-3910)

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-month-picker/gux-month-calendar/gux-month-calendar.tsx
+++ b/packages/genesys-spark-components/src/components/beta/gux-month-picker/gux-month-calendar/gux-month-calendar.tsx
@@ -283,7 +283,7 @@ export class GuxMonthCalendar {
             })}
           </gux-screen-reader-beta>
         </button>
-        <div>{this.getYearLabel(this.year)}</div>
+        <div data-testid="year-label">{this.getYearLabel(this.year)}</div>
         <button
           type="button"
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/beta/gux-month-picker/gux-month-calendar/gux-month-calendar.tsx
```diff
@@ -283,7 +283,7 @@ export class GuxMonthCalendar {
             })}
           </gux-screen-reader-beta>
         </button>
-        <div>{this.getYearLabel(this.year)}</div>
+        <div data-testid="year-label">{this.getYearLabel(this.year)}</div>
```
321gillian, line 286: I understand the idea since we had this previously but I'm wondering what's our stance on testing using the internal details of a component like this? 

Also, since we don't this `data-testid` in the component library it could be removed unwittingly later on so we might need to add a note if we want to make sure that doesn't happen.

jason-evans-genesys: Maybe one benefit of having a class instead of data-testid is if something other than a test file wanted to query this element. Thoughts?

321gillian (approved):

PR #1215, opened 2025-07-03
fix(popover-beta): add method to focus popover

gavin-everett-genesys:
From my understanding of the ticket they have a `for` element which when they press the enter key renders the popover but the `autofocus` will not work due to the fact that the `gux-popover-beta` wont have been rendered therefore missing the `keydownHandler` that we have within the component.

Let me know if there is any drawbacks to this choice.

[✅ Closes: COMUI-3916](https://inindca.atlassian.net/browse/COMUI-3916)

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-popover-beta/example.html
+++ b/packages/genesys-spark-components/src/components/beta/gux-popover-beta/example.html
@@ -27,7 +27,52 @@
   </div>
 </div>
 
-<style>
+<h2>Dynamic Popover with guxFocusPopover()</h2>
+<div>
+  <gux-button id="dynamic-trigger">Show Dynamic Popover</gux-button>
+  <div id="dynamic-popover-container"></div>
+</div>
+
+<style
+  onload="(function () {
+  const dynamicTrigger = document.getElementById('dynamic-trigger');
+  const container = document.getElementById('dynamic-popover-container');
+    
+
+  dynamicTrigger.addEventListener('click', function() {
+      const popover = document.createElement('gux-popover-beta');
+      popover.id = 'dynamic-popover';
+      popover.setAttribute('position', 'bottom');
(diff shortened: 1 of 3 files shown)


packages/genesys-spark-components/src/components/beta/gux-popover-beta/gux-popover.tsx
```diff
@@ -133,6 +133,11 @@ export class GuxPopover {
     this.dismiss();
   }
 
+  @Method()
+  async guxFocusPopover(): Promise<void> {
+    this.focusPopup();
```
jason-evans-genesys, line 138: Is it worth calling this focus method from the example page to demonstrate it works?

packages/genesys-spark-components/src/components/beta/gux-popover-beta/example.html
```diff
@@ -27,7 +27,52 @@ <h1>gux-popover</h1>
...
+  <div id="dynamic-popover-container"></div>
+</div>
+
+<style
+  onload="(function () {
+  const dynamicTrigger = document.getElementById('dynamic-trigger');
+  const container = document.getElementById('dynamic-popover-container');
+    
+
+  dynamicTrigger.addEventListener('click', function() {
```
jason-evans-genesys, line 42: Do we need to be dynamically creating this popover element like this or could we just declare the HTML like normal and then reference that element by `document.getElementById` and then call `popover.guxFocusPopover();` on it?

jason-evans-genesys: Makes sense! No worries. I noticed the autofocus works in chrome but not firefox.

gavin-everett-genesys: Is this occurring for you on the latest version of firefox? I could not reproduce when hitting the `enter` or `space` key ?

PR #1218, opened 2025-07-04
style(menu_option): applying fixed height to menu option (32px)

daragh-king-genesys:


packages/genesys-spark-components/src/components/beta/gux-avatar/tests/gux-avatar.e2e.playwright.ts-snapshots/gux-avatar-beta-Should-render-as-expected-with-image-slot-1-darwin.png
gavin-everett-genesys, line 1: This is my bad this will probably change with every PR I think probably best to remove the renderConfig or just add something static 🙈

gavin-everett-genesys (approved):

PR #1232, opened 2025-07-15
fix(tag): stopping screenreader text from bleeding out in some cases

conor-darcy:
Reported by Anton Filippov working on Knowledge Workbench.
This removes the extra space added to the bottom of the page below, the pagination should be fixed to the bottom.

image

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-tag/gux-tag.scss
+++ b/packages/genesys-spark-components/src/components/stable/gux-tag/gux-tag.scss
@@ -10,7 +10,6 @@ $focus-outline-offset: 5px;
 }
 
 .gux-tag {
-  position: relative;
   display: flex;
   flex-direction: row;
   flex-wrap: nowrap;


packages/genesys-spark/snapshot/scss/_mixins.scss
```diff
@@ -28,8 +28,8 @@
 @mixin gux-sr-only-clip {
   &:not(:focus):not(:active) {
     position: absolute;
-    width: 1px;
-    height: 1px;
+    width: 0;
+    height: 0;
```
daragh-king-genesys, line 32: I don't think we can do this. I think the element has to have some height and width to be rendered

daragh-king-genesys (commented):
Did you have an example that reproduced the issue in our docs site?

packages/genesys-spark/snapshot/scss/_mixins.scss
conor-darcy: @daragh-king-genesys It's the screenreader text, we don't want it rendered, or am I missing something?
Or is there a difference between rendered & visible?

daragh-king-genesys: Ya have a read of this https://webaim.org/techniques/css/invisiblecontent/

conor-darcy: @daragh-king-genesys Thanks. I think I don't need it anyway.

daragh-king-genesys:
@gavin-everett-genesys do you think your screen reader component change might make this change unnecessary?

PR #1250, opened 2025-07-18
chore(other): add token for tooltip max-width

katie-bobbe-genesys:


katie-bobbe-genesys:
> Not sure what the process is here if I am honest. Should this include the code changes or will that be in another PR or will UX have to open another jira issue for that?
> 
> Obviously what you have now is fine for the token addition but just want to make sure we don't miss applying it too.

I am going to make a separate PR for applying the tokens. This is the branch that UX created and I just ran `new-tokens-data` renamed the commit to make it so that it would pass commit lint

PR #1260, opened 2025-07-22
fix(dropdown): Show drop-shadow in dropdowns

321gillian:
✅ Closes: COMUI-4044


packages/genesys-spark-components/src/components/stable/gux-popup/gux-popup.tsx
```diff
@@ -124,8 +124,7 @@ export class GuxPopup {
                   // Everything else is constrained to the width of the target.
                   // Note: if the contents overflow the flip and shift middleware will not detect it
                   Object.assign(elements.floating.style, {
-                    width: `${rects.reference.width}px`,
-                    overflow: 'hidden'
```
321gillian, line 128: I couldn't get back far enough in the git history to see why this was originally added but if anyone remembers please let me know.

PR #1261, opened 2025-07-22
chore(copy-to-clipboard): Added playwright test

jason-evans-genesys:
✅ Closes: COMUI-3789


packages/genesys-spark-components/src/components/stable/gux-copy-to-clipboard/tests/gux-copy-to-clipboard.e2e.playwright.ts
```diff
@@ -0,0 +1,31 @@
...
+    await setContent(page, renderConfig.html);
+
+    const element = page.locator('gux-copy-to-clipboard');
+
+    await element.hover();
+    await page.waitForTimeout(2000);
+    // await page.waitForSelector("gux-tooltip")
+
+    const tooltip = element.locator('gux-tooltip');
+    await expect(tooltip).toContainText('Click to Copy');
```
jason-evans-genesys, line 29: Anyone have any idea why `await expect(tooltip).toContainText('Click to Copy');` doesn't work as intended? Looks like the text that comes back from playwright running this test is an empty string. Anyone have an idea how to fix this?

PR #1272, opened 2025-07-23
fix(switch-legacy): add aria-current to selected item

katie-bobbe-genesys:
Even though this is a legacy component this is an accessibility fix that adds `aria-current` to the selected item

How do I update the snapshots locally? The npm command `npm run test.update-snapshot` in `genesys-spark-components` was not working


daragh-king-genesys:
> Even though this is a legacy component this is an accessibility fix that adds `aria-current` to the selected item
> 
> How do I update the snapshots locally? The npm command `npm run test.update-snapshot` in `genesys-spark-components` was not working

I will have a look at the command and see why it is broken.
As a workaround you can delete the file that contains the failing snapshots and a new one will be created on the next run

PR #1276, opened 2025-07-24

daragh-king-genesys:
update dependencies to wanted


shared-configs/eslint-config-genesys-spark-components/package.json
```diff
@@ -23,7 +23,7 @@
     "prettier": "^3.6.2",
     "prettier-config-genesys-spark-components": "file:../prettier-config-genesys-spark-components",
     "prettier-package-json": "^2.8.0",
-    "typescript-eslint": "^8.37.0"
+    "typescript-eslint": "^8.38.0"
```
jason-evans-genesys, line 26: Correct me if I'm wrong, but wouldn't running npm install in genesys-spark-components do this below:

`When you use npm install with a dependency version specified with a caret (^) in your package.json file, it instructs npm to install the latest compatible version of that package within a defined range`

So wouldn't `^8.37.0` evaluate to the same thing as `^8.38.0`?

jason-evans-genesys:
Just for my own understanding, what's the motivation for updating these dependencies?

PR #1308, opened 2025-08-13
feat(tree-view-beta): tree view component single select

katie-bobbe-genesys:
looking for feedback on API and general comments. Multi select will be added separately. Still need to add keyboard nav

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-tree-view/gux-tree-view-branch/gux-tree-view-branch.tsx
+++ b/packages/genesys-spark-components/src/components/beta/gux-tree-view/gux-tree-view-branch/gux-tree-view-branch.tsx
@@ -1,0 +1,108 @@
+import {
+  Component,
+  Element,
+  Event,
+  EventEmitter,
+  h,
+  JSX,
+  Prop,
+  Watch
+} from '@stencil/core';
+import { GuxTreeViewLayout } from '../gux-tree-view-types';
+import { randomHTMLId } from '@utils/dom/random-html-id';
+
+/**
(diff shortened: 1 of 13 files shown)


packages/genesys-spark-components/src/components/beta/gux-tree-view/gux-tree-view-branch/gux-tree-view-branch.tsx
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  private toggle() {
+    this.open = !this.open;
+    this.selected = true;
+  }
+
+  render(): JSX.Element {
+    return (
+      <Host role="treeitem">
```
katie-bobbe-genesys, line 67: @daragh-king-genesys Could I get a second set of eyes on this if you have a chance? I'm having trouble thinking through how to get an accessible API working here.

question 1: Should the ids and roles be in the light dom or shadow dom? The way I have it now is that it's in the light dom but I'm not sure if this is too limiting
question 2: 
I was trying to think through how to get the branch API to match the example here: https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-1b/

So in the above W3C example, the markdown for a branch is like this:

```
<li role="treeitem"
      aria-level="1"
      aria-setsize="3"
      aria-posinset="1"
      aria-expanded="false"
      aria-selected="false">
    <span>
      Projects
    </span>
    <ul role="group">
    // other list items slotted here
    </ul>
    </li>
    ```
    
Basically I'm not sure how to support this branch markdown with the current API, unless we move the ids and roles to the shadow dom

PR #1320, opened 2025-08-19
fix(tabs): change tab styling so it does not rely on attribute selector

katie-clark-genesys:
This should fix a bug in Vue related to the vertical tab styling


daragh-king-genesys (approved):

daragh-king-genesys (commented):
Code looks good need to update snapshots

PR #1326, opened 2025-08-21
fix(dropdown): Added a null check for filterElement before setting its value

jason-evans-genesys:
✅ Closes: COMUI-4127

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
@@ -96,7 +96,7 @@ export class GuxDropdown {
 
   @Watch('expanded')
   watchExpanded(expanded: boolean) {
-    if (!expanded) {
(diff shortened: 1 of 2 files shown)


packages/genesys-spark-components/src/components/stable/gux-dropdown/gux-dropdown.tsx
```diff
@@ -96,7 +96,7 @@ export class GuxDropdown {
 
   @Watch('expanded')
   watchExpanded(expanded: boolean) {
-    if (!expanded) {
+    if (!expanded && this.filterElement) {
       this.filter = '';
```
gavin-everett-genesys, line 100: solution looks good 👍  I think this also occurs in the `dropdown-multi` so I think we should add it there aswell.

jason-evans-genesys: Good catch, fixed

gavin-everett-genesys (approved):

PR #1331, opened 2025-08-25
chore(other): fix scss build

daragh-king-genesys:
fix scss build


jordanstith15 (approved):
is it a lint issue causing the build failure or something like that?

daragh-king-genesys:
> is it a lint issue causing the build failure or something like that?

It was a sass deprecation warning. Not affecting the build at the moment but I have a preference to keep the number of warnings/ logs during the build to a minimum so new issues don't get missed.

PR #1343, opened 2025-09-01
test(form-field-range): added Playwright tests

daragh-king-genesys:
added Playwright tests

✅ Closes: COMUI-4054


packages/genesys-spark-components/src/test/specTestUtils.ts
```diff
@@ -20,14 +22,17 @@ export async function newSpecPage(opts: NewSpecPageOptions) {
 export async function checkRenders(
   renderConfigs: RenderConfig[],
   components: unknown[],
-  language: string = 'en'
+  language: string = 'en',
+  extraActions: ExtraActionsFn = () => Promise.resolve()
```
321gillian, line 26: What does this do?

PR #1378, opened 2025-09-17
feat(tag): added content slot

daragh-king-genesys:
added content slot

✅ Closes: COMUI-3935


packages/genesys-spark-components/src/components/stable/gux-tag/gux-tag.tsx
```diff
@@ -83,9 +83,15 @@ export class GuxTag {
...
+            name="content"
             aria-hidden="true"
             onSlotchange={this.onSlotChange.bind(this)}
-          />
+          >
+            <slot
+              aria-hidden="true"
+              onSlotchange={this.onSlotChange.bind(this)}
+            />
+          </slot>
```
gavin-everett-genesys: If we are going ahead with this change should we add this to the `v5` epic to be removed?

PR #1383, opened 2025-09-17
fix(other): fix anchor tag focus ring when certain elements are nested

katie-bobbe-genesys:
In Chrome, if certain elements, such as `gux-truncate` or a `div` were nested inside an anchor tag, the focus ring would not show. I have a potential fix here, but I'm not sure if it's a breaking change or not. Would like the team's opinon.


katie-bobbe-genesys:
This change looks good, there seem to be some failing tests though. Once those are resolved I can verify the changes

daragh-king-genesys:
Does it break any of the playwright tests? ie does it cause a visual change in our tests?

PR #1390, opened 2025-09-19
chore(other): update style-dictionary dependency to latest

daragh-king-genesys:
update style-dictionary dependency to latest


packages/genesys-spark-tokens/data/ui/data_display.json
```diff
@@ -993,12 +993,6 @@
       }
     }
   },
-  "global_nav": {
-    "side_menu": {
-      "value": "linear-gradient(135deg, {color.emerald.500} 0%, {color.emerald.200} 100%)",
-      "type": "color"
-    }
-  },
```
daragh-king-genesys, line 1001: @VirtualClover this upgrade found an issue in the data files. This entry was silently stopping all the `global_nav.side_menu.XXXXX` tokens in `navigation.json`
I moved it into `navigation.json` and called it `global_nav.side_menu.background_color`

daragh-king-genesys, line 1001: Confirmed with @samlimby-genesys that this was a bug not a wanted token.
This token was blocking other wanted tokens from being created that is why I deleted instead of deprecating.

PR #1402, opened 2025-09-24
feat(rich-text-editor) - POC drag and drop (Feedback)

gavin-everett-genesys:
This is just a quick POC to demonstrate the drag an drop functionality can be achieved within the `rich-text-editor`. 

I am looking for feedback with the current implementation and if there is anything that you think may not work. There is not tokens or anything as of now I just wanted to see what is possible.

I currently have it setup with `TipTap` but can be configured with other text editors. 

If everyone is happy I will reach out to UX for actual designs for implementation.

### DIFF
--- a/packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
+++ b/packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
@@ -34,7 +34,6 @@ import { OnMutation } from '@utils/decorator/on-mutation';
 })
 export class GuxRichTextEditor {
   private i18n: GetI18nValue;
-  private dragAndDropObserver: MutationObserver;
 
   @Element()
   root: HTMLElement;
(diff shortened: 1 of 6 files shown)


packages/genesys-spark-components/src/components/beta/gux-rich-text-editor/gux-rich-text-editor-beta.tsx
```diff
@@ -309,6 +326,34 @@ export class GuxRichTextEditor {
...
+        this.showDropOverlay = true;
+      });
+
+      editorElement.addEventListener('dragleave', (e: DragEvent) => {
+        if (!editorElement.contains(e.relatedTarget as Node)) {
+          this.showDropOverlay = false;
+        }
+      });
+
+      this.root.addEventListener('filesDropped', (e: CustomEvent) => {
```
gavin-everett-genesys, line 343: The reason for not using `drop` on the editorElement was that Tiptaps editor enviroment was actually intercepting the events from what I could tell so `drop` event would never execute. Let me know if you see anything wrong with this approach.

PR #1403, opened 2025-09-24
fix(button-slot): set inline-size to initial for inline accent

gavin-everett-genesys:
I thought this was not a bug at first but after reading the success criterion ( https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html) it states that if the element is inline or part of a sentence it does not need a minimum width or height.

✅ Closes: COMUI-4203


gavin-everett-genesys:
I added a new renderConfig the build is now failing so I am looking into it.

PR #1404, opened 2025-09-24
chore(column-manager-item): Proof of concept for moving items with mouse click (WIP)

jason-evans-genesys:
This is a WIP and a proof of concept. I need to refine this code more.

✅ Closes: COMUI-3354

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-column-manager/gux-column-manager-item/gux-column-manager-item.tsx
+++ b/packages/genesys-spark-components/src/components/stable/gux-column-manager/gux-column-manager-item/gux-column-manager-item.tsx
@@ -182,6 +182,28 @@ export class GuxColumnManagerItem {
     this.setReorderMode(!this.isReordering, true);
   }
 
+  private onMouseDown(): void {
+    const elementToReorder = document.querySelector(
+      'gux-column-manager-item[gs-reorder-indicator="below"]'
+    );
+
+    if (elementToReorder) {
+      const oldIndex = getIndexInParent(elementToReorder as HTMLElement);
+      const dropIndex = getIndexInParent(this.root);
+      if (oldIndex === dropIndex) {
+        return;
+      }
+
+      this.pendingReorder = 'none';
+
+      const newIndex = getNewIndex(oldIndex, dropIndex, null);
+
+      this.internal_order_change.emit({ oldIndex, newIndex });
+    }
+
+    this.setReorderMode(!this.isReordering, true);
+  }
+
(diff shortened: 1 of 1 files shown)


packages/genesys-spark-components/src/components/stable/gux-column-manager/gux-column-manager.tsx
```diff
@@ -48,6 +48,7 @@ export class GuxColumnManager {
   private searchElement: HTMLInputElement;
   private announceElement: HTMLGuxAnnounceBetaElement;
   private i18n: GetI18nValue;
+  private startMouseClickIndex: number = null;
```
jason-evans-genesys, line 51: Trying to see if there's a way to not introduce this new variable

gavin-everett-genesys:
I think the way of moving it with mouse click seems good to me, the only alternatives I have seen are maybe using an arrows to shift the items but if you have a long list it might be unreasonable. This will probably need UX input.

packages/genesys-spark-components/src/components/stable/gux-column-manager/gux-column-manager-item/gux-column-manager-item.tsx
```diff
@@ -182,6 +185,12 @@ export class GuxColumnManagerItem {
     this.setReorderMode(!this.isReordering, true);
   }
 
+  private onMouseDown(): void {
+    // event.stopPropagation(); // stops the browser from redirecting.
+    this.toggleReorderMode();
+    this.internal_mouse_reorder_move.emit(this.text);
```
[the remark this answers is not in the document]
jason-evans-genesys: I thought about that but it looks like the current implementation I've done doesn't really have much code overlap. Do you think it still makes sense to combine events? Is the purpose of that to limit event calls?

packages/genesys-spark-components/src/components/stable/gux-column-manager/gux-column-manager-item/gux-column-manager-item.tsx
```diff
@@ -231,7 +249,7 @@ export class GuxColumnManagerItem {
 
   render(): JSX.Element {
     return (
-      <Host draggable="true">
+      <Host draggable="true" onMouseUp={event => this.onMouseUp(event)}>
```
jason-evans-genesys, line 252: I added this onMouseUp event call here because using the draggable attribute will have the drag API override onClick and onMouseUp events on clickable elements like a button, so I took a more manual approach here.

PR #1412, opened 2025-09-30
fix(gux-dropdown): Use all slotted option content not just first node

jasonmobley:
Fix an issue where <gux-option> elements whose first slotted node was whitespace would make the selected option text show as blank when that option was selected.

Also addresses an issue where the starts-with filter mode would fail to find the option due to its leading whitespace.

https://inindca.atlassian.net/browse/COMUI-4223

### DIFF
--- a/packages/genesys-spark-components/src/components/stable/gux-listbox/gux-listbox.service.ts
+++ b/packages/genesys-spark-components/src/components/stable/gux-listbox/gux-listbox.service.ts
@@ -1,4 +1,5 @@
 import { ListboxOptionElement } from './options/option-types';
+import { getTextContentFromNodes } from '@utils/dom/get-text-content-from-nodes';
 
 export function getListOptions(
   list: HTMLGuxListboxElement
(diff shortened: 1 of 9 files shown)


packages/genesys-spark-components/src/components/stable/gux-listbox/gux-listbox.service.ts
```diff
@@ -216,17 +217,18 @@ export function matchOption(
...
-    ?.textContent.trim()
+  return getOptionDefaultSlotText(option)
     .toLowerCase()
     .startsWith(matchString.toLowerCase());
 }
 
-export function getOptionDefaultSlot(
-  option: ListboxOptionElement
-): Node | undefined {
-  return option.shadowRoot.querySelector('slot')?.assignedNodes()[0];
```
jasonmobley, line 229: I opted to completely replace the prior method with the new one that returns the text content as a string. All the prior usages were doing the same `.textContent.trim()` thing anyway, and its main problem was that it returned only the first node from `assignedNodes()`. I'm open to keeping it if this would be a breaking change or if there are use cases I'm not thinking of.

daragh-king-genesys (approved):

PR #1422, opened 2025-10-06
feat(button-slot): added ai accent

daragh-king-genesys:
added ai accent

✅ Closes: COMUI-4241


VirtualClover:
Hi @daragh-king-genesys 👋🏼 , overall it LGTM BUT we'll probably also need to add another AI button accent, that looks like a primary button but with the glowy effect on it on the near future

image

So, maybe we can change the accent name from just `ai` to something like `copilot`?
That way, there will be no confusion WDYT?

packages/genesys-spark-components/src/components/stable/gux-button-slot/gux-button-slot.scss
```diff
@@ -197,3 +197,45 @@
...
+  ::slotted(input[type='button']:active:enabled),
+  ::slotted(input[type='submit']:active:enabled),
+  ::slotted(button:active:enabled) {
+    color: #fff;
+    background: linear-gradient(
+      54deg,
+      #{ui.$gse-ui-copilotButton-active-backgroundColor-bottomLeft} 7.17%,
+      #{ui.$gse-ui-copilotButton-active-backgroundColor-middle} 46.7%,
+      #{ui.$gse-ui-copilotButton-active-backgroundColor-topRight} 83.3%
+    );
```
gavin-everett-genesys: This looks fine id be just worried if those percentages change would something like what was done for avatar tokens work here for a linear gradient for example, 

`  --gse-ui-avatar-presenceRing-busy: linear-gradient(
    135deg,
    #e22245 0%,
    #742737 100%
  );`

VirtualClover: Would it be ideal for the token to contain the whole gradient value?

The reason we usually do not create gradient tokens is that Figma has an issue with those(since Figma variables do not support gradients)

But we can find a workaround in Figma if full gradient values are preferred

PR #1425, opened 2025-10-07
fix(dropdown): add listbox role to component

daragh-king-genesys:
add listbox role to component

✅ Closes: COMUI-3579


katie-bobbe-genesys:
The `gux-listbox` which is nested within the `gux-dropdown` also has a role of `listbox`, It seems odd to have a listbox nested within another

